Issue #625, opened 2026-07-07
grouper: add ZFS root filesystem variant


hanthor:
## Summary
Add a ZFS-root variant of grouper (Ubuntu 26.04 Resolute Raccoon) that uses ZFS as the root filesystem instead of the default (ext4/xfs).

## Context
grouper currently builds as a standard bootc image. bootc supports `install to-disk --filesystem zfs` but this requires ZFS kernel modules and userspace tools in the image.

## Required changes
- Add ZFS kernel module (zfs-dkms or similar) to the grouper Containerfile
- Add ZFS userspace tools (zfsutils-linux or openzfs-zfsutils)
- Potentially add a new flavor like `gnome-zfs` or `gnome-ubuntu-zfs` in `.github/build-config.yml`
- Ensure `scripts/resolve-flavor.sh` handles the new flavor
- Create a separate Containerfile.grouper-zfs or add conditional logic to Containerfile.ubuntu

## Considerations
- ZFS DKMS needs the kernel headers package installed
- ZFS on root requires an initramfs hook (zfs-initramfs) and bootloader ZFS support
- Ubuntu 26.04 ships OpenZFS 2.x by default which has good bootc/ostree compatibility
- May want to start as a single `gnome-zfs` flavor before expanding to KDE etc.

hanthor:
## Feasibility investigation (2026-07-10)

**Correction to the premise:** bootc does **not** support `install to-disk --filesystem zfs`. Verified against bootc 1.15.2:

```
--filesystem <FILESYSTEM>   Target root filesystem type   [possible values: xfs, ext4, btrfs]
```

bootc's own help: *"Use `install to-filesystem` for anything more complex such as RAID, LVM, LUKS etc."* — ZFS root is in that bucket. So ZFS-on-root means: create the zpool/dataset yourself, then `bootc install to-filesystem` into it. Also note bootc's default backend is **composefs**, which wants **fs-verity**; ZFS has no fs-verity, so a ZFS root needs `--block-setup direct`-style / non-composefs deployment or `--allow-missing-verity`. This is the real complexity, not the packages.

## What's verified working (the image-side foundation)

grouper installs Ubuntu's `linux-generic` (7.0.0-27). ZFS on Ubuntu 26.04 (resolute) needs **no DKMS** — there's a prebuilt, signed module:

```
/usr/lib/modules/7.0.0-27-generic/kernel/zfs/zfs.ko.zst
[...]

PR #645, opened 2026-07-10
feat(grouper): gnome-zfs — ZFS-root image foundation (#625) [draft]

hanthor:
> **Draft — image-side foundation only.** Addresses #625. The ZFS-root *boot*
> path still needs an install-to-filesystem e2e (see Follow-on) before this is
> promotable, hence draft.
>
> **Stacks on #635** (branched off `fix/cache-invalidation`). Until #635 merges,
> the diff here includes #635's commits; it cleans up once #635 lands. Review
> only the ZFS commit (`feat(grouper): gnome-zfs …`).

## What & why

Issue #625 asked for a ZFS-root grouper via `bootc install to-disk --filesystem zfs`. That flag **doesn't exist** — verified against bootc 1.15.2, `--filesystem` accepts only `xfs, ext4, btrfs`. A ZFS root is instead installed with `bootc install to-filesystem` onto a **pre-created zpool**, and because the default **composefs** backend needs **fs-verity** (ZFS has none), it needs a non-composefs / `--allow-missing-verity` deployment.

So this PR delivers the **image-side foundation**: a `gnome-zfs` grouper flavor whose image ships everything needed to *be* a ZFS root, leaving the install/boot procedure as the tracked follow-on.

## Verified on `ubuntu:resolute` (26.04)
- **Prebuilt, signed `zfs.ko`** in `linux-modules-extra-<kver>` (`linux-main-modules-zfs-<kver>-generic`) — **no DKMS**.
- `zfsutils-linux` 2.4.1 userspace.
- **`zfs-dracut`** ships the dracut `90zfs` module — the right integration since grouper builds its initramfs with **dracut** (in `finalize.sh`), not initramfs-tools.

## Changes
- `build_scripts/zfs.sh` (apt-only): install prebuilt `zfs.ko` + `zfsutils-linux` + `zfs-dracut`; `modules-load.d` + a `dracut.conf.d` drop-in forcing `add_dracutmodules+=" zfs "` (finalize.sh runs `dracut --no-hostonly`, which won't auto-detect a ZFS root at build time).
- `Containerfile.ubuntu`: `gnome-zfs` stage = GNOME + `zfs.sh` **before** `finalize.sh` (finalize bootcifies → apt unusable after).
- `build-config.yml`: grouper `gnome-zfs` flavor, `build_iso: false` for now.

## Follow-on (before un-drafting / promoting)
1. A `--zfs` / install-to-filesystem **boot e2e** — same pattern as the new `--luks` gate (`scripts/iso-e2e.sh`): create a zpool, `bootc install to-filesystem`, boot, confirm the pool imports in the initramfs and root mounts.
2. Decide composefs vs. non-composefs deployment on ZFS.
3. Enable the ISO + KDE/etc. once the boot path is proven.

### DIFF
--- a/.github/build-config.yml
+++ b/.github/build-config.yml
@@ -645,6 +645,12 @@ variants:
         stage: 2
         build_image: true
         build_iso: true
+      # GNOME with ZFS-root support (prebuilt zfs.ko + zfs-dracut). ISO gated
+      # off until the install-to-filesystem ZFS-root path has a boot e2e (#625).
+      - id: gnome-zfs
+        stage: 2
+        build_image: true
+        build_iso: false
       - id: kde
         stage: 2
         build_image: true
--- a/.github/build-config.yml
+++ b/.github/build-config.yml
@@ -75,6 +75,11 @@ variants:
         build_image: true
         build_iso: true
         build_qcow2: true
+      - id: cosmic
+        stage: 2
+        build_image: true
+        build_iso: true
+        build_qcow2: false
 
       - id: kde
         stage: 2
@@ -331,6 +336,11 @@ variants:
         build_image: true
         build_iso: true
         build_qcow2: true
+      - id: cosmic
+        stage: 2
+        build_image: true
+        build_iso: true
+        build_qcow2: false
 
       - id: kde
         stage: 2
@@ -375,6 +385,26 @@ variants:
         build_iso: false
         build_qcow2: false
       - id: cosmic-nvidia
+        stage: 3
+        build_image: true
+        build_iso: true
+        build_qcow2: false
+      - id: kde-hwe
+        stage: 3
+        build_image: true
+        build_iso: false
+        build_qcow2: false
+      - id: kde-nvidia
+        stage: 3
+        build_image: true
+        build_iso: true
+        build_qcow2: false
+      - id: niri-hwe
+        stage: 3
+        build_image: true
+        build_iso: false
+        build_qcow2: false
+      - id: niri-nvidia
         stage: 3
         build_image: true
         build_iso: true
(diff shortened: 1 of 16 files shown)


hanthor:
First `grouper:gnome-zfs` build result: the image assembles gnome + the ZFS packages, but **finalize.sh's dracut fails**: `dracut[E]: Module 'zfs' cannot be found`. So the dracut `90zfs` module from `zfs-dracut` isn't visible to grouper's dracut at initramfs time. Likely causes to chase (the boot-integration TODO this draft flags): grouper's dracut comes from the bootc-deb stack and may not scan `/usr/lib/dracut/modules.d/90zfs`, or `zfs-dracut` didn't fully install against that dracut. Next: confirm `zfs-dracut` installs cleanly alongside the bootc-deb dracut and that the module dir is on dracut's search path (or pass `--dracutmodulesdir`). Foundation (module + tools) is otherwise in place.

hanthor:
**Found the working reference** you mentioned: `tuna-os/ubuntu/ubuntu-26.04` already has a proven ZFS-root Ubuntu-bootc impl (`src/zfs-install.sh` + `Containerfile` with `dracut --add "dmsquash-live zfs"` + BATS tests). Reusing it here instead of reinventing:
- Ported `zfs-install.sh` verbatim to `/usr/local/bin/zfs-install` — the real ZFS-root install-to-filesystem procedure (altroot `-R /mnt/target`, `rpool/root`+`rpool/var` `canmount=noauto`, mount only root because bootc rejects targets with extra submounts, `spl_hostid` karg, post-install `canmount=on`+`/etc/hostid`+`zpool.cache`, optional aes-256-gcm encryption).
- `zfs.sh` now hard-verifies the dracut 90zfs module (the earlier 'Module zfs cannot be found'); the reference confirms the fix is to add the `zfs` dracut module explicitly.

Remaining: confirm `zfs-dracut` installs in grouper's bootc-deb image so finalize.sh's dracut includes 90zfs (align with the reference's explicit `--add`).

PR #679, opened 2026-07-18
LUKS E2E hardening: offline-store flow fixes accumulated to green

hanthor:
Successor to #666 (closed): the offline-store LUKS flow plus every harness fix from the green-chase — swtpm restart over preserved tpmstate for the reboot phase, installed-boot wait honoring --timeout, serial console karg appended to installed BLS entries before live shutdown, screendump evidence on timeout, and the relay pull fallback. Install phase is fully green (encrypted disk + TPM enrollment confirmed, TPM auto-unlock to login verified once); remaining investigation is the installed system's graphical-target reboot cycle, which the now-working serial will expose.


hanthor:
**Root cause isolated — installed system prompts for a passphrase instead of TPM auto-unlocking.** With the serial console now reaching the installed BLS entries (run 29633676626), the reboot phase serial ends at:

```
Starting systemd-cryptsetup@root.service - Cryptography Setup for root...
Please enter passphrase for disk root::
```

So the whole chain up to here is **green**: live boot → offline-store install → `TUNAOS_LUKS_E2E_ENCRYPTED_DISK_CONFIRMED` → `TUNAOS_LUKS_E2E_TPM_ENROLLMENT_CONFIRMED` → installed disk boots and finds the LUKS root — it just doesn't auto-unlock.

The remaining blocker is **TPM2 unlock across the swtpm restart**. swtpm exits when the install-phase QEMU disconnects (confirmed — that's why the reboot phase needs the restart), and the restart uses `--flags startup-clear`. Candidate causes, in order:
1. **PCR-binding mismatch**: fisherman's tpm2-luks enrollment binds the sealed key to PCR value(s) measured in one boot context; the reboot's freshly-`startup-clear`ed swtpm + firmware re-measurement yields different PCRs → policy fails → passphrase fallback. (An earlier run reached login in 15s, suggesting this is timing/context-sensitive.)
2. **NV state not flushed**: swtpm may exit without persisting the enrolled key to tpmstate before the restart reads it.

This needs a deliberate fix (enroll without PCR binding for the test, or a persistent swtpm whose state provably survives), not a quick guess — flagging for design input rather than spending more CI cycles speculating. Everything else in the LUKS flow is proven.

hanthor:
## Root cause of the passphrase prompt: PCR 7 binding + swtpm restart

`fisherman/internal/luks/luks.go:91-93` enrolls with:
```
systemd-cryptenroll --tpm2-device=auto --tpm2-pcrs=7
```
PCR 7 = Secure Boot state. The sealed LUKS key can only be unsealed when
PCR 7 at unlock matches PCR 7 at enrollment.

**Why it fails in the harness:** enrollment happens in the live-install
boot (first QEMU / swtpm instance state), then the reboot phase runs a
**restarted swtpm with `--flags startup-clear`** (PCRs reset to 0, then
re-measured by the installed boot's OVMF). Across that restart, PCR 7 is
not guaranteed to reproduce the enrollment-time value — so unseal fails
and cryptsetup falls back to the passphrase prompt. The install +
enrollment themselves are fully correct (`ENCRYPTED_DISK_CONFIRMED` +
`TPM_ENROLLMENT_CONFIRMED`).

### Fix options (pick one)
1. **Persistent swtpm** — one instance for both QEMU launches (the
   documented design intent + dakota-iso precedent), so PCR 7 is measured
   once and never reset. Requires keeping swtpm alive across the
   install-QEMU exit (it currently dies on ctrl-channel close).
2. **Test-mode enrollment without PCR binding** — a fisherman flag /
   recipe option to enroll `--tpm2-pcrs=` (empty) in CI, testing the TPM
   unlock *mechanism* without the fragile PCR-reproducibility requirement
   that swtpm restarts can't satisfy. Production keeps PCR 7.
3. **Bind to a stable PCR** in test (e.g. none, or PCR 0 with a pinned
   firmware) — narrower.

Recommend option 1 (closest to real hardware behavior) with option 2 as
the CI fallback if a persistent swtpm proves impractical under Actions.

### Coverage status
Install + LUKS-encrypt + TPM-enroll is **green across flavors**; only the
reboot auto-unlock assertion is blocked, and it blocks every flavor
equally — so this is the single gate for community-variant LUKS coverage.

hanthor:
## Correction: the deeper root cause is live-env enrollment, not swtpm

`--tpm2-pcrs=7` seals against PCR 7 **at enrollment time — which happens
inside the live installer**. The installed system then boots a different
chain, and even PCR 7 (Secure Boot state) can differ between the live ISO
boot and the installed-disk boot (different loaders/measurements feed it).
So the sealed key can't unseal on first boot → passphrase fallback. This
is independent of swtpm restart handling; it's inherent to binding PCRs
during a live-env install.

This is a known bootc/tpm2-luks pitfall. The correct fixes are fisherman-
side, in order of robustness:
1. **First-boot enrollment** — install with a passphrase, and a oneshot
   on the *installed* system runs `systemd-cryptenroll --tpm2-pcrs=7`
   then removes the passphrase. PCR 7 is captured in the real target
   environment → always matches. (This is how most bootc installers do
   TPM LUKS.)
2. **Signed PCR policy** (`--tpm2-public-key`) — predictive sealing that
   survives firmware/loader updates; more setup.
3. **CI test mode** — enroll `--tpm2-pcrs=` (no binding) so the E2E
   verifies the unlock *mechanism* without the live-vs-installed PCR gap.

### Bottom line for community-variant LUKS coverage
The community-desktop LUKS test the builder ISOs need is **blocked by the
same fisherman enrollment issue that blocks flagship LUKS** — not by
anything desktop- or builder-specific. Install + encrypt + TPM-enroll is
green for every flavor; the reboot auto-unlock is the one shared gate.
Once fisherman moves to first-boot enrollment (option 1), the existing
LUKS flavor matrix (kde/cosmic/niri/xfce) gives community coverage for
free — the dev ISOs it builds ARE tacklebox/builder-engine ISOs.

Filing option 1 as the actionable fix.

hanthor:
Superseded by #686 — re-scoped onto current main (option 1). #679 had drifted to a 459-file / 10-conflict merge overlapping the already-merged #681/#684/#685; #686 brings only the unique LUKS two-boot TPM enrollment + offline-store work forward on a clean base (10 files, 77 bats tests green). Thanks #679 — most of your content already landed via earlier merges.

Issue #678, opened 2026-07-18
Installer GUI verification matrix: KDE / COSMIC / Niri / XFCE frontends must be provably usable


hanthor:
**Requirement (James):** verification that the bootc-install frontends for kde, cosmic, niri and xfce actually work — GUIs usable, with all the features they need. Backend coverage (fisherman, LUKS+TPM) exists via the LUKS E2E; the *frontends* have none.

## Phased plan

**Phase 1 — installer presence & autostart evidence (mechanical now):**
per-flavor live boot in CI → autologin (livesys baseline) → assert the installer autostarts → screenshot as artifact. Vehicle: `iso-e2e.sh --app-launch` already boots + screenshots apps; add a matrix workflow over {kde, cosmic, niri, xfce} using the dev ISO build the LUKS workflow already does.

**Phase 2 — install-path per DE:**
extend the LUKS E2E matrix beyond yellowfin:kde to cosmic/niri/xfce flavors — same encrypted-install + reboot + unlock proof through each frontend's fisherman handoff (frontends escalate via `flatpak-spawn --host pkexec fisherman`, so the recipe-driven path is what the GUI exercises).

**Phase 3 — GUI usability assertions (VLM):**
screenshot-driven checks of each installer's key screens (disk pick, encryption toggle, summary, progress). The harness's VLM hook exists but points at a tailnet-only endpoint (`lemonade…ts.net`) that CI can't resolve — needs either tailscale in the runner, a hosted VLM key, or screenshot-diff baselines as an interim.

**Feature checklist to assert per frontend** (from the fisherman recipe surface): disk selection, filesystem choice, TPM2-LUKS encryption toggle, hostname, flatpak preload handoff, progress + completion → reboot prompt.

Refs: #543 (product quality driver), #668 (desktop contract), tacklebox#98 (in-repo install E2E).

hanthor:
## Phase 1 result: installer frontends verified for KDE, COSMIC, Niri ✅

Run [29635224340](https://github.com/tuna-os/tunaOS/actions/runs/29635224340) — per-desktop: dev ISO → OVMF boot to `TUNAOS_LIVE_READY` → SSH-assert the installer process is running → desktop screenshot.

| Desktop | Result | Evidence |
|---|---|---|
| KDE | ✅ | `flatpak run org.tunaos.InstallerKde` running |
| COSMIC | ✅ | installer process running |
| Niri | ✅ | installer process running |
[...]

PR #681, opened 2026-07-18
fix(live-iso): cosmic/niri greetd autologin — drop invalid type=login_manager

hanthor:
The installer-smoke screenshots (#678) caught cosmic stuck at its greeter and niri at a black TTY. Root cause: `type = "login_manager"` in the greetd config — an invalid key that makes greetd reject the whole config. Removed from both adapters; autologin is now a clean `initial_session`+`default_session`. Companion to tacklebox#105 (same fix in the live baseline).

Verifies via the hardened installer-smoke gate (now requires an active graphical session, not just a running process).


hanthor:
⚠️ **Not verified — holding.** Same as tacklebox#105: re-ran installer-smoke, cosmic still at greeter and niri still at TTY per the screenshots. The `type` removal didn't land a desktop. Needs interactive debug of the actual greetd/autologin path on cosmic/niri.

PR #683, opened 2026-07-18
fix(iso-builder): fix tag-only parsing, add OOM warning and custom relay URL

hanthor:
This PR introduces several usability and correctness improvements to the browser-based WASM ISO builder configurator:

1. **Default Namespace for Tag-Only Inputs:** Automatically prepends the `tuna-os/` organization namespace when the input image contains no slash `/` (e.g. `yellowfin:kde`). This fixes a bug where `parseImage()` would incorrectly parse tag-only inputs as the registry host.
2. **Browser Compatibility Warning:** Adds a prominent warning message in the UI when the browser lacks direct file writing/streaming capabilities (File System Access API, e.g. Safari and Firefox). This warns users that direct streaming is disabled, that the multi-gigabyte ISO will buffer in memory first, and advises using a Chromium-based browser to prevent Out-Of-Memory crashes.
3. **Custom Registry Relay URL Configuration:** Exposes a "Registry Relay URL" input field under the "Advanced" panel to allow developers to specify their own Cloudflare Worker CORS shims. This setting is also automatically preserved in shared url presets.
4. **Walkthrough screenshots refreshed** to match the new UI elements.


hanthor:
Superseded: the ISO builder moved to its own repo — **https://github.com/tuna-os/iso-builder** (deployed at iso.tunaos.org), extracted with history. Monorepo cleanup + pointer in #684. Closing this in favor of the standalone repo.

PR #824, opened 2026-07-25
fix(kde): autologin to PlasmaLogin, and stop unverified cells reporting green

hanthor:
## The KDE live ISO stops at a password prompt

`installer-smoke` run [29914643652](https://github.com/tuna-os/tunaOS/actions/runs/29914643652) photographed **four minutes of the greeter**, not the installer. The "9 distinct visual states / 8 transitions changed >500px" the harness credited were the greeter clock ticking.

### Root cause (verified, not inferred)

Plasma 6.6 renamed SDDM to PlasmaLogin.

1. `plasma-login-manager-6.6.4-1.el10_3` arrives as a plasma-group dependency at transaction step **687/843**; its scriptlet creates `/etc/systemd/system/display-manager.service → plasmalogin.service`.
2. It does **not** `Obsoletes: sddm`, so the later explicit `dnf install sddm` also lands — but its enable fails:
   ```
   Failed to preset unit: File '/etc/systemd/system/display-manager.service'
   already exists and is a symlink to /usr/lib/systemd/system/plasmalogin.service
   ```
   `safe_enable`'s `|| true` swallowed it.
3. Boot confirms `dm=plasmalogin.service` via `TUNAOS_DESKTOP_CONTRACT_FAIL reason=dm_mismatch expected=^sddm\.service$`.
4. `desktop-kde.sh` wrote autologin to `/etc/sddm.conf.d/`. `rpm -qlp plasma-login-manager` proves it reads **`/etc/plasmalogin.conf.d/`** → autologin silently did nothing → password prompt → no session → no installer.

> Note: the published `ghcr.io/tuna-os/yellowfin:kde` image (Jul 19) still ships plain sddm. The rename landed before the Jul 22 fresh build, so this is invisible if you diagnose from the published image.

### Fix

Resolve the DM **per image** rather than renaming across the board — Fedora/Debian/Ubuntu still ship real sddm.

- `desktop-kde.sh` writes autologin to whichever `conf.d` dirs exist, and now **fails the build** if neither does rather than shipping an ISO that stops at a password prompt. The `[Autologin]` schema is identical; PlasmaLogin is Wayland-only, so SDDM's `[General] DisplayServer`/`CompositorCommand` stay sddm-only.
- New `lib.sh kde_dm_unit()`; used by `40-services.sh` (both sites).
- `configure-desktop-runtime.sh` mirrors the existing xfce lightdm-or-greetd detection.
- Contract checks accept `^(sddm|plasmalogin)\.service$`: `verify-desktop-experience.sh`, `e2e-runtime-checks.sh`, `boot-gate.sh`, `verify-image-packages.sh` (new any-of group). Without these the cell stays red even with autologin fixed.

## Also: the false-green that hid this for weeks

Non-strict cells exist so Smithay-on-GPU-less can't fail on a blank frame it cannot render. But `enforced=strict` kept those checks out of the failure count, so such a cell **exited 0 and reported SUCCESS having proved nothing**.

niri did exactly that in the same run — TAP said **7 failed**, JSON said `"failures": 0`, the job said **success**, and the frames were 964-byte 4-colour greyscale with `advanced_transitions: 0`. Its own TAP self-diagnosed *"the bare desktop with no installer window mapped."*

`installer-walkthrough.py` now exits **2 (UNVERIFIED**, distinct from a regression) unless a named installer screen was actually matched. Pixel movement is explicitly **not** accepted as evidence — the kde greeter clock scored 8/8 transitions while showing no installer at all.

Replayed against the recorded run data:

| cell | before | after |
|---|---|---|
| niri | `success` | **exit 2 UNVERIFIED** |
| kde (broken) | fail | fail (unchanged) |
| kde (fixed) | — | pass |

## Scope / what this does not fix

cosmic, niri and xfce share a single unrelated cause: all three are greetd-based and there is no DRM render node on GitHub runners (`greetd: greeter exited without creating a session` → `start-limit-hit`; cosmic: `libEGL: failed to create dri2 screen`). That needs a GPU host and no code change — those cells will now correctly report UNVERIFIED instead of silently passing.

KDE's greeter rendered a full 1280x800 frame on plain `-vga virtio`, so **this cell is verifiable on GitHub runners as-is**.

### DIFF
--- a/.github/workflows/installer-smoke.yml
+++ b/.github/workflows/installer-smoke.yml
@@ -188,6 +188,14 @@ jobs:
             rc=${PIPESTATUS[0]}
             set -e
             sudo chown -R "$(id -u)" smoke-out 2>/dev/null || true
+            # exit 2 == UNVERIFIED: the cell neither advanced nor matched a
+            # screen, so it proved nothing. It used to exit 0 here and report
+            # SUCCESS, which is how a green niri hid weeks of installer
+            # breakage (run 29914643652). Distinguish it from a regression in
+            # the message, but do NOT let it pass.
+            if [ "$rc" = 2 ]; then
+              echo "::error title=${FLAVOR}: installer UNVERIFIED::the installer never advanced and no screen was detected — this cell proves nothing. Smithay compositors (niri/xfwl4/cosmic) need a DRM render node; run on a host with /dev/dri/renderD128."
+            fi
             exit "$rc"
           else
             echo "::warning::no QEMU monitor socket — walkthrough skipped"
--- a/.github/workflows/installer-smoke.yml
+++ b/.github/workflows/installer-smoke.yml
@@ -231,6 +231,36 @@ jobs:
             echo
             echo "=== windows the compositor actually has ==="
             $SSH 'pgrep -xl "cosmic-comp|niri|xfwl4|kwin_wayland" 2>/dev/null || pgrep -l -x cosmic-comp || pgrep -l -x niri || pgrep -l -x xfwl4 || pgrep -l -x kwin_wayland || echo "(no compositor matched)"; ps -eo pid,comm,args | grep -iE "[I]nstaller" || echo "(no installer process)"' 2>/dev/null || true
+            echo
+            # THE display manager, whatever it is — resolved from the
+            # display-manager.service alias rather than guessed by name.
+            #
+            # Everything above greps for greetd, so when a NON-greetd DM
+            # fails we captured nothing at all. Run 30160664460: the serial
+            # console said only "[FAILED] Failed to start plasmalogin.service"
+            # and the diagnostics dutifully reported "greetd journal: -- No
+            # entries --" plus "package greetd is not installed". The one log
+            # that would have explained the failure was never collected, and
+            # a full ~2h CI cycle produced no diagnosis.
+            echo "=== display manager: unit, config, journal ==="
+            $SSH 'sudo bash -s' <<'DMEOF' 2>/dev/null || true
+              set +e
+              dm=$(systemctl show -P Id display-manager.service 2>/dev/null)
+              echo "display-manager.service -> ${dm:-<unresolved>}"
+              systemctl status --no-pager -l "${dm:-display-manager.service}" 2>&1 | head -30
+              echo "--- journal for ${dm:-display-manager.service} ---"
+              journalctl -u "${dm:-display-manager.service}" -b --no-pager 2>&1 | tail -80
+              echo "--- autologin config actually present ---"
+              for d in /etc/sddm.conf.d /etc/plasmalogin.conf.d /etc/gdm/custom.conf \
+                       /etc/gdm3/custom.conf /etc/lightdm/lightdm.conf.d /etc/greetd/config.toml; do
+                [ -e "$d" ] || continue
+                echo "### $d"
+                if [ -d "$d" ]; then for f in "$d"/*; do [ -f "$f" ] && { echo "--- $f"; cat "$f"; }; done
+                else cat "$d"; fi
+              done
+              echo "--- sessions the image ships ---"
+              ls -1 /usr/share/wayland-sessions/ /usr/share/xsessions/ 2>/dev/null
+          DMEOF
           } > "$out" 2>&1
           echo "--- captured $(wc -l < "$out") lines ---"
           tail -30 "$out"
--- a/.github/workflows/installer-smoke.yml
+++ b/.github/workflows/installer-smoke.yml
@@ -180,6 +180,32 @@ jobs:
           # rendering is possible; elsewhere still record screens for the matrix.
           STRICT=""
           case "${FLAVOR}" in kde|cosmic|gnome) STRICT="--strict" ;; esac
+
+          # Put the INSTALLER in front before driving it.
+          #
+          # Desktops autostart their own welcome/tour app on first login, and
+          # it lands on top of the installer. In run 30166081962 the
+          # walkthrough spent all nine frames driving KDE's Welcome Center —
+          # its Skip/Next buttons, then the taskbar, then Kate — while the
+          # installer sat behind it the whole time, running but never seen.
+          #
+          # This is test setup, not a product change: the live ISO still ships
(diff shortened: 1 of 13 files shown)


hanthor:
## Examined: still needed, but roughly half of it has landed by other routes

Checked each piece against `main` rather than judging from the diff:

| piece | on main? | verdict |
|---|:--:|---|
| Contract accepts `^(sddm\|plasmalogin)\.service$` | **yes** | superseded — `verify-desktop-experience.sh` already has `require_any_unit sddm plasmalogin` |
| `desktop-kde.sh` autologin into `plasmalogin.conf.d` | **yes** | superseded — main already handles both conf.d dirs |
| `lib.sh kde_dm_unit()` | **no** | still needed |
| `40-services.sh` DM resolution | **no** | still needed — main still hardcodes `safe_enable sddm.service` at **two** sites (lines 40, 98) |
| `installer-walkthrough.py` exit 2 UNVERIFIED | **no** | still needed (bootproof's axis) |

So it is neither obsolete nor landable as-is: the parts that conflict are the ones already merged, which is very likely why it went DIRTY. A rebase that **drops the contract and desktop-kde.sh hunks** and keeps `kde_dm_unit` + `40-services.sh` + `configure-desktop-runtime.sh` + the walkthrough change should apply cleanly against my #921 work — I have not touched `40-services.sh` or `lib.sh`.

## It also caught a regression of mine, which is the useful part

Examining this found a live bug in my own #921 work. On the current published `yellowfin:kde`:

```
plasmalogin.service   present
sddm.service          present
/etc/systemd/system/display-manager.service -> plasmalogin.service
```

I had added an unconditional `ln -sf` forcing `display-manager.service` at the manifest's `display_manager`, to fix openSUSE (where the alias is held by `display-manager-legacy.service`, a sysconfig shim rather than a real DM). Because `kde.yaml` says `display_manager: sddm` and `sddm.service` exists, that force would have **repointed every EL10/Fedora KDE image away from PlasmaLogin to SDDM** — silently changing which greeter boots and re-creating exactly the autologin-vs-running-DM mismatch this PR fixes.

Nothing would have gone red: the contract accepts both units, so both outcomes look green.

Fixed in #921 (`1cd60335`) — the force is now scoped to absent / dangling / legacy-shim only, verified across all four states:

```
no alias               -> CLAIM
legacy shim (openSUSE) -> CLAIM
dangling target        -> CLAIM
plasmalogin (valid)    -> LEAVE
```

## Recommendation

Rebase and land the remaining half. The `40-services.sh` hardcoded `sddm.service` is the same failure class as the one I hit on openSUSE — `safe_enable` swallows the failure, so a PlasmaLogin image gets no DM enabled from that path and nothing reports it. That is worth landing on its own merits, independently of the autologin fix.

hanthor:
Follow-up to my earlier comment — the DM-wiring half is now landed elsewhere, so this branch needs less rebasing than it looks.

**Why not a rebase.** `git merge-tree` against main gives 10 conflict markers, and the conflicting files are `live-iso/common/src/desktop-kde.sh`, `installer-smoke.yml`, `boot-gate.sh`, `installer-walkthrough.py` and `installer-screens.yaml` — the live-ISO and installer axes, which other sessions are actively working. Resolving their work to land ~70 lines of DM wiring was the wrong trade, so I ported the DM hunks to #921 instead.

**Now on main / #921 — drop these when you rebase:**

| hunk | where it landed |
|---|---|
| contract accepts `sddm\|plasmalogin` (`verify-desktop-experience.sh`, `e2e-runtime-checks.sh`) | main |
| `desktop-kde.sh` plasmalogin.conf.d autologin | main — as an unconditional write to *both* dirs rather than detect-then-fail |
| `boot-gate.sh`, `verify-image-packages.sh` accept either DM | main, spelled `DM="sddm plasmalogin"` / `"sddm.service\|plasmalogin.service"` |
| `lib.sh kde_dm_unit()` | #921 (`d0f327dd`) |
| `40-services.sh` both `safe_enable sddm.service` sites | #921 |
| `configure-desktop-runtime.sh` kde resolution | #921 |
| `install-desktop.sh` resolution + graphical.target.wants de-conflict | #921 |
| `tests/bats/test_kde_display_manager.bats` | #921 |

**Still yours, unlanded:** `installer-walkthrough.py` exit 2 UNVERIFIED, `installer-smoke.yml`, `installer-screens.yaml`, and the `desktop-kde.sh` livesys-kde conventions from `676c81b9`. That should rebase cleanly once the above are dropped.

**Three notes on what I changed in transit.**

1. `install-desktop.sh` keeps your bare inline `_TD_DM="plasmalogin"` rather than calling `kde_dm_unit()`. `_TD_DM` has `.service` appended at each use, so the helper's suffixed return would produce `plasmalogin.service.service` — `safe_enable` swallows that, the `-f` guard is then false so nothing is force-linked, and the image ships with **no** DM enabled. There's now a bats case asserting the two stay separate.

2. Two of your bats cases redefined `kde_dm_unit` inline and asserted against their own copy — they passed with `lib.sh`'s version deleted. They now source the real `lib.sh` and point it at a fake tree via a `_KDE_DM_ROOT` override. Mutation-checked: removing the function fails them.

3. Your `boot-gate.sh` / `verify-image-packages.sh` cases pinned array-literal syntax that main spells differently, and the `desktop-kde.sh` "fails loudly" case asserted a branch main made unreachable. Retargeted at the requirement rather than the syntax.

**And the reason this mattered beyond your PR.** My own `1cd60335` scoped the openSUSE `display-manager.service` alias claim so it stops stomping an alias the distro already set — which fixed the alias and left `graphical.target.wants/sddm.service` still force-linked on EL10 KDE. That is exactly the two-DMs-racing-for-seat0 state your commit message quotes. Your `rm -f` of the sibling link is what ends it; without porting that hunk, #921 would have merged with the race live. Caught before merge, nothing published is affected.

Separately, the `40-services.sh` hardcode is filed as tunaOS#925 — it's a missing *enablement* rather than a missing package, so the desktop contract as written would not catch it even if it ran everywhere. Worth keeping visible independently of this PR.

hanthor:
Closing as superseded: the display manager / PlasmaLogin wiring hunks landed on main via #921 (d0f327dd / 40-services.sh / lib.sh).

PR #844, opened 2026-07-26
fix(live): drop OnlyShowIn so the installer actually autostarts

hanthor:
**This is the answer to "does the cosmic installer work".** Proven on real hardware today (himachal, DRM render node + KVM), not inferred.

## What a live cosmic ISO actually does

I built the ISO and booted it under QEMU with the virgl path, then kept the VM alive and inspected it:

```
=== compositor ===        1591 cosmic-comp                 ← running
=== DM ===                greetd.service / active / graphical.target
=== sessions ===          liveuser  seat0  tty1            ← autologin worked
=== flatpak list ===      org.tunaos.InstallerCosmic       ← installed
=== installer ===         NO-INSTALLER-PROC                ← never launched
```

The full COSMIC desktop is up — `cosmic-panel`, `cosmic-launcher`, `cosmic-workspaces`, `cosmic-settings-daemon`. 13/13 live-image smoke checks pass. **cosmic-comp has never been shown running before**; on GPU-less GitHub runners it can't start at all, which is why this was invisible.

So the compositor was never the problem. The installer was, and for a reason nothing in the build could surface.

## Root cause

`systemd-xdg-autostart-generator` turns each `/etc/xdg/autostart` entry into a unit whose `ExecCondition` is `systemd-xdg-autostart-condition "<desktop>"`, evaluated against `$XDG_CURRENT_DESKTOP` **in the user manager's environment**. `cosmic-session` does not export it:

```
$ systemctl --user show-environment | grep -i xdg_current
(empty)
```

So the unit is generated and then skipped:

```
app-org.tunaos.installer\x2dlive@autostart.service: Skipped due to 'exec-condition'.
Condition check resulted in app-org.tunaos.installer\x2dlive@autostart.service - Install TunaOS being skipped.
```

**The tell:** gnome-keyring's three entries and `com.system76.CosmicInitialSetup` were skipped in exactly the same way, in the same journal, at the same second. It isn't our entry that's malformed — it's the filter. `xdg-desktop-autostart.target` itself reports `active`, which is what makes this look fine from every angle except the one that counts.

## The fix

A live ISO runs exactly one desktop, so `OnlyShowIn` discriminates nothing and costs the entire feature. Dropped from **cosmic, kde and xfce**.

Setting `XDG_CURRENT_DESKTOP` in the session would also work, but that depends on each compositor's launcher doing it correctly. This doesn't depend on anything.

## Scope caveat, stated plainly

Only **cosmic** is verified this far in. kde and xfce both fail earlier — plasmalogin failing to start (#836 adds the diagnostics) and the xfwl4 theme panic (tunaos-packages#123) respectively — so I could not observe their autostart at all. They carry the identical `OnlyShowIn` pattern, so this is very likely masking the same defect there, but I have not seen it and am not claiming it.

## Verification

`bash -n` + `shellcheck -S error` on all three adapters; bats `test_build_scripts_remaining` 29/29.

### DIFF
--- a/live-iso/common/src/desktop-cosmic.sh
+++ b/live-iso/common/src/desktop-cosmic.sh
@@ -55,6 +55,20 @@ systemctl mask sleep.target suspend.target hibernate.target hybrid-sleep.target 
 
 # Auto-launch the TunaOS installer frontend in the live session.
 # The app is baked into the live squash by customize-live.sh (tacklebox live_customize).
+# NOTE: deliberately no OnlyShowIn=. systemd-xdg-autostart-generator turns
+# each entry into a unit whose ExecCondition is
+# `systemd-xdg-autostart-condition "<desktop>"`, evaluated against
+# $XDG_CURRENT_DESKTOP in the *user manager's* environment. cosmic-session
+# does not export it, so on a live cosmic ISO every OnlyShowIn entry is
+# generated and then skipped:
+#
+#   app-org.tunaos.installer\x2dlive@autostart.service: Skipped due to
+#   'exec-condition'  (verified on hardware, 2026-07-26)
+#
+# gnome-keyring and CosmicInitialSetup were skipped the same way, which is
+# the tell: it is not our entry that is wrong, it is the filter. A live ISO
+# runs exactly one desktop, so OnlyShowIn buys nothing and costs the whole
+# feature.
 mkdir -p /etc/xdg/autostart
 tee /etc/xdg/autostart/org.tunaos.installer-live.desktop <<'DESKEOF'
 [Desktop Entry]
@@ -62,5 +76,4 @@ Type=Application
 Name=Install TunaOS
 Exec=flatpak run org.tunaos.InstallerCosmic
 Icon=org.tunaos.InstallerCosmic
-OnlyShowIn=COSMIC;
 DESKEOF
--- a/live-iso/common/src/desktop-kde.sh
+++ b/live-iso/common/src/desktop-kde.sh
@@ -69,6 +69,20 @@ systemctl mask sleep.target suspend.target hibernate.target hybrid-sleep.target 
 
 # Auto-launch the TunaOS installer frontend in the live session.
 # The app is baked into the live squash by customize-live.sh (tacklebox live_customize).
+# NOTE: deliberately no OnlyShowIn=. systemd-xdg-autostart-generator turns
+# each entry into a unit whose ExecCondition is
+# `systemd-xdg-autostart-condition "<desktop>"`, evaluated against
+# $XDG_CURRENT_DESKTOP in the *user manager's* environment. cosmic-session
+# does not export it, so on a live cosmic ISO every OnlyShowIn entry is
+# generated and then skipped:
+#
+#   app-org.tunaos.installer\x2dlive@autostart.service: Skipped due to
+#   'exec-condition'  (verified on hardware, 2026-07-26)
+#
+# gnome-keyring and CosmicInitialSetup were skipped the same way, which is
+# the tell: it is not our entry that is wrong, it is the filter. A live ISO
+# runs exactly one desktop, so OnlyShowIn buys nothing and costs the whole
+# feature.
 mkdir -p /etc/xdg/autostart
 tee /etc/xdg/autostart/org.tunaos.installer-live.desktop <<'DESKEOF'
 [Desktop Entry]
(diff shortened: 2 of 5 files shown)


hanthor:
## Verified on the live VM — before/after, with the mechanism visible

Applied the change to a running unmodified cosmic live ISO and re-generated the units.

**Before** (`OnlyShowIn=COSMIC;` present) — generated unit carries the condition, and is skipped:

```
app-org.tunaos.installer\x2dlive@autostart.service: Skipped due to 'exec-condition'.
$ systemctl --user is-active app-org.tunaos.installer\\x2dlive@autostart.service
inactive
$ pgrep -af "[t]una-installer-cosmic"
NOT-RUNNING
```

**After** (`OnlyShowIn` stripped, `daemon-reload`) — the regenerated unit has **no `ExecCondition` at all**, only `ExecStart`:

```
$ cat /run/user/1000/systemd/generator.late/app-org.tunaos.installer\\x2dlive@autostart.service
Description=Install TunaOS
ExecStart=:/usr/bin/flatpak run org.tunaos.InstallerCosmic

$ ls /run/user/1000/systemd/generator.late/xdg-desktop-autostart.target.wants/
... app-org.tunaos.installer\x2dlive@autostart.service      ← still wanted by the target

$ systemctl --user start app-org.tunaos.installer\\x2dlive@autostart.service
active
$ pgrep -af "[t]una-installer-cosmic"
3161 /usr/bin/bwrap --args 67 -- tuna-installer-cosmic
3183 /usr/bin/bwrap --args 67 -- tuna-installer-cosmic
```

The installer launches and stays up.

**What is inferred rather than observed:** that on a *fresh* boot the target pulls the now-unconditional unit automatically. The unit is in `xdg-desktop-autostart.target.wants/` with no condition left to fail, so this is ordinary systemd behaviour — but a live ISO can't persist an `/etc` edit across reboot, so confirming it end-to-end needs a rebuilt ISO. Worth doing before this is treated as closed.

## A correction to my own method

An earlier check of mine reported the installer as `RUNNING` when it wasn't. `pgrep -af tuna-installer-cosmic` run via `bash -c` **matched its own command line** — the exact self-match trap that `installer-smoke.yml` already documents and defends against with the `[o]rg` bracket trick. Every result above uses the bracket-proof `[t]una-installer-cosmic`.

Flagging it because the same false positive would make a CI gate report a passing installer on a desktop that never launched one — which is worth checking for elsewhere in the harness.

hanthor:
## Verified end-to-end on a rebuilt ISO — the inferred step is now observed

Rebuilt the cosmic ISO from this branch and booted it clean. **No manual intervention of any kind:**

```
=== OnlyShowIn in shipped autostart entry (expect 0) ===
0
=== compositor ===
1584 cosmic-comp
=== autostart unit ===
active
=== installer process ===
1873 /usr/bin/bwrap --args 67 -- tuna-installer-cosmic
2562 /usr/bin/bwrap --args 67 -- tuna-installer-cosmic
2565 tuna-installer-cosmic
```

The unit goes `active` on its own and the installer runs. The earlier caveat about the fresh-boot step being inferred is resolved.

## Getting here required fixing a cache bug that nearly produced a false negative

My first rebuild of this branch **still shipped `OnlyShowIn`**, and the ISO looked perfectly healthy. Cause: `tacklebox`'s customize cache key hashed only the named scripts, not the sourced siblings, so editing `desktop-cosmic.sh` didn't change the derived-image tag:

```
>>> [customize] derived image cache hit for localhost/yellowfin:cosmic (localhost/tbox-live-custom:d379df962dc5255d)
```

Same tag as the pre-fix build, 52ms customize step, ISO built cleanly. Had I stopped there I'd have concluded this fix doesn't work.

Fixed in tuna-os/tacklebox#153, and the run above used it — the key changed and customize genuinely re-ran, which validates that PR in situ too.

## Side finding worth its own look

On the virgl path, QEMU `screendump` fails:

```
(qemu) screendump /work/iso-e2e-out/shot.ppm
Error: no surface
```

`-display egl-headless` exposes no surface to dump. That matters beyond this PR: `installer-walkthrough.py` and the screenshot steps in `installer-smoke.yml` are **built on `screendump`**, and the virgl path is exactly the one we need for cosmic/niri/xfwl4. So on a GPU host — the only host where those compositors run at all — the screenshot evidence would come back empty. Process-level assertions (as used above) still work. Not fixing it here, but it should be tracked before the smoke workflow is pointed at GPU hardware.

PR #847, opened 2026-07-27
docs: matrix status, regenerated daily from CI

hanthor:
Adds `docs/MATRIX-STATUS.md`, linked from the docs index.

Every cell cites the run that proved it, and **"never tested" is a first-class state** — visually distinct from pass and from fail.

That distinction is the whole point. Until now "no result" and "passing" looked identical from the outside, which is how three separate defects survived for weeks: absence of evidence read as evidence of absence.

## What it records (2026-07-27)

| Axis | State |
|---|---|
| **LUKS E2E** | 12 of 33 non-NVIDIA ISO cells green. **All 24 NVIDIA cells red** — one systemic issue, not 24 |
| **Installer smoke** | **4 of 57 cells have *ever* been tested** — 7% coverage |
| **Live overlay** | 37 tags; only `albacore-xfce` and `marlin-kde` still missing |

The 7% is the number I'd want a reader to leave with. For 53 of 57 shipped ISO combinations, nobody has ever confirmed the installer appears on screen.

## It leads with scope, not results

Passing one axis says nothing about the others, and that repeatedly misled us. **LUKS E2E drives fisherman over SSH and never looks at the screen** — so `yellowfin:cosmic` was LUKS-green on 07-23 while its installer GUI had never once been observed. The doc says this before showing any table.

It also separates *product* failures from *harness* failures. cosmic reads ❌ in CI, but that's a GitHub-runner limitation with no software fallback: on a host with a DRM render node the desktop and installer both work, verified by hand and photographed on 07-26. Recording that as a plain failure would be wrong.

## Failure modes that make red look green

Collected in §5, because each one cost real time today and they'll recur:

- `pgrep -f` via `bash -c` matches its **own** command line — `installer-smoke.yml` documents the bracket fix; other call sites don't use it
- `gh pr checks` renders a **cancelled** run as `fail`
- after a partial `workflow_dispatch`, a bare tag may still point at the previous build — check `org.opencontainers.image.version`
- `install_available` skips unresolvable packages by design, so gaps surface only in a file inside the image that nothing reads on a schedule (three package issues found that way in one morning)

## Caveats, stated in the doc

Apart from `yellowfin:gnome`, **every LUKS result predates the 07-26 fixes** (#833, #836, #839, #844) — best available data, not current data. A full sweep on `main` is the highest-value thing that could be run against this page.

And it's assembled by hand, so it goes stale on the next merge. §6 has the exact queries behind every table; wiring them into a scheduled job that rewrites the tables would be a better use of an afternoon than maintaining it manually — and removes the risk of this page confidently asserting something that stopped being true a week ago.


hanthor:
## Now automated — the doc generates itself daily

The original version made the point that a hand-maintained status page goes stale on the next merge and ends up asserting things that stopped being true — the same failure it exists to prevent. So it now generates itself.

**`scripts/gen-matrix-status.py`** rewrites *only* the block between the `GENERATED` markers. The scope section, systemic gaps and failure-mode notes stay hand-written and untouched, so narrative can be added freely around the data.

**`.github/workflows/matrix-status.yml`** runs it at 06:00 UTC and pushes to `main` with `[skip ci]`, matching the convention the screenshot workflows already use. On `pull_request` it runs `--check` instead of pushing, so a hand-edit inside the generated block — or a `build-config.yml` change that breaks the generator — fails the PR.

## Four design decisions worth reviewing

**No wall-clock timestamp in the output.** An "as of `<now>`" line changes every run, which would make `--check` permanently fail *and* produce a daily commit even when nothing moved. The commit date already records when it ran. The file now changes **only when a cell actually flips**, which makes `git log docs/MATRIX-STATUS.md` a history of the matrix rather than a history of the cron job.

**A result always beats "not built".** My first version keyed the tables purely off `build_iso`, and silently dropped **sailfin** — which has `build_iso: false` everywhere yet *is* covered by LUKS E2E. Four tested cells vanished from a document whose entire purpose is to not lose track of what's tested. Caught by running the generator against real data and diffing it against the hand-built tables; worth noting because the bug and the document's thesis were the same shape.

**Only `success` and `failure` are read.** Cancelled and skipped runs are ignored, because `gh` reports a cancelled run in a way that reads as a failure — that misled me four separate times in one day.

**No `|| true` on the regenerate step.** If the API is unreachable the job fails loudly rather than rewriting every cell to ⬜. A status page that quietly degrades itself is worse than no status page, and would manufacture exactly the "absence of evidence" problem this is meant to kill.

## Verified

Run against live data: output matches the hand-built tables (12/37 LUKS green, 0/24 NVIDIA, 37 overlay tags, `albacore-xfce` + `marlin-kde` missing), and `--check` is idempotent immediately after a regenerate. `actionlint` and `yamllint` clean, bats 29/29.

PR #865, opened 2026-07-27
fix(build): only mount dnf caches into dnf-based variants

hanthor:
## What

`scripts/setup-build-cache.sh` bind-mounts `/var/cache/dnf`, `/var/cache/libdnf5` and `/var/lib/rpm` into **every** build container. Replaces the `grouper`-only denylist with an allowlist of the actually-dnf variants.

## Why

Those caches are inert for zypper/portage/pacman/apt — but they are still *mounts under `/var`*, and the bootc layout step later runs an unguarded `rm -rf /var`. `rm` cannot unlink a mountpoint, so it exits non-zero and `set -e` kills the build:

```
rm: cannot remove '/var/cache/libdnf5': Device or resource busy
rm: cannot remove '/var/cache/dnf': Device or resource busy
Error: building at STEP "RUN mkdir -p /sysroot/ostree ... rm -rf ... /var"
```

A Gentoo build has no dnf at all, yet was being handed a dnf cache.

This is the entire cause of 5 red cells in LUKS sweep [30249218614](https://github.com/tuna-os/tunaOS/actions/runs/30249218614): `sailfin` gnome/kde/niri/xfce and `guppy:kde`. None of them ever produced an ISO.

It stayed hidden because sailfin previously failed *earlier* — a zypper `addrepo` error on 2026-07-24 — so the `/var` wipe was never reached on that variant before.

## Verification

```
yellowfin       6 lines      sailfin       0 lines
albacore        6 lines      guppy         0 lines
skipjack        6 lines      marlin        0 lines
bonito          6 lines      flounder      0 lines
bonito-rawhide  6 lines      flounder-sid  0 lines
albacore-nvidia 6 lines      grouper       0 lines
```

Every currently-green LUKS cell is on the allowlist, so no passing variant loses its cache. `shellcheck` clean.

## Note on the allowlist direction

The previous denylist meant each new non-dnf variant silently inherited mounts it could not use. An allowlist fails safe — a new variant gets no cache until someone opts it in, which costs a slower build rather than a broken one.

---
<a href="https://app.blacksmith.sh/tuna-os/codesmith/tunaOS/pr/865"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>

### DIFF
--- a/scripts/setup-build-cache.sh
+++ b/scripts/setup-build-cache.sh
@@ -23,9 +23,22 @@ BASE_VARIANT="${BASE_VARIANT%-hwe}"
 BASE_VARIANT="${BASE_VARIANT%-kde}"
 BASE_VARIANT="${BASE_VARIANT%-dx}"
 
-if [[ "$BASE_VARIANT" == "grouper" ]]; then
-	exit 0
-fi
+# Only dnf-based variants get these mounts. The mounts land on /var/cache/dnf,
+# /var/cache/libdnf5 and /var/lib/rpm, which are meaningless to zypper, portage,
+# pacman and apt — but they are still *mounts under /var*, and the bootc layout
+# step later does an unguarded `rm -rf /var`. rm cannot unlink a mountpoint, so
+# it exits non-zero and `set -e` kills the build. That is why sailfin (zypper)
+# and guppy (portage) died at "Device or resource busy" on /var/cache/libdnf5 —
+# a Gentoo build has no dnf at all, yet was being handed a dnf cache.
+#
+# This was previously a denylist naming only grouper, which meant every non-dnf
+# variant added since then silently inherited mounts it could not use.
+# An allowlist fails safe: a new variant gets no cache until someone opts it in,
+# which costs a slower build rather than a broken one.
+case "$BASE_VARIANT" in
+yellowfin | albacore | skipjack | bonito | bonito-rawhide) ;;
+*) exit 0 ;;
+esac
 
 REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
 CACHE_BASE="${REPO_ROOT}/.rpm-cache/shared"


hanthor:
Possible second effect, flagged as a **hypothesis** rather than a claim — it needs a real run to confirm.

`marlin` (Arch) fails in the rechunk step with:

```
error: Unable to open sqlite database /proc/self/fd/4/var/lib/rpm/rpmdb.sqlite: unable to open database file
error: cannot open Packages index using sqlite - Operation not permitted (1)
Error: loading components
Caused by: 0: loading rpmdb  1: loading rpmdb from rootfs
```

Arch has no rpm and no `/var/lib/rpm` — **the only reason that path exists in the build container is the `--volume .../rpm:/var/lib/rpm:z` mount this PR stops emitting for marlin.** So the rechunker may be finding an empty, unreadable rpmdb directory that it would otherwise not see at all.

I am not asserting this fixes `marlin:gnome`/`marlin:kde`. It is equally possible the rechunker fails on Arch regardless of the mount, just with a different error. Worth re-running those two cells after merge to find out.

Unrelated and confirmed separately: `marlin:cosmic` fails at the #857 desktop contract gate (`cosmic.yaml has no .packages.pacman list`). That is the gate working as designed, not a regression, and is tracked as a manifest gap.

hanthor:
Update on the marlin hypothesis above — I have now traced the mechanism, and it is real but indirect.

The failing step is **chunkah**, not the build itself:

```
==> Running chunkah on flounder:gnome-pre-chunk...
error: cannot open Packages database in /proc/self/fd/4/var/lib/rpm
Caused by: 0: loading rpmdb  1: loading rpmdb from rootfs
```

chunkah reads the *committed image* (`--mount type=image`), so the cache mount's **contents** are irrelevant — but the mount still leaves an empty `/var/lib/rpm` **directory** committed in the image. A Debian or Arch image has no reason to contain that path at all; this PR is why it does.

So dropping the mount plausibly does fix marlin and flounder. I have not proven it, and I would rather not rely on it: #869 fixes the same nine cells directly by having chunkah skip any image with no rpmdb, and its probe treats an *empty* `/var/lib/rpm` as "no rpmdb" — so it is correct whether or not this PR lands first.

This PR still stands on its own for the EBUSY failures (sailfin ×4, guppy ×2, flounder:cosmic, flounder-sid:cosmic), which are unrelated to chunkah.

PR #904, opened 2026-07-30
fix(luks-e2e): pin root podman to a crun new enough for it

### LINKED ISSUE #903: luks-e2e: sudo podman picks up the old /usr/bin/crun (1.14.1) and every dev ISO build dies
## Symptom

Cells fail in **Build dev ISO** at the first `RUN`:

```
error running container: from /usr/bin/crun creating container for
[/bin/sh -c echo "${BUILD_SCRIPTS_HASH}" > /dev/null]: unknown version specified
```

Intermittent: `yellowfin:gnome` passed this step in run 30521033221 while `albacore:gnome` failed it in run 30522159277, same workflow, same runner image (`20260707.563`).

## Diagnosis

There are **two crun binaries on the runner**:

- `/usr/bin/crun` — Ubuntu's, **1.14.1** (apt reports `crun is already the newest version (1.14.1-1)`)
- something newer earlier in `PATH` — `crun --version` prints **1.28**

podman is **5.8.4** (not Ubuntu's 4.9.3 — something in the setup provides a much newer podman). podman 5.8 emits an OCI spec version that crun 1.14.1 rejects outright, hence "unknown version specified".

The ISO build runs under `sudo`. `sudo` resets `PATH` to `secure_path`, so `sudo podman` resolves `/usr/bin/crun` (1.14.1) even though an interactive `crun --version` in the same job reports 1.28. That is both why it fails and why it looks flaky — it depends on which podman/crun pair a given step actually resolves.

## Note on #902

I added `crun` to the apt line in #902 on a version-mismatch theory. That was wrong in mechanism — crun was already installed, so the change is a no-op for this bug (harmless, and the version echo it added is what made the 1.14.1-vs-1.28 split visible). It should not be mistaken for a fix.

## Candidate fixes

1. Point podman at the new runtime explicitly, e.g. `/etc/containers/containers.conf.d/`:
   ```
   [engine.runtimes]
   crun = ["/usr/local/bin/crun", "/usr/bin/crun"]
   ```
   (root-scoped, so it applies under sudo — the existing cgroupfs config is written to `~/.config/containers`, which `sudo podman` never reads.)
2. Or install a crun matching podman 5.8.4 over `/usr/bin/crun`.
3. Or adopt `projectbluefin/actions/bootc-build/setup-runner` with `update-podman: "true"`, which is what `reusable-build-image.yml` uses and why that workflow builds images fine on the same runners.

Worth noting the existing cgroup-manager workaround in this file has the same root-vs-user config problem: it writes `~/.config/containers/containers.conf`, which a `sudo podman` build does not read.

## Impact

Blocks the LUKS matrix (#763). Cells that get past it produce real results — `yellowfin:gnome` completed a full `tpm2-luks` install + unlock verification in run 30521033221.

hanthor:
Closes #903.

## The failure

Four of eleven cells in run [30522159277](https://github.com/tuna-os/tunaOS/actions/runs/30522159277) died in **Build dev ISO**, before LUKS was exercised at all:

```
error running container: from /usr/bin/crun creating container for
[...]: unknown version specified
```

## Root cause

podman on these runners is **5.8.4** and generates an OCI config that Ubuntu's **crun 1.14.1** rejects outright.

A newer **crun 1.28** *is* present and first in the interactive PATH — the job's own `crun --version` prints it. But the ISO build runs under **sudo**, and sudo resets PATH to `secure_path`, so podman resolves `/usr/bin/crun` and gets 1.14.1.

That also explains the apparent flakiness: whether a cell failed depended on which pair a given step resolved, **not on the image**.

| cell | outcome |
|---|---|
| `yellowfin:gnome`, `bonito:gnome` | ✅ full `tpm2-luks` install + unlock |
| `albacore`, `guppy`, `grouper`, `flounder` | ❌ this bug |
| `sailfin:gnome` | ❌ unrelated — `/usr/local` symlink breaks `mkdir -p` in the ISO recipe |

## The fix

Select the newest crun actually present and pin **root's** podman to it by absolute path via `/etc/containers/containers.conf.d/99-crun-path.conf`.

Root scope is the point. The existing cgroup-manager workaround writes `~/.config/containers/containers.conf`, which a `sudo podman` never reads — the same blind spot, worth fixing separately.

## On diagnosability

My previous attempt (#902) was **wrong about the mechanism** — it added `crun` to the apt line on a version-mismatch theory, but crun was already installed, so it was a no-op. Its only real contribution was the version echo that exposed the 1.14.1-vs-1.28 split.

This one prints every candidate crun with its version, the chosen one, and root podman's resolved runtime afterwards. If it's still wrong, the log says which cruns existed and which won, instead of leaving another bare crun error to interpret.

---
<a href="https://app.blacksmith.sh/tuna-os/codesmith/tunaOS/pr/904"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>

### DIFF
--- a/.github/workflows/luks-e2e.yml
+++ b/.github/workflows/luks-e2e.yml
@@ -230,24 +230,41 @@ jobs:
       # by absolute path. The candidate list is printed either way, so if this
       # is still wrong the log says which cruns existed and which was chosen
       # instead of leaving another crun error to interpret.
-      - name: Point root podman at a crun new enough for it
+      # podman on these runners is not a fixed version: some jobs get
+      # Ubuntu's 4.9.3, others get 5.8.4 from the setup actions. crun is
+      # always Ubuntu's 1.14.1, which 4.9.3 is happy with and 5.8.4 is not:
+      #
+      #   error running container: from /usr/bin/crun creating container for
+      #   [...]: unknown version specified
+      #
+      # That is the whole "flakiness" — albacore failed in run 30522159277
+      # under podman 5.8.4 and passed in run 30524767610 under 4.9.3, same
+      # workflow, same runner image. Selecting among the cruns already present
+      # cannot fix it, because when 5.8.4 lands 1.14.1 is the only one there.
+      #
+      # So install a crun that satisfies the newer podman and pin root's
+      # podman to it by absolute path. Root scope matters: the ISO build runs
+      # under sudo, and the existing cgroup-manager workaround writes
+      # ~/.config/containers/containers.conf, which sudo podman never reads.
+      #
+      # Versions and the resolved runtime are printed either way, so if this
+      # is still wrong the log says what was installed and what podman
+      # actually picked.
+      - name: Install a crun matching podman and pin root podman to it
+        env:
+          CRUN_VERSION: "1.28"
         run: |
           set -euo pipefail
           echo "podman: $(command -v podman) — $(podman --version)"
-          best=""; bestv=""
-          for c in $(type -a -P crun 2>/dev/null || true) /usr/local/bin/crun /usr/bin/crun /usr/sbin/crun; do
-            [ -x "$c" ] || continue
-            v=$("$c" --version 2>/dev/null | head -1 | grep -oE '[0-9]+(\.[0-9]+)+' | head -1) || true
-            [ -n "$v" ] || continue
-            echo "  candidate: $c -> $v"
-            if [ -z "$bestv" ] || [ "$(printf '%s\n%s\n' "$bestv" "$v" | sort -V | tail -1)" = "$v" ]; then
-              best="$c"; bestv="$v"
-            fi
-          done
-          [ -n "$best" ] || { echo "::error::no crun found on this runner"; exit 1; }
-          echo "chosen: $best ($bestv)"
+          echo "distro crun: $(crun --version 2>/dev/null | head -1 || echo none)"
+
+          sudo curl -fsSL -o /usr/local/bin/crun \
+            "https://github.com/containers/crun/releases/download/${CRUN_VERSION}/crun-${CRUN_VERSION}-linux-amd64"
+          sudo chmod +x /usr/local/bin/crun
+          echo "installed: $(/usr/local/bin/crun --version | head -1)"
+
           sudo mkdir -p /etc/containers/containers.conf.d
-          printf '[engine]\nruntime = "crun"\n\n[engine.runtimes]\ncrun = ["%s"]\n' "$best" \
+          printf '[engine]\nruntime = "crun"\n\n[engine.runtimes]\ncrun = ["/usr/local/bin/crun"]\n' \
             | sudo tee /etc/containers/containers.conf.d/99-crun-path.conf
           echo "root podman now reports:"
           sudo podman info --format '  runtime={{.Host.OCIRuntime.Path}} version={{.Host.OCIRuntime.Version}}'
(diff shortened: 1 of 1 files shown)


hanthor:
## Revised — the first version passed validation for the wrong reason

I validated the original step on `albacore` (the cell that actually failed) and it went green, including a full LUKS install + unlock. But the diagnostic it printed shows the fix wasn't what did it:

```
podman: /usr/bin/podman — podman version 4.9.3
  candidate: /usr/bin/crun -> 1.14.1
  candidate: /bin/crun     -> 1.14.1
```

podman was **4.9.3** in that run, and 4.9.3 is perfectly happy with crun 1.14.1. The failing cells had podman **5.8.4**, which rejects it. Since 1.14.1 is the only crun Ubuntu ships, "pick the newest crun present" is a **no-op in exactly the case that breaks**.

## The actual variable is podman, not crun

| run | podman | crun | outcome |
|---|---|---|---|
| 30522159277 (`albacore`) | **5.8.4** | 1.14.1 | ❌ `unknown version specified` |
| 30524767610 (`albacore`) | **4.9.3** | 1.14.1 | ✅ full pass |

Same workflow, same runner image `20260707.563`. Some jobs get Ubuntu's podman, others get 5.8.4 from the setup actions — that's the whole "flakiness", and it's why this looked image-specific when it never was.

## What the step does now

Installs **crun 1.28** outright and pins root's podman to it by absolute path. Correct under either podman, rather than depending on which one the runner happens to hand out.

Root scope still matters: the ISO build runs under `sudo`, and the existing cgroup-manager workaround writes `~/.config/containers/containers.conf`, which `sudo podman` never reads.

Needs re-validation against a job that actually gets podman 5.8.4 — a pass under 4.9.3 proves nothing here, which is the trap I just fell into.

hanthor:
## Not ready to merge — still unvalidated under the podman that breaks

Two revalidation attempts, neither discriminating:

| run | cell | podman | outcome | proves |
|---|---|---|---|---|
| 30524767610 | albacore | **4.9.3** | full pass incl. LUKS | **nothing** — 4.9.3 works with crun 1.14.1 anyway |
| 30526884305 | guppy | — | **cancelled** | nothing |

The bug only appears when a job draws **podman 5.8.4**, which rejects crun 1.14.1. A pass under 4.9.3 is the same false signal that made the first version of this PR look correct.

**Merge criterion:** a `Build dev ISO` that succeeds in a job whose log shows `podman version 5.8.4`. The step prints the podman version and the installed crun for exactly this reason — check both before merging, don't take a green tick.

Re-dispatch with `gh workflow run luks-e2e.yml -f variant=<any> -f flavor=gnome --ref fix/luks-e2e-crun-runtime-path` until one draws 5.8.4; the podman version varies per job, so it may take a couple of attempts.

hanthor:
Superseded — closing. The crun handling this PR adds is already on main: #902 installs crun in the same transaction as podman, and #886 (merged) added the "Pin podman to the newest crun on the runner" step which selects the newest crun present and pins it system-wide via /etc/containers/containers.conf.d, covering root's podman under sudo. This PR also conflicts (DIRTY) after #886 landed, and its approach curls an unpinned crun binary from GitHub releases at CI time, which the landed approach avoids. Thanks — the diagnosis in the description was useful in confirming the landed fix covers the sudo/secure_path case.

Issue #777, opened 2026-07-23
asahi: EL10 package gap — OBS project for m1n1/u-boot-asahi/audio (skipjack, yellowfin, albacore, bluefin-lts)


hanthor:
The CentOS Hyperscale SIG packages-asahi repo has the 16k asahi kernel + dracut-asahi + update-m1n1 but is ~10 months stale and ships NO m1n1/u-boot-asahi/asahi-audio/speakersafetyd. Plan: OBS (build.opensuse.org, free, builds RPM+deb+Arch on real aarch64 workers with signed repos) project building the missing pieces for el10 aarch64 — adapt specs from Fedora (@asahi COPRs) and OBS home:mrkcee. Needs James's OBS account to create the project. Until then the skipjack/EL asahi branch warns loudly and cannot boot hardware.

hanthor:
**This gap now has harness proof, and it shipped — `skipjack:gnome-asahi` was promoted despite this issue being open.**

The first full sweep of all promoted `:gnome-asahi` tags (#776) put yellowfin and skipjack in the same bucket with **byte-identical failure sets**:

| variant | result | run |
|---|---|---|
| yellowfin | 29 passed, 7 failed | [30550215291](https://github.com/tuna-os/tunaOS/actions/runs/30550215291) |
| skipjack | 29 passed, 7 failed | [30550239433](https://github.com/tuna-os/tunaOS/actions/runs/30550239433) |

```
FAIL /usr/bin/update-m1n1 missing
FAIL update-m1n1 kernel hook missing
FAIL /usr/bin/asahi-fwupdate missing
FAIL /usr/bin/tiny-dfr missing
FAIL /usr/lib/systemd/system/tiny-dfr.service missing
FAIL /usr/bin/asahi-diagnose missing
FAIL apple-isp.ko not in modules.dep
```

Everything else is right on both: the asahi kernel, the 16K page size, Apple DTBs, the m1n1 **and** U-Boot payloads, the dracut firmware flow, and the full audio stack all pass. This is precisely and only the EL10 userspace gap this issue describes — nothing more, nothing less.

**Why it is worse than a missing convenience tool.** `update-m1n1` and its kernel install hook are what refresh the m1n1 stage-2 payload on the ESP when the kernel changes. Without them these images can boot once and then **break on the first `bootc upgrade`** — the deployed kernel moves and the boot chain does not follow. For an image-based OS whose entire premise is upgrading, that is a shipping blocker rather than a rough edge.

[...]

Issue #776, opened 2026-07-23
asahi: validate first gnome-asahi images (bonito, grouper) + triage overlay fallout


hanthor:
First builds from the merged asahi wiring (#773, #774). Steps: watch the dispatched Build Bonito run (30004082336, flavor=gnome-asahi); on success run the Verify Asahi image workflow against ghcr.io/tuna-os/bonito:gnome-asahi-testing (35-point harness in scripts/verify-asahi-image.sh); triage build-log warnings from build_scripts/overlay/asahi.sh (package-name unknowns in UbuntuAsahi PPA / Hyperscale SIG paths); repeat for grouper gnome-asahi (stage-1 full build). Fix package names/paths until harness passes.

hanthor:
Milestone: **first TunaOS asahi image verified — 35/35 harness pass.**

- `bonito:gnome-asahi` built, manifested, and signed on run [30004082336](https://github.com/tuna-os/tunaOS/actions/runs/30004082336); published as `ghcr.io/tuna-os/bonito:gnome-asahi-testing`.
- Verify Asahi harness run [30037380244](https://github.com/tuna-os/tunaOS/actions/runs/30037380244): **35 passed, 0 failed**, 1 warning (`lsinitrd unavailable` on the runner — fixed by #790 so initramfs checks verify instead of skip).
- The only build failure was the promotion Gate: it's an x86 QEMU/KVM gate and the manifest is arm64-only — fixed by #788 (gate skips for flavors with no amd64 leg; the harness is the asahi gate). Rebuild dispatched post-merge: run [30038343741](https://github.com/tuna-os/tunaOS/actions/runs/30038343741) — on success the bare `:gnome-asahi` tag promotes for the first time.
- grouper gnome-asahi (stage-1 full build, UbuntuAsahi PPA path) dispatched: run [30038321475](https://github.com/tuna-os/tunaOS/actions/runs/30038321475).
- Installer payload build against the -testing image dispatched: [bootc-installer-asahi run 30038319060](https://github.com/tuna-os/bootc-installer-asahi/actions/runs/30038319060).

🤖 Posted by Claude Code (session: https://claude.ai/code/session_018bGymcRkVtV7DPToWDjZyZ)

hanthor:
[...]

PR #910, opened 2026-07-30
ci(asahi): sweep every flavor on a schedule, and grade the asahi kernel

hanthor:
The Asahi harness was dispatch-only. Nothing re-checked a promoted image, so "green" decayed silently — the last run before today was **2026-07-24**, six days of `main` ago.

I swept all eight promoted `:gnome-asahi` tags. **Six of the eight are unbootable** — full matrix and root-cause breakdown in #776. A tag existing in ghcr only means some build promoted it once; it is not evidence the image is correct.

### What this changes

- Splits the single-image job into a reusable `verify-asahi-one.yml`, driven two ways:
  - **sweep** — daily schedule (or dispatch with an empty image), one matrix leg per flavor across all eight
  - **triage** — dispatch with an explicit ref, as before, for iterating on one variant or a `-testing` tag
- `fail-fast: false`, so one red flavor cannot hide the other seven.
- The legs are deliberately **not** `continue-on-error`. A red leg has to fail the run — job-level `continue-on-error` makes run conclusions lie, which is a landmine this repo has already paid for once.
- Each leg writes its score and failing checks to `$GITHUB_STEP_SUMMARY`, so the red/green matrix is readable without opening raw logs.
- `guppy` is deliberately absent from the matrix — its overlay is stubbed, not broken.

### Harness bug the sweep exposed

Grading used `kvers[0]` — the first entry `ls` returns — not the asahi kernel. albacore ships a correct asahi kernel (`6.16.4-0.hs100.hs+asahi.el10.aarch64+16k`) alongside an unremoved stock one, and `6.12.0-…el10_2` sorts first. So every subsequent check graded the *stock* kernel, and 24 cascading failures buried the single real defect. albacore's honest score is one failure — "two kernels installed" — not 25.

Two kernels stays a hard failure (ambiguous boot is a real defect); it just no longer decides which kernel the rest of the harness reads.

Verified: `bash -n` on the harness, YAML parses, and the selection loop unit-checked against the two-kernel, single-asahi, single-non-asahi, and empty cases.

This also delivers RFC [bootc-installer-asahi#6](https://github.com/tuna-os/bootc-installer-asahi/issues/6) §4 (published red/green matrix).

Refs #776, #777.

https://claude.ai/code/session_01VvyH2muGbDWLmmmtrPNkwG

---
<a href="https://app.blacksmith.sh/tuna-os/codesmith/tunaOS/pr/910"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>

### DIFF
--- a/.github/workflows/verify-asahi-one.yml
+++ b/.github/workflows/verify-asahi-one.yml
@@ -1,0 +1,64 @@
+name: Verify Asahi image (one)
+
+# Reusable single-image leg of the Asahi golden-manifest harness. Called by
+# verify-asahi.yml, once per flavor in sweep mode and once for an explicit ref
+# in triage mode. Not meant to be dispatched directly — use "Verify Asahi
+# image", which drives this.
+
+on:
+  workflow_call:
+    inputs:
+      image:
+        description: "Asahi image ref to verify"
+        required: true
+        type: string
+
+permissions:
+  contents: read
+  packages: read
+
+jobs:
+  verify:
+    runs-on: ubuntu-24.04-arm
+    timeout-minutes: 30
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: Install lsinitrd
+        # dracut-core provides lsinitrd; without it the harness skips the
+        # initramfs-content checks with a warning instead of verifying them.
+        run: sudo apt-get update -y && sudo apt-get install -y --no-install-recommends dracut-core
+
+      - name: Login to GHCR
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: echo "$GH_TOKEN" | podman login ghcr.io --username "${{ github.actor }}" --password-stdin
+
+      - name: Verify image against Asahi golden manifest
+        id: verify
+        run: |
+          set -o pipefail
+          ./scripts/verify-asahi-image.sh "${{ inputs.image }}" 2>&1 | tee /tmp/verify.log
+
+      - name: Summarize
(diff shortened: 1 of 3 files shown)


hanthor:
**Guard verified empirically, and it needed the fix.**

The original `if: inputs.image == ''` relied on a null `inputs` context comparing equal to `''` on the `schedule` path. If that ever evaluated false, *both* jobs would skip and the workflow would report **success having verified nothing** — a daily sweep that looks like coverage and checks zero images. That is the same lying-conclusion failure this workflow exists to avoid, arriving through the back door. Now explicit: `github.event_name == 'schedule' || inputs.image == ''`.

Confirmed by dispatching from this branch with an empty image field — [run 30551057711](https://github.com/tuna-os/tunaOS/actions/runs/30551057711):

```
legs started: 9
  sweep (bonito) / verify      in_progress
  sweep (grouper) / verify     in_progress
  sweep (albacore) / verify    in_progress
  sweep (yellowfin) / verify   in_progress
  sweep (sailfin) / verify     in_progress
  sweep (flounder) / verify    in_progress
  sweep (marlin) / verify      in_progress
  sweep (skipjack) / verify    in_progress
  single                       skipped     ← correct
```

Eight legs, one per flavor, and the triage job correctly skipped. That run is expected to end **red** (6 of 8 images are broken — #776), which is the point: the legs are not `continue-on-error`, so a broken fleet fails the run rather than reporting a comfortable green.

Issue #915, opened 2026-07-30
flounder:niri and flounder-sid:niri are published with no niri binary — niri is not packaged for apt anywhere


hanthor:
## What

`ghcr.io/tuna-os/flounder:niri` is published and installable, and contains **no niri binary at all**:

```
$ podman run --rm --entrypoint /bin/sh ghcr.io/tuna-os/flounder:niri -c "command -v niri || echo NO-NIRI"
NO-NIRI
```

The desktop contract confirms it:

```
flounder:niri -> exit=1  missing required command: niri
```

`flounder-sid:niri` ships from the same manifest and is affected identically.

## Why

`manifests/desktops/niri.yaml` declares package lists for `fedora`, `el10`, `zypper` and `emerge` — **there is no `apt` section**. `install-desktop.sh` guards its install on a non-empty list, so on a Debian base it installs nothing and continues.

This is not a missing translation. **niri is not packaged for apt at all**, verified against all three relevant suites:

| suite | `niri` | `xwayland-satellite` |
|---|---|---|
| Debian 13 trixie | absent | absent |
| Debian sid | absent | absent |
| Ubuntu 26.04 | absent | — |

Everything else a niri session needs *is* available in trixie (`greetd` 0.10.3, `waybar`, `fuzzel`, `alacritty`, `swaylock`, `swayidle`, `mako-notifier`, `xdg-desktop-portal-gnome`/`-gtk`). The compositor itself is the only gap — and it is the one that matters.

## Why it shipped

The published tag dates to **2026-07-13**. `Containerfile.debian:230-232` does run `install-desktop.sh niri`, which invokes the contract gate, so a rebuild today would fail correctly. The image predates enforcement and has been sitting published since.

Related: the comment at `build_scripts/desktop/install-desktop.sh:100-102` records that this exact class already happened once — *"every apt flavor shipped a desktop-less image that still passed CI."*

## Why it matters now

A user who picks the Debian niri edition installs a system with no compositor. That is the worst failure mode available: not a build error, an install.

## Options

1. **Drop niri from the Debian/Ubuntu build matrix and de-publish the tags.** Honest and immediate; niri remains available on the Fedora/EL/openSUSE/Arch/Gentoo variants where it is packaged.
[...]

PR #921, opened 2026-07-31
fix(sailfin): give KDE, XFCE and niri a usable desktop, and fix the DM wiring that hid it

hanthor:
Fixes package parity for the three sailfin desktops that #917 did not cover, plus
the installer and gate bugs that measuring them exposed.

Rebased on main (includes #917 and #918).

## The lists

| desktop | before | after | resolves to |
|---|---|---|---|
| kde | 3 | 47 | 741 pkgs |
| xfce | 3 | 37 | 280 pkgs |
| niri | 2 | 35 | 236 pkgs |

**The GNOME skeleton theory did not repeat.** `patterns-kde-kde` and the XFCE
patterns both resolve to a *real* session — plasma6-desktop/workspace/kwin6/
systemsettings6/kscreen6/powerdevil6/kwallet/portal-kde6, and xfce4-session/
panel/settings/xfdesktop/thunar. What both lacked was the application and
integration layer. Verified on the published images: `sailfin:kde` had
plasmashell but no dolphin, konsole, kate or ark, and **no plasma6-nm and no
plasma6-pa — the panel had neither a network nor a volume applet**.
`sailfin:xfce` had no terminal, no tumbler (no thumbnails), no gvfs (no
removable media) and **no xdg-desktop-portal at all** (broken Flatpak file
dialogs). Only niri was genuinely bare: `niri` + `greetd` and nothing else.

Every name verified with `zypper se -x -t package`, and each list dry-runs
clean — exit 0, no unresolved names, **no vendor changes**, no removals —
against a probe built to match `Containerfile.opensuse` (`pattern
enhanced_base` + packman-essentials at priority 90). Plain Tumbleweed is not a
sufficient test bed: packman ships rival multimedia builds, and a package that
forces a vendor change makes `zypper install -y` exit non-zero, which under
`set -euo pipefail` fails the build.

## Two bugs the gate could not have caught

**1. `sailfin:xfce` had no display manager enabled at all.** The top-level
`display_manager: gdm` is right for Fedora/EL10, but openSUSE's XFCE is
lightdm, and a list-shaped zypper section had nowhere to say so. Nothing
failed: `safe_enable` no-ops on a missing unit, the `graphical.target.wants`
link is guarded by the unit file existing, and `set-default graphical.target`
ran regardless. Verified on the published image —
`/etc/systemd/system/graphical.target.wants/` did not exist. The zypper
section can now be map-shaped (`display_manager` + `packages`), same as apt.

**2. `systemctl enable <dm>` does not partially succeed on openSUSE — it
fails.** `displaymanager-sysconfig` already owns
`/etc/systemd/system/display-manager.service`, pointing it at its own
`display-manager-legacy.service`, and systemd refuses to alias over an
existing symlink:

```
Failed to enable unit: File '/etc/systemd/system/display-manager.service'
already exists and is a symlink to .../display-manager-legacy.service
```

`safe_enable` swallows that, so enable creates **nothing** — not the alias,
not the wants link. The forced wants-link from b6c6f35 was doing all the work,
and the alias still resolved to the legacy launcher, so the runtime contract
(`systemctl show -P Id display-manager.service`) would have failed a *working*
desktop while the legacy launcher was free to start a second DM. The installer
now forces the alias too.

Both are invisible to the gate as written: it asserts a DM unit *file exists*,
never that one is *enabled*. Same "shipped and green" failure mode as
sailfin:gnome.

**3. Related, not fixed here: the contract never runs on apt variants.** The
apt path `exit 0`s before the shared tail, so no apt desktop is gated at all.
That is the mechanism that published `flounder:niri` with no compositor
(#915). This PR only removes the last stale declaration (grouper, missed by
#918); moving the contract ahead of the early exit needs its own measured
change, since it could redden flounder/grouper builds that have never been
gated.

## Contract extension

Rule used, because a build-time assertion that a manifest never asks for turns
a working build red: **a requirement must be explicitly declared in every
manifest section that builds that desktop.** So kde asserts `dolphin` and
`konsole` (in all seven KDE sections; measured present on yellowfin:kde) but
NOT portal-kde (absent from the emerge list, and guppy builds KDE) and NOT
plasma6-nm/kwallet (nowhere explicit). xfce asserts `thunar` + gtk portal but
NOT `xfce4-terminal` (Debian gets it via Recommends). niri asserts a portal
backend + `gnome-keyring-daemon`. **cosmic is deliberately untouched** —
`cosmic-files` comes from a COPR on el10 and copr installs are best-effort, so
asserting it would turn a flaky COPR into a hard build failure.

A new bats test cross-checks every contract-required application against the
manifests that must declare it, so the gate cannot drift ahead of the lists.

## Also

- **niri had no greeter.** greetd's stock config is `agreety --cmd /bin/sh` —
  a text prompt into a bare shell. Wired up gtkgreet+cage, reusing the block
  already proven in `xfce.sh`, with a boot-time fallback to software rendering
  when there is no DRM render node. Without that, cage cannot initialise GL on
  a virgl-less guest and greetd restart-loops on a black screen — that is
  GNOME Boxes, VirtualBox and any `-vga virtio` VM, which is how most people
  will first try TunaOS. The Gate caught this.
- **remora was never installed on sailfin**, so every sailfin Gate failed on
  `reason=remora_not_found` — including gnome, whose list nothing had touched.
  `Containerfile.opensuse` runs none of the numbered build_scripts (26-packages-post.sh
  would rebuild the initramfs with Fedora kernel naming and clobber the
  composefs/bootc one), so the remora install is split into its own script
  called from both paths.
- openSUSE/DMS wording corrected to "not yet a Tideforge target"
  (tunaos-packages#111) rather than "impossible" — quickshell/dms are becoming
  first-class Tideforge packages via tunaos-packages#125.

## Verification

Real CI on this branch: all four sailfin desktops built green (`linux-amd64`,
`Manifest`, `Sign`), the new lists installed with no vendor conflicts, and the
serial logs show `display-manager.service` resolving to `gdm`, `sddm`,
**`lightdm`** and `greetd` respectively — xfce previously had none.

Contract controls: an image built from the old 3-package KDE list now fails
with `exit=1 missing required command: dolphin`; yellowfin:kde and openSUSE
with the new list both satisfy it.

Local suite: 657 pass, 2 pre-existing qemu failures that also fail on main.

---
<a href="https://app.blacksmith.sh/tuna-os/codesmith/tunaOS/pr/921"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>


hanthor:
## Gate results — verified in CI (run 30594521039)

Three of four desktops now pass the full boot gate and were promoted:

| flavor | contract marker | `display-manager.service` resolves to | Gate | Promote |
|---|---|---|---|---|
| gnome | `TUNAOS_DESKTOP_CONTRACT_OK` | `gdm.service` | ✅ | ✅ |
| kde | `TUNAOS_DESKTOP_CONTRACT_OK` | `sddm.service` | ✅ | ✅ |
| **xfce** | `TUNAOS_DESKTOP_CONTRACT_OK` | **`lightdm.service`** | ✅ | ✅ |
| niri | — (never booted, see below) | — | ❌ | — |

**xfce is the one to look at**: it previously had *no* display manager enabled
at all, and `/etc/systemd/system/graphical.target.wants/` did not exist on the
published image. It now boots with lightdm as `display-manager.service`,
which also proves the second fix — the forced alias, since openSUSE's
`displaymanager-sysconfig` owns that symlink and `systemctl enable` fails
outright rather than partially succeeding.

`reason=remora_not_found` is gone everywhere. That had been failing all four
Gates, including gnome, whose package list nothing in this PR touches.

### niri's failure was not this PR

It failed at **"Build qcow2 from testing image"**, before the VM ever booted —
no boot evidence artifact was produced:

```
Error: OCI runtime error: crun: unknown version specified
```

Root-caused in c1c8685 (thanks): newer `ubuntu-latest` images ship podman
under `/usr/local`, so `apt-get install podman` layers Ubuntu's 4.9.3 on top
and the mixed stack cannot start a container. kde/xfce/gnome passed in the
same run only because their runners already had podman apt-installed, making
the install a no-op — it looks like a flake and is not.

That fix is now on this branch, and a fresh run on the rebased head
(30598079408) is in flight to confirm niri's Gate and re-validate everything
post-rebase.

PR #952, opened 2026-08-01
fix(build): install an SSH server on every package manager, not just rpm bases

### LINKED ISSUE #951: apt variants can never build a dev ISO: openssh-server is only installed when ENABLE_SSHD=1
## Symptom

Every LUKS cell for `flounder` and `flounder-sid` fails **before LUKS is attempted**, in *Build dev ISO*:

```
ERROR: dev ISO requested but no SSH service is installed
error: Recipe `iso` failed with exit code 1
```

Runs [30675951080](https://github.com/tuna-os/tunaOS/actions/runs/30675951080) (flounder:kde, :cosmic, :xfce) and [30675954952](https://github.com/tuna-os/tunaOS/actions/runs/30675954952) (flounder-sid:gnome, :cosmic), both from the 00:36 sweep.

This is **not** the storage-driver bug fixed in #949 — those cells never reach the install layer.

## Cause: an asymmetry between the rpm and apt paths

`build_scripts/40-services.sh:53-63`, apt branch:

```bash
# Security default: sshd closed (live ISOs may re-enable for dev).
if [[ "${ENABLE_SSHD:-0}" == "1" ]]; then
    apt-get install -y openssh-server
    safe_enable sshd.service
    ...
else
    safe_disable sshd.service
    ...
fi
```

On apt, `openssh-server` is **installed only when `ENABLE_SSHD=1`**. Otherwise the package is absent entirely.

On the rpm side the base images already ship `openssh-server`, and the same "security default" is achieved by *disabling the service* — the package still exists. That is why `yellowfin` and `albacore` build dev ISOs and the Debian variants cannot.

`live-iso/common/src/customize-live.sh:158-168` then requires a real unit file to exist, and is already Debian-aware — it deliberately falls through from `sshd.service` (a compat symlink on Debian/Ubuntu) to `ssh.service`. The check is correct; there is simply no package installed for it to find.

## Scope

Blocks the dev ISO, and therefore every LUKS and installer cell, for all apt-based variants: **flounder, flounder-sid, grouper**. That is a large share of the 30 never-tested LUKS cells the 08-01 00:36 sweep was dispatched to convert.

## Proposed

Install `openssh-server` unconditionally on the apt path and keep the *service* disabled by default, matching what the rpm path already does. "sshd closed" should mean the service is off, not that the package is missing — the current comment describes the rpm behaviour while the code does something stricter.

Verify by checking a built apt image has `/usr/lib/systemd/system/ssh.service` present and the unit disabled, then re-running one flounder LUKS cell.

hanthor:
Closes #951.

The rpm bases already ship `openssh-server`, so there "sshd closed by default" has always meant the **service** is disabled. The apt path gated the *package install* on `ENABLE_SSHD=1` — stricter than the comment beside it claims, and it breaks the dev ISO:

```
ERROR: dev ISO requested but no SSH service is installed
```

Every flounder and flounder-sid cell in the 00:36 LUKS sweep died there (runs `30675951080`, `30675954952`) before reaching the install layer. The dev ISO gates the installer axis too, so one packaging asymmetry was holding a large share of the never-tested cells on **both** goal axes — while presenting as a LUKS failure, which it is not.

`customize-live.sh` is not changed: its fallthrough from `sshd.service` (a compat symlink on Debian, which `systemctl enable` refuses to operate on) to the real `ssh.service` was already right. There was nothing installed for it to find.

### The part worth reviewing

**Disabling is now load-bearing, not belt-and-braces.** Debian's `openssh-server` postinst *enables* `ssh.service` on install. With the unconditional install, a disable that only knows the Fedora spelling would ship published flounder images **with sshd running** — a build fix turned into a security regression. So both spellings and both socket units are named, on the enable and disable sides alike.

### Tests

Two bats guards, each checked to **fail against the pre-fix file** rather than merely pass against the new one:

- the apt install must not be nested inside the `ENABLE_SSHD` branch
- the disable path must name `ssh.service` and `ssh.socket`, not just `sshd.*`

---
<a href="https://app.blacksmith.sh/tuna-os/codesmith/tunaOS/pr/952"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>

### DIFF
--- a/build_scripts/40-services.sh
+++ b/build_scripts/40-services.sh
@@ -51,16 +51,45 @@ if [[ "${PKG_MGR:-}" == "apt" ]]; then
 	safe_enable tunaos-live-ready.service
 
 	# Security default: sshd closed (live ISOs may re-enable for dev).
+	#
+	# openssh-server is installed UNCONDITIONALLY, matching the rpm path.
+	# The rpm bases already ship it, so on those images "sshd closed" has
+	# always meant the SERVICE is disabled. Gating the apt install on
+	# ENABLE_SSHD made the apt path do something stricter than the comment
+	# claimed — the package was absent entirely — and that is not a harmless
+	# extra bit of hardening: customize-live.sh:217 needs a real unit file to
+	# enable for a dev ISO, finds none, and aborts with
+	#
+	#   ERROR: dev ISO requested but no SSH service is installed
+	#
+	# Every flounder and flounder-sid cell in the 00:36 LUKS sweep died there
+	# (runs 30675951080, 30675954952) without reaching the install layer, and
+	# the dev ISO gates the installer axis too — so one packaging asymmetry
+	# was holding a large share of the never-tested cells on both axes.
+	apt-get install -y --no-install-recommends openssh-server
 	if [[ "${ENABLE_SSHD:-0}" == "1" ]]; then
-		apt-get install -y openssh-server
+		# Debian/Ubuntu ship the real unit as ssh.service, with sshd.service a
+		# compat symlink that `systemctl enable` refuses to operate on ("linked
+		# unit file"). safe_enable swallows that failure, so name both and let
+		# the one that exists win — the same fallthrough customize-live.sh does
+		# deliberately at :212-216.
 		safe_enable sshd.service
+		safe_enable ssh.service
 		if ! id liveuser &>/dev/null; then
 			useradd -m -s /bin/bash -G sudo liveuser
 		fi
 		echo 'liveuser:live' | chpasswd
 	else
+		# Now load-bearing rather than belt-and-braces: Debian's
+		# openssh-server postinst ENABLES ssh.service on install, so with the
+		# unconditional install above, a disable that misses the Debian unit
+		# names would ship published flounder images with sshd running. That
+		# would turn a build fix into a security regression, which is why both
+		# spellings and both socket units are named here.
 		safe_disable sshd.service
-		safe_disable sshd.socket 2>/dev/null || systemctl mask sshd.socket || true
+		safe_disable ssh.service
+		safe_disable sshd.socket
+		safe_disable ssh.socket
 	fi
 
 	# Units that exist on Ubuntu once their packages are installed.
--- a/build_scripts/40-services.sh
+++ b/build_scripts/40-services.sh
@@ -12,6 +12,48 @@ export MAJOR_VERSION_NUMBER
 
 source /run/context/build_scripts/lib.sh
 
+# Make sure an SSH server EXISTS on every variant, whatever it is packaged as.
+#
+# tunaOS#951. openssh-server was installed in exactly one place in this file —
+# the apt branch, under ENABLE_SSHD=1. Nothing installed it for pacman or
+# zypper at all. The rpm variants only ever worked because their BASE IMAGES
+# happen to ship the package, so "sshd closed by default" meant a disabled
+# SERVICE there and a MISSING PACKAGE everywhere else.
+#
+# The consequence was structural, not cosmetic: customize-live.sh:217 needs a
+# real unit file to enable when building a dev ISO, and aborts with
+#
+#   ERROR: dev ISO requested but no SSH service is installed
+#
+# The dev ISO gates every LUKS and installer cell, so flounder, flounder-sid,
+# grouper, marlin and sailfin were not under-exercised in the matrix — they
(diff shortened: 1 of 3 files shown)


hanthor:
## Scope: this fixes ONE of the two failure modes

**flounder / flounder-sid (build-time) — fixed here.** `openssh-server` was absent, `customize-live.sh:217` found no unit, the dev ISO aborted.

**grouper (runtime) — NOT fixed here, and the proposed cause is disproved.**

The hypothesis was that `safe_enable sshd.service` fails silently on Ubuntu because `sshd.service` is a compat symlink. I checked it two ways instead of assuming.

**1. Against a real image on a build host** (`ghcr.io/tuna-os/grouper:xfce`, Ubuntu 26.04). The symlink exists as documented — but the enable *succeeds*:

```
lrwxrwxrwx sshd.service -> ssh.service
$ systemctl enable sshd.service
Created symlink '/etc/systemd/system/multi-user.target.wants/ssh.service' → '/usr/lib/systemd/system/ssh.service'.
exit=0
```

`systemctl list-unit-files` also returns 0 for the present unit and 1 for an absent one, so `safe_enable`'s guard is behaving correctly too.

**2. Against grouper's own build log** (run `30675947459`, job `91303126208`), which settles it directly:

```
+ safe_enable sshd.service
+ systemctl enable sshd.service
Created symlink '/etc/systemd/system/sshd.service' → '/usr/lib/systemd/system/ssh.service'.
Created symlink '/etc/systemd/system/multi-user.target.wants/ssh.service' → '/usr/lib/systemd/system/ssh.service'.
```

The enable worked. `openssh-server` was installed (`ENABLE_SSHD=1` was exported and the package downloaded). And the guest serial shows the daemon **running**:

```
OK  Started ssh.service - OpenBSD Secure Shell server.
```

So grouper is a third, separate problem: **sshd comes up in the guest and the host cannot connect to it** — 25 × `ERROR: SSH check failed` over 60s while the daemon is up. All five grouper cells failed this way, not just xfce.

The `safe_enable ssh.service` line in this PR is therefore belt-and-braces, not the grouper fix. It stays because it is correct for the general apt case, but nobody should expect grouper to go green from it.

**Concrete next step for whoever takes grouper:** `ERROR: SSH check failed` discards ssh's own stderr, so the log cannot distinguish *connection refused* (port-forward) from *permission denied* (auth) from *connection reset*. That one-line diagnosability gap is why this looked like the flounder bug. Surfacing the real error is the cheapest thing that would move it.

hanthor:
## Widened past apt — the scope was bigger than the title

`openssh-server` was installed in exactly **one** place in `40-services.sh`: the apt branch, under `ENABLE_SSHD=1`. Nothing installed it for pacman or zypper at all. The rpm variants only ever worked because their **base images** ship it — so "sshd closed by default" meant a *disabled service* on rpm and a *missing package* everywhere else, while one comment claimed a single policy for all of them.

`ensure_openssh_installed()` now handles the per-manager names: Arch `openssh`, Gentoo `net-misc/openssh`, openSUSE `openssh-server` falling back to `openssh` (older bases have not split it), everything else `openssh-server`.

It is **presence-guarded on the unit file**, not unconditional — a no-op on rpm and gentoo bases that already ship a server. No extra build time, no behaviour change, and no new failure mode on images this was never broken for.

**Called from both branches deliberately.** The apt branch `exit 0`s, so a single call sited in either one silently skips the other family — that asymmetry *is* the original bug, not an incidental detail of it. A bats guard asserts two call sites and that neither is nested under an `ENABLE_SSHD` test.

### Why this matters beyond one red cell

The dev ISO gates every LUKS and installer cell. flounder, flounder-sid, grouper, marlin (`30675943360`) and sailfin were not *under-exercised* in the matrix — they were **unbuildable** for both goal axes. That, not variant maturity, is why LUKS coverage clusters entirely on the rpm family.

### Still true from the earlier comment

grouper has a **separate** runtime failure this does not fix: sshd starts in the guest and the host cannot connect. Verified in its build log and guest serial. Do not expect grouper to go green from this PR.

Both new guards were checked to fail against the pre-fix file, not merely to pass against this one. Rebased onto `4e675fa2`/`b2fffa5c` rather than force-pushed over them.

Issue #1015, opened 2026-08-05
LUKS E2E: marlin (Arch) — 40-services.sh lacks pacman/zypper/emerge path (fixed in fix/marlin-luks-green)


hanthor:
## Problem

When wiring the standardized base scripts (\`40-services.sh\`, \`90-image-info.sh\`, \`99-cleanup.sh\`) into \`Containerfile.arch\`, the Arch build fails because \`40-services.sh\` falls through to the dnf/RPM-specific code path which calls tools and edits files that don't exist on Arch:

1. \`sed -i ... /usr/lib/systemd/system/uupd.service\` — file doesn't exist
2. \`sed -i ... /usr/lib/systemd/logind.conf\` — may not exist  
3. \`authselect select sssd --force\` — command doesn't exist

Additionally:
- \`--mount=type=tmpfs,dst=/opt\` fails because \`/opt\` is a symlink (\`-> var/opt\`) on Arch/Debian/openSUSE/Gentoo after ostree layout setup

## Fix (committed on fix/marlin-luks-green)

1. Added dedicated \`pacman/zypper/emerge\` early-exit path in \`40-services.sh\` (matching the apt pattern)
2. Removed \`--mount=type=tmpfs,dst=/opt\` from all symlink-/opt Containerfiles
3. Added \`command -v authselect\` guard to the dnf path
4. Added \`IS_HUMMINGBIRD\`, \`IS_OPENSUSE\`, \`IS_GENTOO\` to lib.sh env cache + exports

## Status

Fix is on \`fix/marlin-luks-green\` — canary marlin:gnome run dispatched (31013418173), awaiting results.

## Affected Cells (before fix)

- marlin: ALL 10 cells (gnome, kde, cosmic, niri, xfce + 5 cachyos)

hanthor:
The `40-services.sh` pacman path fixed the build failure this issue describes, and the canary ([run 31013418173](https://github.com/tuna-os/tunaOS/actions/runs/31013418173)) got all the way to the branding gate. It then failed there with all ten fields upstream — so there's a **second** marlin blocker behind this one, and it isn't the sparse-keys problem that 4f139df9 fixed.

`90-image-info.sh` writes only `/usr/lib/os-release`. Arch ships a *separate real* `/etc/os-release`, not the conventional symlink — its `filesystem` package installs only `usr/lib/os-release`:

```
$ tar -tvf filesystem-2025.10.12-1-any.pkg.tar.zst | grep os-release
[...]

Issue #1008, opened 2026-08-05
LUKS E2E: KDE branding failure — LookAndFeelPackage not set to TunaOS theme (bonito, flounder)


hanthor:
## Problem

KDE cells fail the branding-kde contract with:

```
FAIL: LookAndFeelPackage=org.fedoraproject.fedora.desktop — not a TunaOS theme
TUNAOS_BRANDING_KDE_FAIL variant=bonito failures=1
```

On flounder it's:

```
FAIL: no SDDM/plasmalogin theme directory or config found — greeter not installed
TUNAOS_BRANDING_KDE_FAIL variant=flounder failures=1
```

## Root Cause

The TunaOS KDE LookAndFeel package (`org.tunaos.desktop`) is shipped in `system_files/usr/share/plasma/look-and-feel/` (added in the branding commit), but `kdeglobals` or the SDDM/plasmalogin config is either not installed or not pointing to the TunaOS theme on these variants.

## Affected Cells

- bonito: kde
- flounder: kde
- Likely also skipjack:kde, yellowfin:kde, albacore:kde (same failure class, masked by VERSION_CODENAME failure occurring first)

## Branch

`fix/marlin-luks-green`

---

## Resolution

Both halves are fixed and both are green. The two symptoms had different causes.

**flounder — "greeter not installed" was a false negative in the check, not a missing greeter.**
`verify-branding-kde.sh` looked only for a themes directory or a `sddm.conf.d`. Verified against the Debian trixie `sddm` package file list: it ships `/usr/bin/sddm`, `/usr/lib/systemd/system/sddm.service`, `/etc/pam.d/sddm`, `/usr/share/sddm/faces` and `/usr/share/sddm/flags` — and **no** `/usr/share/sddm/themes` and **no** `sddm.conf` or `sddm.conf.d` anywhere. flounder had a perfectly good greeter and the contract called it absent. The check now looks at the unit and the binary first, which is the load-bearing evidence, and treats themes and config as additional signal. It also covers the KDE 6.5+ rename to `plasma-login-manager` — both names are in the wild, so every candidate location is checked rather than detecting which DM is installed.

**bonito — `LookAndFeelPackage` really was stock.**
[...]

Issue #1007, opened 2026-08-05
LUKS E2E: VERSION_CODENAME branding failure on yellowfin, skipjack, albacore desktop cells


hanthor:
## Problem

The branding contract check fails on yellowfin (AlmaLinux Kitten), skipjack (CentOS Stream 10), and albacore (AlmaLinux 10) standard desktop cells (gnome, kde, cosmic, niri, xfce) with:

\`\`\`
FAIL: VERSION_CODENAME is '', want 'Thunnus albacares' (upstream codename survived)
TUNAOS_BRANDING_FAIL variant=yellowfin failures=1
\`\`\`

The \`-hwe\` and \`-nvidia\` overlay variants pass (gnome-hwe, gnome-nvidia, base-hwe, base-nvidia all green), indicating that the overlay's \`90-image-info.sh\` re-run sets VERSION_CODENAME correctly, but the base desktop stages do not.

## Root Cause (suspected)

\`90-image-info.sh\` writes VERSION_CODENAME into \`/usr/lib/os-release\`, but the desktop stage's \`install-desktop.sh\` re-runs the verify check which expects it. The overlays' re-run of \`90-image-info.sh\` paper-patches this, but the standard desktops (no overlay) see the unset codename.

## Affected Cells (from runs 31010714483, 31010718896, 31010736963)

- yellowfin: gnome, kde, xfce, niri, cosmic (5 cells)
- skipjack: gnome, kde, xfce, niri, cosmic (5 cells)
- albacore: gnome, kde, xfce, niri, cosmic (5 cells)

## Passes

- All three: base-hwe, base-nvidia, gnome-hwe, gnome-nvidia, gnome-nvidia-hwe

## Branch

\`fix/marlin-luks-green\`

Issue #979, opened 2026-08-02
Tracking: Complete LUKS E2E Verification across remaining published desktop cells


hanthor:
## Context
Under the TunaOS Production-Readiness Standard, an image edition must satisfy two independent gates:
1. **Desktop & Asset Contract Sweep** (static/runtime DE binaries, portal units, wallpapers, Plymouth theme, compiled dconf database).
2. **LUKS E2E Verification** (QEMU live ISO boot -> fisherman TPM2 LUKS installation -> reboot -> auto-unlock -> desktop boot contract).

## Verified Green (Production Ready)
- `sailfin:gnome` (openSUSE Tumbleweed GNOME) — Run 30738570002
- `yellowfin:niri` (AlmaLinux Kitten 10 Niri) — Run 30725495374
- `grouper:gnome` (Ubuntu 26.04 GNOME)
- `flounder:xfce` (Debian 13 Trixie XFCE)
- `marlin:niri` (Arch Linux Niri)

## Pending LUKS Verification
- `marlin:cosmic` & `marlin:cosmic-cachyos` (Runs 30748694693 & 30748696967 in progress)
- `albacore` (gnome/kde/niri/xfce)
- `yellowfin` (gnome/kde/xfce)
- `skipjack` (gnome/kde/niri/xfce)
- `sailfin` (kde/xfce)
- `grouper` (kde/xfce)
- `flounder` (gnome/kde)

## Action Plan
Drive remaining desktop image matrix cells through targeted `luks-e2e.yml` runs until all published variants achieve 100% LUKS E2E verification.

hanthor:
Opened tuna-os/tunaos#1225.

This issue's own "Verified Green"/"Pending" lists are stale — the live docs/MATRIX-STATUS.md snapshot (2026-08-08) shows the real state is worse in places than what's listed here: yellowfin is 0/5, albacore is 2/5 (gnome/kde/cosmic red, niri/xfce already green). Before treating any of that as new bugs to diagnose, I pulled the actual failing job logs.

[...]

PR #1020, opened 2026-08-05
fix: close six LUKS E2E defects — os-release, build-config drift, copr, KDE, zypper

### LINKED ISSUE #1011: LUKS E2E: sailfin — openSUSE Tumbleweed 404s on zypper install (rolling breakage)
## Problem

All sailfin (openSUSE Tumbleweed) cells fail during package installation with 404 errors from openSUSE mirrors:

\`\`\`
Preloading: libwebpmux3-1.6.0-2.3.x86_64.rpm [Error: "The requested URL returned error: 404", trying next mirror.]
Preloading: fontconfig-2.18.1-1.3.x86_64.rpm [Error: "The requested URL returned error: 404", trying next mirror.]
Preloading: libharfbuzz0-14.2.1-1.3.x86_64.rpm [Error: "The requested URL returned error: 404", trying next mirror.]
\`\`\`

This is Tumbleweed rolling — packages that were available when manifests were pinned have since been replaced by newer versions.

## Root Cause

Sailfin's desktop manifests pin package versions that are no longer in the Tumbleweed repos. Tumbleweed aggressively rotates packages (old versions are removed within days).

## Fix Options

1. **Remove version pins** from sailfin manifests (let zypper install latest)
2. **Pin to a snapshot repo** (e.g., \`download.opensuse.org/history/...\`)
3. **Use zypper's \`--no-recommends\`** with unversioned package names in the manifest

## Affected Cells

- sailfin: ALL (gnome, kde, xfce, niri, cosmic)

## Additional Note

sailfin:cosmic and sailfin:niri also fail with 'target not found in Dockerfile' — these desktop stages don't exist yet in Containerfile.opensuse (only gnome, kde, xfce are defined). Pre-existing gap tracked separately from the 404 issue.

## Branch

\`fix/marlin-luks-green\`

### LINKED ISSUE #1012: LUKS E2E: sailfin/guppy missing Containerfile targets for cosmic and niri
## Problem

The LUKS E2E matrix generates cells for all desktop flavors defined in \`build-config.yml\`, but some Containerfiles don't have matching build targets:

**sailfin (openSUSE):**
\`\`\`
Error: The target "cosmic" was not found in the provided Dockerfile
Error: The target "niri" was not found in the provided Dockerfile
\`\`\`

**guppy (Gentoo):**
\`\`\`
Error: The target "xfce" was not found in the provided Dockerfile
\`\`\`

## Root Cause

\`build-config.yml\` declares these flavors with \`build_image: true\`, but:
- \`Containerfile.opensuse\` only has a single \`desktop\` target (no per-flavor aliases for cosmic/niri)
- \`Containerfile.gentoo\` is missing the xfce target

## Fix

Add the missing \`FROM ... AS <flavor>\` target aliases to these Containerfiles, matching the pattern in Containerfile.arch/debian/el10 where each flavor gets its own named stage.

## Affected Cells

- sailfin: cosmic, niri
- guppy: xfce

## Related

#923 (guppy: niri and cosmic have no ebuilds in the Gentoo main tree)

## Branch

\`fix/marlin-luks-green\`

### LINKED ISSUE #1014: LUKS E2E: gurnard (Ubuntu noble) — 'Unknown variant' in justfile/build system
## Problem

Gurnard (Ubuntu noble) fails immediately with:

\`\`\`
Unknown variant: gurnard
error: Recipe \\`build\\` failed with exit code 1
\`\`\`

## Root Cause

The \`gurnard\` variant is defined in \`build-config.yml\` but the justfile's \`build\` recipe doesn't recognize it — likely it's not mapped to its Containerfile (\`Containerfile.ubuntu\`) in whatever lookup the build system uses.

## Fix

Add \`gurnard\` to the variant → Containerfile mapping in the justfile or the \`resolve-flavor.sh\` / \`build-image-inner.sh\` scripts.

## Affected Cells

- gurnard: ALL (pantheon — the only desktop it ships)

## Branch

\`fix/marlin-luks-green\`

### LINKED ISSUE #1016: LUKS E2E: yellowfin/skipjack — 'Unable to find a match: dnf5-command(copr) dnf5-command(config-manager)'
## Problem

Yellowfin and skipjack desktop builds emit a non-fatal warning during package installation:

\`\`\`
Error: Unable to find a match: dnf5-command(copr) dnf5-command(config-manager)
\`\`\`

This doesn't directly cause the build failure (the VERSION_CODENAME branding check does — see #1007), but it indicates that \`dnf5-plugins\` or equivalent providing the \`copr\` and \`config-manager\` commands is missing from the AlmaLinux Kitten / CentOS Stream 10 base images.

## Root Cause

The build scripts likely try to \`dnf install dnf5-command(copr) dnf5-command(config-manager)\` to enable COPR repos during the build. These are virtual provides from \`dnf5-plugins\` which is not available in the EL10 base container images.

## Impact

Non-fatal currently (the build continues), but if any desktop package depends on COPR-provided packages being available during build, they'll silently be missing.

## Affected Variants

- yellowfin (AlmaLinux Kitten)
- skipjack (CentOS Stream 10)
- Possibly albacore (AlmaLinux 10)

## Branch

\`fix/marlin-luks-green\`

hanthor:
Targets `fix/marlin-luks-green` (#1018) rather than `main`, so the diff here is only the delta on top of it.

Working the open `luks-e2e` issues. In four of the six, the issue's stated root cause turned out to be wrong or incomplete — each is written up below with the evidence, because the wrong cause is what made them survive.

| Issue | Stated cause | Actual cause |
|---|---|---|
| #1015 marlin | `40-services.sh` lacks pacman path | that, **plus** os-release written to the wrong file |
| #1007 branding | — | `sed` no-ops on bases that omit the key |
| #1014 gurnard | not mapped to its Containerfile | that, **plus** missing from the base-image table, **plus** no `pantheon` stage |
| #1012 targets | missing stage aliases | confirmed as stated |
| #1016 copr | cosmetic warning | **not cosmetic** — silently drops COPR packages |
| #1008 KDE | theme not set | overwritten by package install; flounder half was a **false negative** |
| #1011 sailfin | manifests pin versions | they pin nothing — stale metadata |

---

### 1. os-release written to a file nothing reads (#1015, #1007)

The marlin canary ([run 31013418173](https://github.com/tuna-os/tunaOS/actions/runs/31013418173)) reported **all ten** branding fields as upstream while `image-info.json` — written by the same script a few lines earlier — was correct. That combination made it look like `90-image-info.sh` never ran. It ran; it wrote the wrong file.

Two independent defects, both required:

**Sparse bases.** A bare `sed s|^KEY=.*|...|` only lands when the base already defines the key, and silently exits 0 otherwise. No RPM base defines `VERSION_CODENAME`; Arch defines neither it nor `VARIANT_ID`. `4f139df9` fixed this by adding `osr_set` calls *alongside* the sed block; this converts the sed block itself, so there's one mechanism rather than two that must agree.

**Split os-release.** Confirmed by pulling the pinned base image's layers rather than inferring:

```
=== layer 3fa976739799 (rootfs) ===
etc/os-release       SYMLINK -> /usr/lib/os-release
usr/lib/os-release   REGULAR FILE 382

=== layer 5120c48a8d37 (top, 9 KB) ===
etc/os-release       REGULAR FILE 411        # replaces the symlink
```

That top layer's contents match all ten failures line for line. Everything that reads system identity — systemd, GNOME About, GDM, fastfetch, `verify-branding.sh:61` — consults `/etc/os-release` first.

Uses `-ef` (device+inode, through symlinks) rather than an `IS_ARCH` flag: Arch is only split because a later layer clobbered the symlink, so the condition must be evaluated against the filesystem as built. It also avoids the opposite bug — `sed -i` replaces a symlink with a regular file, which would split `/etc` from `/usr/lib` on every RPM base.

### 2. build-config.yml vs. hand-maintained copies (#1014, #1012)

`scripts/get-base-image.sh` carried its own variant → base_image table, drifted three ways:

| variant | build-config.yml | get-base-image.sh |
|---|---|---|
| bonito | fedora-bootc:**44** | fedora-bootc:**43** |
| guppy | stage3@**sha256:9d03b48…** | stage3:latest (unpinned) |
| gurnard | ubuntu:noble | **absent** |

Only gurnard failed loudly. The other two are worse *because* they don't: a build through the script got a different base image than one reading the config, with no error either way. `reusable-build-image.yml` already falls back to reading build-config when the script exits non-zero — which is why gurnard built in CI and only died on the `just build` path (`Justfile:126`). The fallback hid the gap rather than closing it.

`Containerfile.gentoo` was the same bug in another spelling: it declared `ARG BASE_IMAGE`, ignored it, and hardcoded `stage3@sha256:08914e15…` against the config's `9d03b48…`.

gurnard had two more gaps behind the first — `resolve-flavor.sh` never mapped it (so an Ubuntu noble variant resolved to `Containerfile.el10`), and `Containerfile.ubuntu` had no `pantheon` stage.

Added the two missing aliases from #1012. Deliberately **not** adding gentoo niri/cosmic: no ebuilds in the main tree (#923), so an alias would only move the failure from "no target" to "no package".

### 3. copr plugin — one transaction, all-or-nothing (#1016)

`dnf -y install 'dnf5-command(copr)' 'dnf-command(copr)' ...` put every provider name in **one** transaction. dnf fails the whole transaction when any single name is unmatched, and the `dnf5-*` provides don't exist on EL10 — so yellowfin and skipjack installed **none** of them, including `dnf-command(copr)`, which they do have. The error names only the unmatched pair, which is why it read as cosmetic.

It wasn't. With no copr plugin, `dnf copr enable` fails, the repo is never enabled, and the install falls back to base repos — so a COPR-only package is silently absent while every step reports success. The issue flagged that as hypothetical; it was already happening.

### 4. KDE (#1008)

**LookAndFeel:** `system_files` ships `kdeglobals` with our theme, copied in during the base stage — and Fedora's `kde-settings` RPM owns that exact path, so installing Plasma overwrote it. Re-asserted after desktop install. Written to **both** `/etc/xdg/kdeglobals` and the kde-settings profile dir: on Fedora/EL that profile dir precedes `/etc/xdg` in `XDG_CONFIG_DIRS`, so writing only `/etc/xdg` would satisfy the checker (it greps `/etc/xdg` first) while Plasma still loaded Fedora's theme — a green check on an unbranded image, which is worse than the red one.

**Greeter — a false negative.** `flounder:kde` reported "greeter not installed". It had one. Debian trixie's `sddm` package ships `/usr/bin/sddm`, `/usr/lib/systemd/system/sddm.service`, `/etc/pam.d/sddm`, `/etc/sddm/Xsession`, `/usr/share/sddm/faces`, `/usr/share/sddm/flags` — and **no** `/usr/share/sddm/themes`, **no** `sddm.conf.d` (verified against the package file list). The check only looked for themes and conf.d, a Fedora/Arch packaging shape. The unit and binary are what prove installation; themes are incidental to a distro's packaging split.

### 5. sailfin 404s (#1011)

The issue read the 404 storm as pinned versions rotated out of Tumbleweed. The manifests pin **nothing** — gnome 62 packages, kde 47, xfce 37, cosmic 41, niri 40, zero version constraints.

The mechanism is stale metadata: the desktop stage installs using the index cached by the *base* stage, layer caching makes that arbitrarily old, and Tumbleweed deletes superseded builds within days. "Trying next mirror" forever is the signature of a resolution *no* mirror can satisfy, not one bad mirror. `Containerfile.opensuse` refreshes at lines 71 and 81 before its own installs; the desktop stage never did. A plain retry is useless — the same stale resolution 404s identically, so the index has to change between attempts, not the delay.

---

## Testing

New tests target the *class*, not the instances — the existing per-variant tests pinned literal strings, which is exactly what let the base-image drift live (they passed while the config said something else):

- every variant in `build-config.yml` resolves to that config's `base_image`
- every `build_image: true` flavor has a matching `FROM … AS <target>` in the Containerfile `resolve-flavor.sh` picks. **Verified it fails and names the cell** when an alias is removed
- os-release writing against three fixtures: sparse base, split `/etc`, conventional symlink
- KDE `LookAndFeelPackage` merge: Fedora overwrite, `[KDE]` without the key, no `[KDE]`, `[KDE]` last, no file, idempotency, and an unrelated section left alone

Tests extract the real functions from the scripts rather than restating them, so they track the implementation.

Full suite: 690 pass, 21 fail — the same 21 fail without these commits (local env has the Python `yq` jq-wrapper instead of Go `yq`, and no `shellcheck`); verified by stashing. `shellcheck --severity=error --exclude=SC1091,SC2114` clean repo-wide, matching `lint.yml`.

LUKS E2E dispatched on this branch for `marlin:gnome`, `marlin:kde`, `yellowfin:gnome`, `bonito:kde`, `flounder:kde`, `sailfin:cosmic`, `guppy:xfce`, `gurnard:pantheon`.

Closes #1011, closes #1012, closes #1014, closes #1016
Refs #1007, #1008, #1015, #979


hanthor:
Pulled the actual base image layers from the registry to confirm the diagnosis rather than leave it as inference. It's sharper than the PR body says — the symlink *is* there in the base layer, and the image's own final layer overwrites it.

Base image is pinned in `Containerfile.arch` to `docker.io/archlinux/archlinux@sha256:0b429ff…`, which has two layers:

```
=== layer 3fa976739799 (rootfs) ===
etc/os-release       SYMLINK -> /usr/lib/os-release
usr/lib/os-release   REGULAR FILE 382

=== layer 5120c48a8d37 (top, 9 KB) ===
etc/os-release       REGULAR FILE 411        # replaces the symlink
```

That 9 KB top layer contains four entries — `etc/ld.so.cache`, `var/cache/ldconfig/aux-cache`, and `etc/os-release`. Its contents:

```
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
VERSION_ID=20260805.0.568464
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo
```

Every one of the ten failures in [run 31013418173](https://github.com/tuna-os/tunaOS/actions/runs/31013418173) is this file, line for line — `PRETTY_NAME="Arch Linux"`, `SUPPORT_URL="https://bbs.archlinux.org/"`, that exact `BUG_REPORT_URL`, `LOGO=archlinux-logo`, and no `VERSION_CODENAME`, `VARIANT_ID`, `VARIANT`, `IMAGE_ID`, `IMAGE_VERSION` or `DEFAULT_HOSTNAME` anywhere in it. The build was branding `/usr/lib/os-release` correctly the whole time and the checker was reading this.

Two things follow for the fix as written:

- The `-ef` guard is load-bearing in **both** directions, not just as RPM-base politeness. Arch is only a split because a later layer clobbered the symlink — so the condition has to be evaluated against the filesystem as built, which is what `-ef` does, rather than keyed off a distro flag like `IS_ARCH`. If upstream ever drops that top layer, the guard silently collapses back to one write with no code change.
- `VERSION_ID` is present here but absent from `/usr/lib/os-release`, which is why `40-services.sh:7` (`. /usr/lib/os-release; echo ${VERSION_ID%.*}`) yields empty on Arch. Harmless today — nothing on the pacman path reads `MAJOR_VERSION_NUMBER` — so I'm leaving it alone rather than widening this PR, but noting it since it's the same split.

---

hanthor:
The `Connection timed out during banner exchange` cells are **not one bug**. Splitting them, with the serial logs as evidence.

## What the diagnostic found

`grouper:kde` (run 31116305245) emitted:

```
# ipv4:
# tcp listeners: 0.0.0.0:22 127.0.0.53%lo:53 127.0.0.54:53 [::]:22
```

sshd is listening. The guest has **no global IPv4 address**. QEMU user-mode networking accepts the connection host-side and then has nothing to deliver it to, which is exactly what "timed out during banner exchange" looks like from the host. `ok - a network manager is active` was never evidence of a lease — and here `NetworkManager-wait-online.service` **Finished successfully**, which it will do when there is no managed device to wait for.

## Why the guest has no address: it has no NIC

Counting `virtio_net` lines in each serial:

| cell | `virtio_net` | verdict |
|---|---|---|
| `flounder-sid:xfce` (**passes**) | 3 — `virtio_net virtio4 ens6: renamed from eth0` | NIC present |
| `guppy:xfce` (fails) | 1 — `virtio_net virtio4 ens6: renamed from eth0` | **NIC present** |
| `grouper:xfce` (fails) | 0 | no NIC |
| `grouper:kde` (fails) | 0 | no NIC |

grouper enumerates `virtio_blk virtio2`, `virtio_blk virtio3`, `virtio_scsi virtio1` — and stops. No `virtio4`, and no virtio-vga either, while guppy and flounder-sid have both.

**So there are at least two distinct causes behind one symptom.** grouper has no network device at all; guppy has `ens6` and fails for some other reason. Anything that "fixes guppy" and is then assumed to fix grouper (or vice versa) will be a wrong fix that looks right.

## What I could not determine

Whether grouper's image is missing the `virtio_net`/`virtio_gpu` modules, or the devices are present and unbound. Two facts pull against each other and I'd rather flag that than pick one:

- `scripts/iso-e2e.sh` attaches `-netdev user,... -device virtio-net-pci` at **every** QEMU invocation site (810, 1514, 2041, 2149, 2214), so the device is being offered.
- but `modprobe@efi_pstore.service` **succeeded** in the same boot, so module loading works in general — which argues against a wholesale modules/kernel mismatch.

grouper runs `7.0.0-29-generic` (resolute GA, selected by `ubuntu-kernel.sh` because it has `CONFIG_EROFS_FS_BACKED_BY_FILE=y`); the passing control runs `7.1.6+deb14-amd64`.

The next step needs the image, which I can't pull here:

```
podman run --rm ghcr.io/tuna-os/grouper:kde \
  sh -c 'ls /usr/lib/modules/*/kernel/drivers/net/virtio_net.ko* /usr/lib/modules/*/kernel/drivers/gpu/drm/virtio/ 2>&1; ls -d /usr/lib/modules/*'
```

If `virtio_net.ko` is absent, or `/usr/lib/modules/<ver>` doesn't match `7.0.0-29-generic`, that closes it.

## One testable prediction

`grouper:gnome` is currently green on the board. If grouper's **base** lacks the NIC, that green is stale and gnome must fail the same way now — the kernel and modules come from the base, not the desktop stage. Re-running `grouper:gnome` decides it: if it fails, this is a base-image regression affecting all four grouper flavors and the board is overstating grouper by one cell.

Diagnostic emission has been moved into `tunaos-live-ready.service` (7910bdea) as well, because the version in `e2e-runtime-checks.sh` only runs on `graphical.target` — so `grouper:xfce`, which fails exactly this way, emitted no TAP at all and the serial recorded nothing.

---

hanthor:
**Correction to my earlier comment.** The `virtio_net` table above is wrong, and the conclusion I drew from it does not hold.

I predicted that if grouper's base lacked a NIC, `grouper:gnome`'s green must be stale and it would fail identically. It was run to decide exactly that ([run 31126550532](https://github.com/tuna-os/tunaOS/actions/runs/31126550532)):

**`grouper:gnome` PASSED**, including LUKS install + unlock verification. And its serial shows:

```
# ipv4: ens4             UP             10.0.2.15/24 ;
TUNAOS_LIVE_NET ipv4=10.0.2.15/24,
TUNAOS_LIVE_NET listeners=0.0.0.0:22 127.0.0.53%lo:53 127.0.0.54:53 [::]:22
```

a working NIC with an address — **and zero `virtio_net` lines anywhere in the log.**

So the inference "no `virtio_net` line ⇒ no NIC" is false. grouper simply does not log the interface rename to the serial console; the absence of that line says nothing about whether a NIC exists. I reasoned from the absence of a log line, which is not evidence.

What that invalidates, and what survives:

- ❌ **"grouper:xfce and grouper:kde have no NIC"** — withdrawn. Unsupported.
- ❌ **"grouper's base is broken for all four flavors"** — falsified. `grouper:gnome` passes on the same base, so its board green is **not** stale and no downward correction to `docs/MATRIX-STATUS.md` is needed.
- ✅ **"grouper:kde has no global IPv4 address"** — still stands, and on direct evidence rather than inference: its own `# ipv4:` line was empty while `# tcp listeners:` showed `0.0.0.0:22`. That is the diagnostic reporting the address list itself, not me reading absence into a driver message.
- ✅ **guppy is a different bug from grouper** — still stands, though now for a weaker reason than I gave. It was based partly on the same bad `virtio_net` signal, so treat the two as *not yet shown to share a cause* rather than as *shown to differ*.

Net: grouper:kde/xfce reach the live environment with no address, and **why** is once again open. The mechanism I proposed was wrong.

One thing did work as intended: the diagnostic relocated in 7910bdea fires from `tunaos-live-ready.service` and reported `ipv4=` and `listeners=` twice (live and installed) on a variant where the old placement in `e2e-runtime-checks.sh` would have run — and, on `grouper:xfce`, would not have. That marker is now the thing to read, rather than driver lines.

---

PR #1021, opened 2026-08-06
refactor(bootc): deduplicate Containerfiles, starting with dracut config

hanthor:
Targets `claude/marlin-matrix-green-nnhntx` (#1020) so the diff here is only the refactor, and #1020 keeps verifying its bug fixes undisturbed.

## The duplication, measured

Surveying the six Containerfiles turned up four structural duplications. These aren't stylistic — **each one already cost matrix cells in the last day**:

| Duplication | Copies | What it cost |
|---|---|---|
| composefs ⇒ erofs in initramfs | 5 composefs variants, 1 carries the driver | marlin boots to emergency shell |
| Identity `ARG`/`ENV` block | 6, opensuse+gentoo had none | every sailfin and guppy cell |
| Context stage `COPY` set | 6, all different | gurnard `No manifest found` (#1014) |
| ostree/bootc layout | 3–4 hand-written | latent |

Plus literal redundancy: **gentoo writes `prepare-root.conf` twice** (lines 170 and 251) with its whole symlink set duplicated, and opensuse writes its symlink set twice.

The pattern is consistent: a fact true of a *class* of variants has to be retyped per file, and the copies drift silently.

## This PR

Extracts the dracut configuration into `build_scripts/bootc/dracut-config.sh` — the repo's existing idiom, the same move `install-desktop.sh` already made for desktops. Wires `Containerfile.arch`.

I picked this slice first because it carries a live bug. Each Containerfile had its own dracut config, in its own order, with its own omissions:

```
arch      bootc crypt dm, omit tpm2-tss pcsc, add_drivers erofs
debian    bootc
opensuse  bootc crypt dm, omit tpm2-tss pcsc, install_items
gentoo    (none — CLI flags only)
ubuntu    omit pcsc
```

Those aren't distro tuning, they're the order the variants happened to be debugged in. A composefs root *is* an EROFS image with an overlayfs on top, so "this image is composefs" and "its initramfs can mount erofs" are one fact written twice — and they'd drifted apart on four variants.

The script now **derives** the drivers from the image, reading the same `prepare-root.conf` that `scripts/lib/common.sh` probes to choose the install path. The initramfs and the installer can no longer disagree about what the image is.

**Drivers are probed, not asserted.** `add_drivers` on a name the kernel lacks is a dracut-install error, so listing erofs unconditionally would turn a fix for one variant into a build failure on a base that builds it in (`=y`) or omits it.

## What the tests caught

Both bugs here were mine, found before they reached CI.

The script takes a `TUNAOS_SYSROOT` prefix so it runs against fixture trees, following `TUNAOS_OS_RELEASE` in `verify-branding.sh`. That paid off immediately — the first run reported `composefs: 0` for a *composefs* fixture, because one path substitution hadn't applied and detection was still reading the real `/usr`.

The ordering guard caught the second. Wiring arch put `dracut-config.sh` **before** the block writing `prepare-root.conf`, so composefs read as off and the initramfs would have shipped with no erofs — silently, every step exiting 0. Fixture tests can't see that; it's a property of the Containerfile. The test asserts the declaration precedes the call in every file that calls it, and names the file and both line numbers. Verified it goes red when arch is swapped back.

## Not yet wired, and why

All three remaining variants **build their initramfs before declaring what kind of image they are**:

```
debian    prepare-root.conf:238   dracut:178
opensuse  prepare-root.conf:219   dracut:198
gentoo    prepare-root.conf:170   dracut:137
```

That ordering is the structural reason the erofs fact couldn't be derived anywhere. Fixing it means moving each declaration ahead of the initramfs build — a real change per file (debian's layout block also relocates the dpkg database; gentoo's runs twice), not a mechanical substitution. Putting "moved code" and "changed boot order" in one diff would make both unreviewable.

Each needs a LUKS run to verify, so they follow separately. The guard test means none of them can be wired incorrectly in the meantime.

Still to come from the survey: context-stage and identity `ARG`/`ENV` convergence (both enforced by test rather than extracted, since Dockerfile syntax can't be shared without codegen), the ostree-layout extraction, and gentoo's double-write.

## Testing

7 new bats cases, all negative-checked. Full suite unchanged at 21 pre-existing env failures (Python `yq` jq-wrapper, no `shellcheck`); `shellcheck --severity=error` clean repo-wide.

**Not yet verified by a build.** arch's dracut path needs a `marlin` LUKS run, which I'd rather do once #1020's erofs hypothesis is confirmed or refuted — otherwise a failure is ambiguous between the two.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Gu5fBnu8whXPqCFKd5aA7B)_

---
<a href="https://app.blacksmith.sh/tuna-os/codesmith/tunaOS/pr/1021"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>


hanthor:
The erofs hypothesis this PR is built on is now **confirmed**, which changes this PR's risk profile.

`marlin:gnome` passed end to end on `1965b8c` — [run 31060729168](https://github.com/tuna-os/tunaOS/actions/runs/31060729168), every step including `LUKS install + unlock verification`. The only change between that and the previous run, which died in the initramfs with

```
[FAILED] Failed to mount /sysroot.
[DEPEND] Dependency failed for bootc setup root.
Entering emergency mode.
```

was `add_drivers+=" erofs overlay xfs "`. A composefs root is an EROFS image with an overlayfs on top; the initramfs could not mount it.

So the four unwired variants — debian, opensuse, gentoo, ubuntu — are not "possibly missing a driver". They are composefs images whose initramfs demonstrably cannot mount their own root, and the only reason it has not surfaced on all of them is that most were failing earlier for other reasons. `flounder:kde` passing ([run 31061011519](https://github.com/tuna-os/tunaOS/actions/runs/31061011519)) means Debian now gets far enough to be exposed.

Two consequences for the follow-up work described in the PR body:

1. Moving each variant's `prepare-root.conf` write ahead of its initramfs build is no longer a tidiness argument — it is what lets the driver be derived at all. Worth doing next rather than last.

2. The ordering guard is doing more than style enforcement. A variant wired in the wrong order gets `composefs=0`, no erofs, and an image that builds clean and boots to an emergency shell — the precise failure that cost this matrix a day. That is why it names the file and both line numbers rather than just failing.

Still worth a `marlin` run against this branch before merge, to confirm the extracted script produces the same config the inline `printf` did. The passing run above used the inline version.

---

hanthor:
Correcting myself on one point above, because it changes what the follow-up wiring is *for*.

I wrote that the four unwired variants "are composefs images whose initramfs demonstrably cannot mount their own root". That overstates it, and `flounder:kde` passing in the same round disproves it for Debian: that run installed to a composefs root and booted it. Debian's initramfs mounts erofs fine today — the driver is presumably built in (`=y`) or already pulled in by its dracut module set, neither of which needs `add_drivers`.

What is actually established:

- **Arch needed it.** Confirmed by run 31060729168 against the identical prior failure.
- **Debian does not need it.** Confirmed by flounder:kde passing without it.
- **openSUSE, Gentoo, Ubuntu: unknown.** No cell has yet booted an installed system on those, so nothing has tested the claim either way.

Which is the case for probing rather than asserting, and the reason the script skips drivers that are built in or absent instead of listing them unconditionally — `add_drivers` on a name the kernel lacks is a dracut-install error, so a blanket list would have broken Debian to fix Arch.

So the remaining wiring is **deduplication, not a pending bug fix**. Its value is that the next variant to hit this gets the right answer derived from its own image rather than depending on whether someone remembered to type it — and that ordering the declaration before the initramfs build is what makes that derivation possible at all. Still worth doing next; just not for the reason I gave.

---

Issue #941, opened 2026-07-31
LUKS E2E: guest podman cannot see the offline store, so every cell falls into the SSH image-transfer fallback


hanthor:
Split out of #939. #940 bounds the fallback so it fails in 30 minutes instead of hanging for four hours. **This issue is the reason the fallback runs at all**, and it is what actually blocks the LUKS axis.

## The fallback is not the normal path

Passing cells never copy anything. They resolve the image in the guest and take Fisherman `bootcDirect`:

| run | cell | line |
|---|---|---|
| [30595701955](https://github.com/tuna-os/tunaOS/actions/runs/30595701955) | `yellowfin:xfce` | `Found canonical offline image ghcr.io/tuna-os/yellowfin:xfce — using Fisherman bootcDirect (no OCI copy)` |
| [30624963765](https://github.com/tuna-os/tunaOS/actions/runs/30624963765) | `bonito:gnome` | `Found canonical offline image ghcr.io/tuna-os/bonito:gnome — using Fisherman bootcDirect (no OCI copy)` |

All 13-14 green LUKS cells took that route. The `scp` at `scripts/iso-e2e.sh` is what happens when the guest cannot.

## What the failing runs show

From [30627886736](https://github.com/tuna-os/tunaOS/actions/runs/30627886736), `yellowfin:niri` on `main`:

```
==> Probing for ghcr.io/tuna-os/yellowfin:niri...
==> Found ghcr.io/tuna-os/yellowfin:niri in offline store images.json (podman query missed it)
==> Probing for localhost/yellowfin:niri...
==> No local image in offline store — transferring from host via SSH
```

The image **is** in `/var/lib/superiso-store/overlay-images/images.json` under its canonical ref. `sudo podman image exists` in the guest does not see it. The probe loop deliberately declines to trust `images.json` alone:

```
# The image exists but podman/bootc can not resolve it from the
# additional store (overlay driver mismatch). Do not set
# found_local here — let the code fall through to the network pull
```

so `found_local` stays 0 and control reaches the transfer.

## Not the tacklebox pin

`bonito:gnome` resolved fine on the same `fd95174` pin and the same `[composefs] enabled = no`, roughly an hour before `yellowfin:niri` failed to. So the differentiator is not the pin and not the composefs backend.

## Why it matters more than it looks

[...]

PR #1030, opened 2026-08-06
feat(iso-e2e): generic bootc install path for images without fisherman (tacklebox#178)

hanthor:
## What

The next rung of [tacklebox#178](https://github.com/tuna-os/tacklebox/issues/178)'s ladder, on top of the vsock SSH fallback (#1028): the harness's hard `fisherman not found → exit 3` gate blocks any full-LUKS e2e for the reference cells (aurora, bluefin) — stock bootc images that ship no fisherman.

When the live image has no fisherman **and** the caller names the image via `TBOX_E2E_IMAGE` (bare `org/name:tag` gets a `ghcr.io/` prefix), `run_install` now branches to a new `run_install_generic`:

- **Storage**: a new 32G sparse scratch disk is attached to every live boot (costs nothing unless written) and formatted/mounted at `/var/lib/containers` only by the generic path — the live overlay upperdir is an 8G tmpfs, so a stock-image pull (aurora ≈ 9G uncompressed) physically cannot land there (#941's lesson from the other direction). Swap is enabled as in the fisherman path.
- **Install**: guest MTU clamped (same SLIRP PMTU blackhole as bug #20), image pulled in the guest, then the canonical containerized `bootc install to-disk --wipe` — with `--block-setup tpm2-luks` under `--luks`, and the serial/plymouth kargs baked via `--karg` (bootc owns its BLS entries; `--karg` is the supported way in).
- **LUKS evidence**: reuses `e2e-luks-checks.sh`. Unlike fisherman's deliberate post-install enrollment model, bootc enrolls a systemd-tpm2 token at install time — so a missing token is a hard failure on this path (while remaining a NOTE on the fisherman path).
- **Reboot gate**: boots the installed disk **with the same swtpm state** — `start_swtpm` gains a `keep` mode, since swtpm exits with its QEMU and the sealed key only unseals against the same state — and treats a serial login prompt (or reached target) as the unlock proof. There is no known passphrase to inject (bootc generates and seals its own); a locked root hangs in the initramfs at the cryptsetup prompt and the gate times out with serial evidence + screenshot.

Without `TBOX_E2E_IMAGE`, the old hard error stands: the harness refuses to guess a registry ref from an ISO filename (the #1006 failure mode).

## Refactors

The guest heartbeat, BLS-karg/ESP-evidence dump, and poweroff-wait blocks move unchanged into shared helpers (`start_guest_heartbeat`, `append_installed_serial_kargs`, `poweroff_and_wait_vm`) used by both install paths.

## Testing

- 5 new bats tests (imgref qualification evaluated out of the script, fallback gating, scratch-disk staging, bootc flags, swtpm keep-state) plus updated boot-count guards (serial chardev 3→4, bootindex 2→3, with comments naming each boot).
- Full `test_iso_e2e.bats`: 69 ok locally; the only failures are the 4 `e2e-smoke/runtime-checks` cases that fail identically on unmodified main in this container (missing local stub deps).
- `bash -n`, shellcheck (no findings at warning+), `shfmt -d` clean.

Real-world validation: iso-builder aurora attempt 7 is running now against tunaOS main (vsock validation); once this merges plus the iso-builder side passes `TBOX_E2E_IMAGE` into the LUKS step, the next dispatch exercises this path end-to-end.

https://claude.ai/code/session_01S9hTUH9j7C8fJbCpuXefQN

---
_Generated by [Claude Code](https://claude.ai/code/session_01S9hTUH9j7C8fJbCpuXefQN)_

---
<a href="https://app.blacksmith.sh/tuna-os/codesmith/tunaOS/pr/1030"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>

### DIFF
--- a/.github/build-config.yml
+++ b/.github/build-config.yml
@@ -531,6 +531,12 @@ variants:
         build_image: true
         build_iso: true
         build_qcow2: false
+      - id: xfce-nvidia
+        stage: 3
+        build_image: true
+        build_iso: true
+        build_qcow2: false
+
   # ── Hummingbird ────────────────────────────────────────────────────────────
   # Fedora Hummingbird: container-native, hardened bootc distribution
   # (Konflux hermetic build model, CKI ARK kernel).
--- a/scripts/iso-e2e.sh
+++ b/scripts/iso-e2e.sh
@@ -73,6 +73,14 @@
 #                         vsock, authenticated by a per-run keypair injected
 #                         through the SMBIOS credential ssh.authorized_keys.root
 #                         (tacklebox#178). See the Guest SSH transport section.
+#   TBOX_E2E_IMAGE        Image ref for the generic (no-fisherman) install
+#                         path, e.g. "ublue-os/aurora:stable" (bare org/name
+#                         gets a ghcr.io/ prefix) or a fully qualified ref.
+#                         When the live image ships no fisherman and this is
+#                         set, run_install pulls it in the guest and installs
+#                         with `bootc install to-disk` (tpm2-luks under
+#                         --luks; the reboot gate then proves TPM auto-unlock
+#                         instead of passphrase injection).
 #
 # Exit codes:
 #   0  success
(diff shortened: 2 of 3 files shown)


hanthor:
Merging with the `regenerate` check red, deliberately. That check regenerates `docs/MATRIX-STATUS.md` against **live** Actions data using the PR's `build-config.yml`; the committed doc predates today's LUKS runs, so the drift is result-movement plus ca0c929's bonito flavor restore — not anything this PR's harness changes touch. It can't be fixed from the PR side without racing live data: a regeneration on current *main* would produce a 53-cell shape (main still lacks the restored flavor), which would still drift against this branch's 54-cell config. Post-merge, main has the restored flavor, and I'm dispatching the Matrix Status workflow so the scoreboard refreshes through its normal automation PR.

All substantive checks are green — Unit Tests fully (thanks to the bonito root-cause fix in ca0c929), all lint legs, both E2E cells, CodeQL.

---

PR #1039, opened 2026-08-07
ci: make the installer walkthrough's parity report actually reach anyone

hanthor:
`docs/INSTALLER-FRONTENDS.md` has been sitting on `_pending_` rows. The job meant to fill them — `installer-screenshots.yml` — has been producing nothing, silently, on every green run.

Three independent defects, any one of which was sufficient:

**1. `tesseract-ocr` was never installed.**
Deps are `qemu-system-x86 qemu-utils ovmf socat netpbm python3-pil`. `installer-walkthrough.py` does `shutil.which("tesseract")` and, when it's absent, prints `tesseract not installed — screen detection skipped`, writes `"ocr": false`, and **exits 0**. Every run published a report with all six screen rows unfilled, and a green tick. `installer-smoke.yml` has always installed it; this job never did.

**2. `FLAVOR` was never passed.**
The step calls `sudo bash scripts/run-walkthrough.sh "$ISO" walkthrough-out`. `run-walkthrough.sh` defaults to `FLAVOR="${FLAVOR:-de}"`, and `sudo` drops the environment regardless — so both matrix legs wrote `walkthrough-de.json`, a parity report naming neither frontend it came from. Now `sudo -E FLAVOR=…`.

**3. The report was never uploaded.**
It is written into `walkthrough-out/` while the artifact step uploads only `staged/*.png`. The file tunaOS's own installer matrix consumes never left the runner.

### The assertion that stops this recurring

A missing dependency must not look like a passing run — that is precisely the failure above. The walkthrough step now fails if the report is absent or reports `ocr: false`, and prints the screen rows it did fill.

### Also: gnome joins the smoke matrix

`installer-smoke.yml` ran `kde,cosmic,niri,xfce`. gnome is the flow the four forks are being asked to match, and was the only frontend with no per-DE smoke leg at all. Nothing else needed changing — the `APP` mapping already falls through to `org.bootcinstaller.Installer` for it, and the walkthrough's `--strict` list already names it. Only the default flavor list kept it from ever running.

### What this does not do

This is the reporting layer, not coverage. `installer-smoke.yml` still stops at *"the installer starts and reaches screens"* and never completes an install; `installer-screenshots.yml` still installs and boot-verifies only gnome and cosmic, and that verification is still `continue-on-error`. Closing that — phase 2, per-DE GUI-driven install through to a bootable system — is a separate change, and it has to answer the `--strict` constraint first: niri and xfwl4 are Smithay compositors that cannot render without virgl, so a GPU-less runner shows legitimately blank frames.

### Verification

`yamllint` clean on both files (only pre-existing long-line warnings, none on added lines). The behavioural claims are read from the scripts in this repo: `installer-walkthrough.py:358` for the `which("tesseract")` degrade, `:373` for the skip message, `:438` for the output filename, and `run-walkthrough.sh:196` for the `FLAVOR` default. Neither workflow runs on PRs, so CI here cannot exercise them — the real test is the next scheduled run or a `workflow_dispatch`.

https://claude.ai/code/session_01KNgsizQRcVzS3KiJ5BduiC

---

---
<a href="https://app.blacksmith.sh/tuna-os/codesmith/tunaOS/pr/1039"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>

### DIFF
--- a/.github/workflows/installer-screenshots.yml
+++ b/.github/workflows/installer-screenshots.yml
@@ -1,7 +1,10 @@
 name: Installer Walkthrough Screenshots
 
 # Boots a freshly built live ISO in QEMU, drives the bootc-install GUI with
-# scripts/run-walkthrough.sh (sendkey choreography + screendumps), and commits
+# scripts/run-walkthrough.sh (which delegates the driving to
+# scripts/installer-walkthrough.py — the fixed sendkey choreography this
+# comment used to describe was removed after it labelled three welcome-screen
+# frames as disk_select and confirm), and commits
 # the resulting flow screenshots to docs/images/installer/ — replacing the
 # repo's installer imagery with real captures on a schedule.
 #
@@ -70,7 +73,8 @@ jobs:
           sudo apt-get update -qq
           sudo apt-get install -y --no-install-recommends \
             just podman jq golang-go git \
-            qemu-system-x86 qemu-utils ovmf socat netpbm python3-pil
+            qemu-system-x86 qemu-utils ovmf socat netpbm python3-pil \
+            tesseract-ocr
           sudo wget -qO /usr/bin/yq https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64 && sudo chmod +x /usr/bin/yq
 
       - name: Log in to GHCR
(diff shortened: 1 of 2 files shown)


hanthor:
Correction to this PR's description, from dispatching the workflow ([run 31157709875](https://github.com/tuna-os/tunaOS/actions/runs/31157709875)).

**The framing was wrong.** I wrote that this job had been "producing nothing, silently, on every green run" and "published a green tick". It has not been green. Every recorded run has failed:

| | | |
|---|---|---|
| 2026-08-07 | dispatch | failure |
| 2026-08-03 | schedule | failure |
| 2026-07-27 | schedule | failure |
| 2026-07-23 | dispatch | failure |
| 2026-07-20 | schedule | failure |
| 2026-07-13 | schedule | failure |
| 2026-07-06 | schedule | failure |

Seven for seven, back to at least July 6.

**The three defects are still real** — `tesseract-ocr` is confirmed installed now (`Setting up tesseract-ocr (5.3.4-1build5)` in the dispatched run), `FLAVOR` is passed, the report is uploaded. But they were latent behind an earlier failure, so fixing them unblocked nothing on its own. That is the part I got wrong, and it changes the priority.

**What actually stops this job** is the ISO build, about six minutes in, before the walkthrough is ever reached:

```
Error: live customize for skipjack-cosmic: ... podman run --name tbox-customize-… 
stderr: Error: Failed to download metadata for repo 'baseos': Cannot prepare internal
mirrorlist: Curl error (6): Could not resolve hostname for
https://mirrors.centos.org/metalink?repo=centos-baseos-10-stream&arch=x86_64
[Could not resolve host: mirrors.centos.org]
```

DNS does not resolve inside tacklebox's `live-customize` container. The gnome leg fails the same way at the same stage; the 2026-08-03 run's two legs failed on the same ~4-minute timeline, consistent with the same cause, though I have only read the log for this run in full.

I am not guessing at a fix for this one. Podman DNS in CI is not a one-line change, and I have already been wrong twice today by reasoning past evidence I could have read — so this gets filed rather than patched. It is also the real precondition for phase 2, ahead of anything in this PR: there is no point extending per-DE install coverage on top of a job that cannot build an ISO.

---

PR #1056, opened 2026-08-07
live-iso: actually launch the installer on gnome and niri

hanthor:
The live ISO has one job: run the installer. On two of five flavors it never started one.

| flavor | launches installer? |
|---|---|
| kde | ✅ `/etc/xdg/autostart/org.tunaos.installer-live.desktop` |
| cosmic | ✅ same |
| xfce | ✅ same |
| **gnome** | ❌ **nothing** — pinned to the dash and left there |
| **niri** | ❌ **a TODO** |

niri's, in the repo's own words:

```
# Installer frontend: niri has no XDG autostart; org.tunaos.InstallerNiri is
# baked into the live squash by customize-live.sh and launched from
# the shell / DMS launcher. Add spawn-at-startup to the niri config when the
# live config.kdl is introduced.
```

### This is what run 31171184497 actually recorded

And I misread it twice before reading the adapters.

The frames show the **empty GNOME Activities overview** — which is what GNOME shows at login when *no windows exist* — with the installer merely pinned to the dash. 11 frames, 1 distinct visual state, 0/10 transitions, all six screen rows false.

Nothing was broken. Nothing had been launched.

My earlier reading of those frames as *"the installer window is mapped but rendering black"* was **wrong**: the two black rounded rectangles are empty workspace previews, not window thumbnails, and the dash icon is a favourite, which is present whether or not the app is running. I inferred a running app from a pinned icon.

### The fix

- **gnome** gets the same XDG autostart entry as its siblings.
- **niri** gets a systemd user unit bound to `graphical-session.target` instead. This script writes `/etc/xdg/niri-session.override`, not a `config.kdl`; dropping a partial `/etc/niri/config.kdl` would *replace* niri's defaults and take the keybinds and output setup with it. `niri-session` runs a systemd user session, so a unit reaches the same result without touching niri's configuration at all.

### The gate

`tests/bats/test_live_installer_launch.bats` asserts the property directly, with no VM:

1. every adapter contains **some** mechanism that starts its installer;
2. each references the app id `customize-live.sh` actually installs.

It deliberately does not prescribe *which* mechanism — doing so would have blocked niri. Both assertions pass on this branch and fail on `main` for gnome and niri. The `Unit Tests` job globs `tests/bats/*.bats`, so this runs on every PR.

### Why it survived

From outside, a live ISO with no installer looks exactly like a working one: the ISO builds, the VM boots, the desktop appears. And `installer-smoke.yml` — the workflow whose entire purpose is asserting the installer process is running — did not include gnome in its matrix until #1039.

### Verification

`bats` is not installed here, so both assertions were reproduced in plain bash against the five adapters: no-launcher set is empty, and all five app ids resolve. `bash -n` reports a syntax error on this file *and* on pre-existing `.bats` files, so it is not a valid check for the format.

The real proof is a dispatch of `installer-screenshots.yml` after merge — with #1053 merged, a run that still detects nothing now goes **red** instead of green.

https://claude.ai/code/session_01KNgsizQRcVzS3KiJ5BduiC

---

---
<a href="https://app.blacksmith.sh/tuna-os/codesmith/tunaOS/pr/1056"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>

### DIFF
--- a/live-iso/common/src/desktop-gnome.sh
+++ b/live-iso/common/src/desktop-gnome.sh
@@ -29,6 +29,31 @@ welcome-dialog-last-shown-version='4294967295'
 favorite-apps = ['org.bootcinstaller.Installer.desktop', 'bootc-installer.desktop', 'firefox.desktop', 'org.gnome.Nautilus.desktop']
 EOF
 
+# Launch the installer at session start.
+#
+# GNOME had NO autostart entry while kde, cosmic and xfce all did, so on the
+# gnome live ISO the installer was never started — the session came up in the
+# empty Activities overview with the app merely pinned to the dash. Run
+# 31171184497 recorded the result: 11 frames, 1 distinct visual state, 0/10
+# transitions, and all six screen rows false. Nothing was broken; nothing had
+# been launched.
+#
+# It went unnoticed because gnome was absent from installer-smoke.yml's flavor
+# matrix (added in #1039), and that is the workflow whose whole job is to
+# assert the installer process is running.
+#
+# Deliberately no OnlyShowIn=, for the reason spelled out in desktop-cosmic.sh:
+# systemd-xdg-autostart-generator gates on systemd-xdg-autostart-condition
+# against XDG_CURRENT_DESKTOP, and a mismatch silently skips the unit.
+mkdir -p /etc/xdg/autostart
+tee /etc/xdg/autostart/org.tunaos.installer-live.desktop <<'DESKEOF'
+[Desktop Entry]
+Type=Application
+Name=Install TunaOS
+Exec=flatpak run org.bootcinstaller.Installer
+Icon=org.bootcinstaller.Installer
+DESKEOF
+
 # Disable suspend/sleep so the installer doesn't go to sleep mid-install
 tee /usr/share/glib-2.0/schemas/zz3-tunaos-installer-power.gschema.override <<'EOF'
 [org.gnome.settings-daemon.plugins.power]
--- a/live-iso/common/src/desktop-niri.sh
+++ b/live-iso/common/src/desktop-niri.sh
@@ -51,52 +51,52 @@ ln -sf /usr/lib/systemd/system/greetd.service /etc/systemd/system/display-manage
 systemctl set-default graphical.target 2>/dev/null ||
 	ln -sf /usr/lib/systemd/system/graphical.target /etc/systemd/system/default.target 2>/dev/null || true
 
-# Disable screen lock for the live session
-mkdir -p /etc/xdg
-tee /etc/xdg/niri-session.override <<'NIRIEOF'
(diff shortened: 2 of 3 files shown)


hanthor:
Two corrections to the niri half, after going looking for the `config.kdl` the old TODO said did not exist.

### 1. It does exist — so niri now uses its own idiom

TunaOS ships **`system_files/usr/share/niri/config.kdl`** as the compositor default, and the entire session is already driven from it:

```kdl
spawn-at-startup "swaybg" "-i" "/usr/share/backgrounds/tunaos/tunaos-default.png" "-m" "fill"
spawn-at-startup "waybar"
spawn-at-startup "nm-applet"
spawn-at-startup "swaync"
spawn-at-startup "wl-paste" "--watch" "cliphist" "store"
```

So the installer gets one more line appended to that file, replacing the systemd user unit I pushed first. `spawn-at-startup` is niri's own mechanism and is demonstrably working in this exact image; the user unit rested on an assumption about how far `niri-session` takes `graphical-session.target`, which I had not verified.

This is still not the thing the original TODO warned about. Creating `/etc/niri/config.kdl` would *replace* the shipped default and take the keybinds, output setup and window rules with it. **Appending to the default does not.** The branch is fatal if the file is absent, since a live niri ISO with no installer launcher is precisely the defect this closes and is invisible from outside.

### 2. `/etc/xdg/niri-session.override` was inert

The adapter wrote:

```ini
[idle]
inhibit-when-fullscreen = false

[bind]
mod = Super
```

niri's configuration is **KDL**, and it reads `~/.config/niri/config.kdl`, `/etc/niri/config.kdl` or (on TunaOS) `/usr/share/niri/config.kdl` — never that path. Nothing else in the repo references it. It configured nothing; it only made the live session *look* configured.

Removed rather than ported, because there is nothing to port to: the shipped `config.kdl` spawns no swayidle or swaylock, so there was no screen lock for it to be disabling. Sleep is masked for real a few lines below, via `systemctl mask sleep.target ...`.

### Current state of the branch

| flavor | mechanism | app id |
|---|---|---|
| gnome | xdg-autostart | ✅ `org.bootcinstaller.Installer` |
| kde | xdg-autostart | ✅ `org.tunaos.InstallerKde` |
| cosmic | xdg-autostart | ✅ `org.tunaos.InstallerCosmic` |
| niri | spawn-at-startup | ✅ `org.tunaos.InstallerNiri` |
| xfce | xdg-autostart | ✅ `org.tunaos.InstallerXfce` |

---

PR #1084, opened 2026-08-07
smoke: give the guest 8G, and fail loudly when the kernel OOM-kills it

hanthor:
## The finding

Run 31216208138's cosmic leg captured nine **282-byte** PNGs — pure black — and reported:

```
not ok - cosmic: screen is not blank
  # 0/9 frames above stddev 0.02 (blank everywhere usually means no GL — niri/xfwl4 need virgl)
```

That hint is wrong, and it is the reason a GPU hypothesis survived most of today. The serial log says what actually happened:

```
oom-kill:constraint=CONSTRAINT_NONE ... task=cosmic-comp,pid=2146,uid=974
Out of memory: Killed process 2146 (cosmic-comp) total-vm:2697460kB, anon-rss:128268kB
Out of memory: Killed process 1746 (cosmic-comp) total-vm:2689756kB, anon-rss:158864kB
Out of memory: Killed process 3426 (cosmic-comp) total-vm:2639056kB, anon-rss:117840kB
```

**cosmic-comp was OOM-killed three times.** The compositor kept dying and restarting, so the framebuffer stayed black. Nothing to do with virgl — the same run's kernel log reports `virtio-vga detected` and a render node exists.

## Why it happened

`iso-e2e.sh` defaults to `--memory 4096` and `installer-smoke.yml` never overrode it. The live overlay's upperdir alone is an 8G tmpfs backed by guest RAM; a Wayland compositor on top of that does not fit in 4G.

## The changes

1. **`--memory 8192`** on the boot step. `ubuntu-latest` runners have 16G, so this leaves ample host.
2. **An OOM gate right after boot** — grep the serial log, fail naming the killed process. An OOM kill is indistinguishable from a GL failure in the frames, costs a full 25-minute leg to misdiagnose, and is one grep from certainty. It must never reach the walkthrough as "blank screen".
3. **Fix the misleading hint.** The blank-screen diagnostic and the DIAGNOSIS list now name the OOM check first and demote the GL path, rather than pointing at virgl.

The number is the least durable part of this; (2) and (3) are what stop the next person losing an afternoon to the same misread.

## Verification

- `yaml.safe_load` clean on the workflow; `py_compile` clean on the harness.
- The OOM lines quoted above are copied from the real `serial.log` in artifact 9009434604, not reconstructed.

---

---
<a href="https://app.blacksmith.sh/tuna-os/codesmith/tunaOS/pr/1084"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>

### DIFF
--- a/.github/workflows/installer-smoke.yml
+++ b/.github/workflows/installer-smoke.yml
@@ -210,11 +210,21 @@ jobs:
         env:
           FLAVOR: ${{ matrix.flavor }}
         run: |
-          # Smithay compositors (niri, xfwl4) cannot render without virgl, so a
-          # GPU-less runner shows blank frames legitimately — enforce only where
-          # rendering is possible; elsewhere still record screens for the matrix.
-          STRICT=""
-          case "${FLAVOR}" in kde|cosmic|gnome) STRICT="--strict" ;; esac
+          # Every flavor is strict. This used to exempt niri and xfce on the
+          # premise that "Smithay compositors cannot render without virgl, so a
+          # GPU-less runner shows blank frames legitimately" — which is false on
+          # both halves. The kernel in these very runs reports
+          #   [drm] pci: virtio-vga detected
+          # so a render node exists; and the blank frames that motivated the
+          # carve-out were the COMPOSITOR BEING OOM-KILLED (run 31216208138:
+          # cosmic-comp killed three times), fixed by the 8G guest above.
+          #
+          # An exemption granted on a wrong premise is worse than no gate: it
+          # made two of the five frontends permanently unable to fail the render
+          # and screen-coverage checks, which is exactly the drift this workflow
+          # exists to catch. If niri or xfce genuinely cannot render here, that
+          # must show up as a red leg with frames to look at, not as silence.
+          STRICT="--strict"
           if sudo test -S smoke-out/monitor.sock; then
             set +e
             sudo python3 scripts/installer-walkthrough.py \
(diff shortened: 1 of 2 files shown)


hanthor:
Pushed a second commit to this branch — same root cause, and arguably the more important half.

`STRICT` was set only for `kde|cosmic|gnome`:

```bash
STRICT=""
case "${FLAVOR}" in kde|cosmic|gnome) STRICT="--strict" ;; esac
```

with the justification *"Smithay compositors (niri, xfwl4) cannot render without virgl, so a GPU-less runner shows blank frames legitimately"*.

Both halves of that premise are false:

- The kernel in these same runs reports `[drm] pci: virtio-vga detected` — a render node exists. `-virgl` means virgl is disabled, not that there is no DRM device.
- The blank frames that motivated the carve-out were the compositor being **OOM-killed**, which the 8G guest in the first commit fixes.

The effect was that **two of the five frontends were structurally incapable of failing** the render and screen-coverage checks — the exact drift this workflow exists to detect. Now all five are strict.

Expect this to make niri and xfce legs go red before they go green. That is the intended outcome: a red leg ships frames you can look at, an exemption ships nothing.

---

Issue #1278, opened 2026-08-10
ci: make image-factory lifecycle coverage a generated required gate


hanthor:
## Gap

`.github/build-config.yml` currently declares 13 variants, 142 flavor definitions, and multiple platforms, while `docs/MATRIX-STATUS.md` reports several independent CI signals. There is no machine-enforced rule that every enabled matrix cell has the complete evidence required by its declared capabilities before it is considered shippable.

This allows "never tested", stale, or partially covered cells to coexist with green aggregate workflows.

## Proposed contract

Generate a lifecycle ledger directly from `.github/build-config.yml`. For every enabled `variant × flavor × platform` cell, record:

- container build and immutable published digest;
- desktop/package contract;
- declared output availability (OCI, ISO, QCOW2, hardware installer);
- boot evidence for each declared output;
- installer evidence where installable;
- update/rebase/rollback evidence;
- signature, SBOM, and provenance evidence;
- freshness timestamp and source workflow/run.

Each capability must be explicit: required, not applicable with a reason, experimental, or temporarily waived with an owner and expiry.

## Acceptance criteria

- [ ] One generated machine-readable ledger is the source for the human matrix-status page.
- [ ] Adding a variant/flavor/platform without capability declarations fails CI.
- [ ] Stable cells fail the release/promotion gate when required evidence is missing, red, or stale.
- [ ] Beta/experimental waivers are visible and expire automatically.
- [ ] Aggregate green cannot hide a red or never-tested required cell.
- [ ] The tracker links directly to the latest evidence run for every requirement.

Related: #1175 (lifecycle policy), #1196/#1270 (admission governance), #979 (LUKS coverage), #1187/#1193 (supply chain), #1254 (release parity).

PR #1435, opened 2026-08-13
supply-chain: attest provenance alongside SBOM for signed images

hanthor:
## Summary

#1187 is a large checklist; most of it is already implemented (verified by reading the live workflows before starting):

- Keyless Cosign OIDC signing of every platform manifest + multi-arch index by digest, verified immediately (`reusable-build-image.yml`'s `sign` job)
- SPDX SBOM generation + signed attestation per platform image
- ISO checksum manifests + Cosign v3 `sign-blob`/bundle, uploaded alongside the ISO to R2 (`publish-iso-groups.yml`)
- `docs/VERIFY-ARTIFACTS.md` with copy/paste verify commands and a documented trust boundary (narrow identity + issuer, no permissive regex)
- Signing/attestation jobs gated on `github.event_name != 'pull_request'`, so fork/PR events structurally cannot reach them

One concrete, named checklist item was still missing: **"Attach provenance for the source commit, workflow/run, variant, flavor, platform, base-image digest, and build configuration revision"** — only the SBOM attestation existed, no separate provenance attestation.

This PR adds it, in the same `sign` job, right next to the existing SBOM attestation, using the identical identity/issuer/verify pattern that's already proven to work:

- A `https://tunaos.org/provenance/v1` in-toto predicate per platform image, with `sourceCommit`, `sourceRepository`, `workflow`, `runId`, `runUrl`, `variant`, `flavor`, `platform`, `buildConfigRevision`.
- `cosign attest` + immediate `cosign verify-attestation`, same as the SBOM path.
- `docs/VERIFY-ARTIFACTS.md` gets a matching "Verify the provenance attestation" section.

**Deliberately not included**: the base-image digest. `get-base-image.sh` resolves a *tag*, not a pinned digest, inside the per-platform `build_push` matrix job (a live, 180-minute-timeout build path I can't test). Pinning it would mean adding a `skopeo inspect` call to every build cell across every variant/platform — a change to the hot build path I'm not willing to make blind. Left as a documented, explicit gap in both the code comment and the docs, for a follow-up that can actually be tested against a real build.

Remaining #1187 scope not touched here: the negative test (unapproved workflow/ref rejected) and the fork/PR test are process items around already-existing `if:` gates, and #1278 lifecycle-ledger reporting — all separate, larger follow-ups.

## Test plan
- [x] `yaml.safe_load` on the modified workflow — parses cleanly
- [x] Extracted and `bash -n`-checked the modified shell script — no syntax errors
- [x] Ran the `jq -n` predicate-construction command standalone with representative values — produces valid, correctly-shaped JSON
- [x] Confirmed the new step reuses the exact same `identity`/`issuer` variables already used (and working) for image signing and SBOM attestation — no new trust-boundary logic introduced
- [x] Grepped for other references to the renamed step name — none found

---
🐝 **Hive Agent**: `contributor` | **SHA:** `7c3ef7cb`

---
<a href="https://app.blacksmith.sh/tuna-os/codesmith/tunaOS/pr/1435"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>

### DIFF
--- a/.github/workflows/reusable-build-image.yml
+++ b/.github/workflows/reusable-build-image.yml
@@ -819,6 +819,8 @@ jobs:
       IMAGE_REGISTRY: ${{ vars.IMAGE_REGISTRY || format('ghcr.io/{0}', github.repository_owner) }}
       DEFAULT_TAG: ${{ inputs.default-tag }}
       INDEX_DIGEST: ${{ needs.manifest.outputs.digest }}
+      IMAGE_VARIANT: ${{ inputs.image-variant }}
+      FLAVOR: ${{ inputs.flavor }}
       COSIGN_YES: "true"
     steps:
       - name: Install Cosign
@@ -846,7 +848,7 @@ jobs:
           merge-multiple: true
           path: supply-chain/sboms
 
-      - name: Sign images and attest SBOMs
+      - name: Sign images and attest SBOMs + provenance
         env:
           REQUIRE_SBOM: ${{ inputs.sbom }}
         run: |
@@ -854,6 +856,14 @@ jobs:
           identity="https://github.com/${GITHUB_REPOSITORY}/.github/workflows/reusable-build-image.yml@${GITHUB_REF}"
           issuer="https://token.actions.githubusercontent.com"
           index_ref="${IMAGE_REGISTRY}/${IMAGE_NAME}@${INDEX_DIGEST}"
+          # https://tunaos.org/provenance/v1 fields: source commit, workflow/run,
+          # variant, flavor, platform, build-config revision (tunaos/tunaos#1187).
+          # Base-image digest is deliberately omitted: get-base-image.sh resolves
+          # a tag, not a digest, in the per-platform build job, and pinning it
+          # would mean adding a `skopeo inspect` call to that live 180-minute
+          # build matrix — not something to do blind in this change.
+          run_url="${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID}"
+          provenance_dir="$(mktemp -d)"
 
           cosign sign "$index_ref"
           cosign verify "$index_ref" \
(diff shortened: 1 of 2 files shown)


hanthor:
Leaving this open, but it needs a rework before it can land — flagging early rather than after a rebase.

The provenance predicate itself is the right thing and closes a named #1187 item. The problem is where it goes. This PR adds `cosign attest` + `cosign verify-attestation` per platform image **inside the `sign` job**, and #1748 is in the process of taking attestation *out* of that job for a measured reason:

On [run 31858324517](https://github.com/tuna-os/tunaOS/actions/runs/31858324517) every albacore flavor built, pushed and passed its gates, and all twelve Promote jobs were skipped — because `cosign attest` could not reach Rekor. Promote requires `sign`, so every Rekor round trip inside `sign` is another way for a transparency-log outage to block promotion of images that are already correct. #1748 splits it the way projectbluefin/dakota does: `sign` does image signatures only (a promotion precondition), `attest_sbom` runs `continue-on-error` off the critical path.

As written, this PR adds **two more Rekor round trips per platform per flavor** to the critical path — roughly doubling the exposure the split is removing.

The fix is small: move the provenance steps into `attest_sbom` alongside the SBOM attestation (or a sibling non-blocking job if you want them separable), and use the shared `cosign_retry` from `.github/scripts/cosign-retry.sh` rather than a bare invocation. The predicate construction, the identity/issuer reuse and the docs section all carry over unchanged.

Also worth keeping when you rebase: the explicit note about *not* pinning the base-image digest, and why. That is the right call and the reasoning should not get lost.

---

hanthor:
Keeping this open — provenance attestation is genuinely not on `main` yet, and this is a sound implementation.

Verified: the signing job on `main` still signs the index, signs each platform digest and attests SPDX SBOMs, but emits no provenance predicate. The `https://tunaos.org/provenance/v1` statement added here (source commit/repo, workflow, run id/url, variant, flavor, platform) plus the `cosign verify-attestation` round-trip is the missing piece of #1187, and the comment explaining why the base-image digest is deliberately omitted is the right call rather than adding a `skopeo inspect` into the 180-minute build matrix.

Blocked only by drift: it conflicts with `main` in `.github/workflows/reusable-build-image.yml`, which has changed substantially since 08-13 (corral boot gate, registry-mirror work). The `docs/VERIFY-ARTIFACTS.md` half applies cleanly.

I could not rebase it myself — this PR's head is on the `hanthor/tunaos-999` fork, which I have no write access to. Worth someone with that access re-cutting it against current `main`; also still a draft.

---

hanthor:
Closing as stale, not as wrong. Provenance attestation genuinely does not exist on main, so the idea stands.

The job this patches has been rewritten underneath it: #1560 moved SBOM attestation off the critical path, and the step now uses `cosign_retry` with a deadline and backoff, with different variable names (`attested`/`missing` where this branch has `signed`/`attested`). The overlap is about 120 lines of structural conflict, so resolving it means re-implementing against the new shape rather than rebasing.

Worth reopening as a fresh PR against current main, alongside #2282 which is about that same job's failure behaviour.

---

PR #1445, opened 2026-08-13
ci: retry cosign sign/attest on transient rekor 502s

hanthor:
## Summary

This weekly boot report (#1276) shows **build ❌ on nearly every single variant/flavor combo**, which looked like a mass regression. Investigating the actual failed runs (e.g. run 31663674297) found the real image builds mostly succeeded — the failures cluster in the `Sign + attest` and `Gate` jobs:

```
Error: signing ghcr.io/tuna-os/albacore@sha256:...: creating bundle: signing bundle:
error signing bundle: Post "https://rekor.sigstore.dev/api/v1/log/entries": giving up after
2 attempt(s): status 502: <!doctype html>...502 Bad Gateway
```

This is already tracked in #1377 (open since 2026-08-07, confirmed still recurring 2026-08-13 while investigating this report) — rekor.sigstore.dev has had multi-day intermittent 502s that outlast cosign's own internal single retry, and a `Gate`/`Sign + attest` failure fails the whole per-flavor job even though the actual OCI build succeeded. #1377 already specifies the exact fix (a retry wrapper around the sign/attest step) but notes the agent that filed it was blocked on `workflows` push permission — I have that via the normal fork+PR flow, so this PR implements it.

(Separately: a smaller cluster of `linux-amd64` build-step failures for `*-nvidia` flavors is a **different**, already-tracked regression — #1382/#1383, an `rpm-ostree kernel-install ... Invalid cross-device link` issue on nvidia/hwe kernels. Not touched by this PR.)

## Change

Wraps each `cosign sign`/`cosign attest` call in `reusable-build-image.yml`'s `sign` job in a `with_retry` helper: 3 attempts, 30s/60s backoff. Retries the individual failed call, not the whole step — cosign sign/attest are idempotent per digest (re-running against an already-signed ref is safe), and this avoids re-signing images that already succeeded in an earlier iteration of the loop.

## Test plan
- [x] `yaml.safe_load` — parses cleanly
- [x] Extracted and `bash -n`-checked the modified script — no syntax errors
- [x] Verified `until "$@"; do ... done` retry semantics standalone (a stub command that fails twice then succeeds resolves after exactly 3 invocations, matching the intended 3-attempt/30s-60s-backoff behavior)
- [x] Confirmed via `gh api search/issues` that #1377 already proposes this exact fix and is still open/recurring, so this isn't speculative — it's implementing an already-diagnosed, ready fix

---
🐝 **Hive Agent**: `contributor` | **SHA:** `dee84be8`

---
<a href="https://app.blacksmith.sh/tuna-os/codesmith/tunaOS/pr/1445"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>


hanthor:
Merged `main` into the branch to clear the conflict in `.github/workflows/reusable-build-image.yml`.

Resolution kept main's side: #1501 (`fix(sign): retry cosign past a transient Sigstore blip`) already landed a strict superset of this branch's change — the same rekor-502 retry, but as a `cosign_retry` helper wrapping *every* cosign invocation (`sign`, `verify`, `attest`, `verify-attestation`) rather than just sign/attest, using the `MAX_RETRIES`/linear-backoff idiom the "Push to GHCR" step already uses, plus `tests/test_sign_step_retries_cosign.py` pinning the behavior. This branch's `with_retry` was dropped as the duplicate.

The PR is now mergeable (`clean`) but has a **0-file diff** against `main`, so there is nothing left to merge — it can be closed as already fixed by #1501.

PR #1489, opened 2026-08-13
fix(ci): disable weekly schedule trigger on installer-smoke until EGL window readiness is resolved (#1396)

### LINKED ISSUE #1396: [ci-maintainer] Installer Smoke now weekly-scheduled and red across board: installer runs but never reports a window (EGL dri2 failure, 22+ jobs)
## CI Issue

The `Installer Smoke` workflow is **now weekly-scheduled** (`cron: "0 9 * * 3"`, first run 31585741267 on 08-12 10:03) and the first scheduled run is **failing across the board**: 25+ of 46 jobs failed, **zero successes** among completed jobs, and the remaining jobs hang past 1h (the historical time-out shape). The previous runs were `workflow_dispatch`-only (last: 08-08), so this is the first automatic exposure.

**New dominant failure mode (22 jobs): `Assert live desktop session + installer frontend`**
```
::error::gnome: org.bootcinstaller.Installer is running but never reported a window
The process is alive (check 2 passed) and no readiness stamp
```
The installer flatpak runs (`bootc-installer` python process alive via bwrap) but **never renders a window**. VM journal shows `libEGL warning: egl: failed to create dri2 screen` — the installer cannot get a display/GPU surface under the OVMF VM, so the window-reached-screen assertion added in #1173 (2fb29e8, 08-09) can never pass.

**Secondary failure modes:**
- `Build dev ISO` (3 jobs): skipjack:xfce fails on the tracked gtkgreet checksum mismatch; hummingbird:base fails `flatpak not installed; cannot pre-install org.bootcinstaller.Installer`
- nvidia variants fail at earlier build stages (dracut regression #1382)

## Evidence

- Run 31585741267 (08-12 10:03, schedule): 25 failures / 0 successes / 4 in_progress at 1h15m
- yellowfin:gnome job 94079137386: `org.bootcinstaller.Installer is running but never reported a window`; `libEGL warning: egl: failed to create dri2 screen`; `STAMP_SIGNAL=<none>`
- Workflow diff: `2fb29e8` added window-reached-screen assertion; schedule trigger added after 08-10 (`cron 0 9 * * 3`)

## Recommendation

1. **Investigate window rendering in the smoke VM first**: `egl: failed to create dri2 screen` suggests the installer needs a working GL/EGL surface (llvmpipe/software rendering or a VLM backend) that the OVMF VM does not provide. Check whether the frontend's readiness stamp is written after first-frame or only after mapping — the smoke harness and the app disagree on what "ready" means.
2. If the assertion is correct, the fix is in the ISO/frontend (ensure the installer renders under software GL); if the VM cannot host it, the smoke should document the limitation and skip window assertion for flavors that use EGL-dependent frontends.
3. The weekly cadence will keep failing loudly every Wednesday until this is resolved — consider downgrading to `workflow_dispatch` while the window assertion is debugged, or add the run to a triage rotation.

---

---

**Status (2026-08-14):** PR #1489 implements recommendation #3 exactly
(disabled the weekly `schedule` trigger, kept `workflow_dispatch`) — open,
clean, mergeable, unchanged since I opened it.

Recommendations #1/#2 (root-cause the `egl: failed to create dri2 screen`
failure — whether it's a software-GL/llvmpipe gap in the OVMF VM, or a
readiness-stamp timing mismatch between the smoke harness and the
installer frontend) need live iteration against a real QEMU/OVMF VM to
diagnose safely — not something to guess at from static log reading. Not
attempted here; recommend whoever picks this up next have that
environment available, same reasoning as this session's other
build/VM-infra findings that needed real hardware/tooling to verify.

hanthor:
Fixes #1396.

### Summary
- Temporarily commented out the weekly `schedule` trigger (`cron: "0 9 * * 3"`) in `.github/workflows/installer-smoke.yml`.
- Prevents recurring weekly CI failures across 46 jobs while window readiness under OVMF software GL (`libEGL warning: egl: failed to create dri2 screen`) is investigated and resolved.
- Retains manual `workflow_dispatch` execution.

---
<a href="https://app.blacksmith.sh/tuna-os/codesmith/tunaOS/pr/1489"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>

### DIFF
--- a/.github/workflows/installer-smoke.yml
+++ b/.github/workflows/installer-smoke.yml
@@ -29,9 +29,9 @@ on:
         # this default kept it from ever running.
         default: "gnome,kde,cosmic,niri,xfce"
         type: string
-  schedule:
-    # Weekly, after the Tuesday image builds settle.
-    - cron: "0 9 * * 3"
+# schedule:
+  #   # Weekly, after the Tuesday image builds settle. Disabled until window readiness stamp under OVMF software GL is resolved (#1396)
+  #   - cron: "0 9 * * 3"
 
 permissions:
   contents: read


hanthor:
Holding this one — it turns off the only axis that answers the question it exists to answer.

`docs/MATRIX-STATUS.md` is explicit about what installer smoke uniquely covers: *"This is the only axis that checks a human could actually install."* LUKS E2E drives fisherman over SSH and never looks at the screen; Desktop Contract Sweep doesn't boot; Build proves neither. Silencing the schedule means the installer GUI stops being observed at all, and the failure mode that document was written about — `yellowfin:cosmic` sitting LUKS-green for weeks while its installer GUI had never once been seen — is exactly what returns.

The 46 weekly failures are real noise and worth fixing. But the same document already explains most of them, and it isn't a defect in the product:

> cosmic, niri, xfwl4 **and kde** need a DRM render node. GitHub runners have none, so on hosted CI the compositor never starts — no configuration change alters this.

So a large share of those 46 jobs are a harness limitation that will still be there whenever the schedule is switched back on. Waiting for "EGL window readiness to be resolved" waits for something that hosted runners cannot provide.

Two alternatives that keep the signal:

1. **Narrow the matrix instead of the trigger.** Keep the weekly schedule, restrict it to `gnome` — the one desktop MATRIX-STATUS says is still thought verifiable without a render node, and which it also notes *has never actually been run*. That converts 46 noisy jobs into a handful that answer an open question.
2. **Keep it running, stop it alarming.** `continue-on-error` on the render-node-dependent flavors, so the results still land in the sweep artifact and `gen-matrix-status.py` keeps seeing them, without a red check every Wednesday.

Either way the cells stay ❌ (a run asserted it and it failed) rather than decaying to ⬜ (never tested) — which matters, because #1570's exit criteria and the generator both read that distinction.

If the intent is a deliberate, time-boxed pause rather than an indefinite one, say so in the workflow comment with the condition for turning it back on, and I'll merge it as written. Right now the comment says "until resolved" for something that may not be resolvable on this runner class.

---

PR #1495, opened 2026-08-13
chore(ci): measure arm64 ulimits before touching ghcr-login

hanthor:
Diagnostic only — no behaviour change. This produces one table and then gets deleted.

## The failure it targets

`Login to Podman (sudo)` fails in ~2 seconds on `ubuntu-24.04-arm`:

```
Error: failed to open 2048 locks in /libpod_lock: numerical result out of range
```

exit 125. In the **same job**, `Login to Podman (user)` — the same login, the same `ulimit -n 4096`, without sudo — reports `Login Succeeded!`.

Each failure costs an arm64 image, which fails `Manifest` and skips `Sign`, `Gate` and `Promote`. Across yellowfin and grouper that is roughly **six cells** of the README matrix.

## Why a probe instead of a fix

`.github/actions/ghcr-login/action.yml` **already documents this exact error** and already carries `ulimit -n 4096` as the remedy. That remedy works unsudoed and doesn't work under sudo, and I don't know why yet.

The plausible explanations predict different fixes:

| Story | Fix it implies |
| :--- | :--- |
| sudo/PAM caps the hard limit below the soft limit being requested | raise `ulimit -Hn` first, or use `prlimit` |
| `ulimit -n 4096` is *rejected* outright, so the `&&` chain never reaches the login | same, but the observed error would be a red herring |
| podman wants fewer locks, not more descriptors | pin `num_locks` in `containers.conf` — the repo currently pins none |

`ghcr-login` runs in **every build in this repository**. A wrong guess there is a repo-wide outage, not a failed experiment.

## What it measures

On `ubuntu-24.04-arm`, with `ubuntu-24.04` as the control (the same action succeeds there today):

- soft and hard `nofile` in the plain job shell
- soft and hard `nofile` under `sudo bash -c`
- whether the sudo shell will *accept* `ulimit -n 4096`, capturing the rejection message if not — this separates "the ulimit was refused" from "podman ran and disliked what it got"
- whether `sudo podman info` reproduces the lock-file failure. `podman login` needs `/libpod_lock` before it needs a registry, so `info` exercises the same init path with no credentials involved
- which `containers.conf` files exist and whether any sets `num_locks`

Results land in the job summary as a table.

## Trigger

`workflow_dispatch`, plus a `pull_request` trigger **scoped to this file's own path**. GitHub doesn't offer `workflow_dispatch` for a workflow that isn't on the default branch yet, so dispatch-only would mean merging a diagnostic before it can diagnose anything. Path-scoped, it runs exactly once — on this PR — and never again.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Gu5fBnu8whXPqCFKd5aA7B)_

---
<a href="https://app.blacksmith.sh/tuna-os/codesmith/tunaOS/pr/1495"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>

### DIFF
--- a/.github/workflows/arm64-ulimit-probe.yml
+++ b/.github/workflows/arm64-ulimit-probe.yml
@@ -1,0 +1,92 @@
+# Diagnostic only. Measures the file-descriptor limits podman actually sees on
+# an arm64 runner, under the plain shell and under sudo, before anyone edits
+# .github/actions/ghcr-login. That action is on the critical path of every
+# build in this repository, so a guess pushed straight into it is a repo-wide
+# outage if the guess is wrong.
+#
+# What we are chasing: "Login to Podman (sudo)" fails in ~2s on
+# ubuntu-24.04-arm with
+#   Error: failed to open 2048 locks in /libpod_lock: numerical result out of range
+# (exit 125) while "Login to Podman (user)" -- the same login, same
+# `ulimit -n 4096`, without sudo -- reports "Login Succeeded!" in the same job.
+#
+# Delete this workflow once the fix lands; it exists to produce one table.
+name: arm64 ulimit probe (diagnostic)
+
+on:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  probe:
+    name: ulimits on ${{ matrix.runner }}
+    runs-on: ${{ matrix.runner }}
+    strategy:
+      fail-fast: false
+      matrix:
+        # x64 is the control: the same action succeeds there today, so any
+        # difference in these numbers is the thing to explain.
+        runner: [ubuntu-24.04-arm, ubuntu-24.04]
+    steps:
+      - name: Limits as the job shell sees them
+        run: |
+          echo "### ${{ matrix.runner }}" >> "$GITHUB_STEP_SUMMARY"
+          printf '| context | soft nofile | hard nofile |\n| :-- | --: | --: |\n' \
+            >> "$GITHUB_STEP_SUMMARY"
+          printf '| plain shell | %s | %s |\n' "$(ulimit -Sn)" "$(ulimit -Hn)" \
+            >> "$GITHUB_STEP_SUMMARY"
+          echo "plain: soft=$(ulimit -Sn) hard=$(ulimit -Hn)"
+
+      - name: Limits under sudo
+        run: |
+          # Read them the way the failing step's shell would see them.
+          sudo bash -c 'printf "sudo: soft=%s hard=%s\n" "$(ulimit -Sn)" "$(ulimit -Hn)"'
+          soft=$(sudo bash -c 'ulimit -Sn')
+          hard=$(sudo bash -c 'ulimit -Hn')
+          printf '| sudo shell | %s | %s |\n' "$soft" "$hard" >> "$GITHUB_STEP_SUMMARY"
+
+      - name: Can the sudo shell actually raise it to 4096?
+        run: |
+          # This is the exact construct ghcr-login uses. `&&` means a failure
+          # here stops the login from ever running, so distinguish "the ulimit
+          # was rejected" from "podman ran and disliked what it got".
+          if sudo bash -c 'ulimit -n 4096' 2>/tmp/ulimit.err; then
+            echo "sudo ulimit -n 4096: OK"
+            echo '| sudo `ulimit -n 4096` | accepted | |' >> "$GITHUB_STEP_SUMMARY"
+          else
+            echo "sudo ulimit -n 4096: REJECTED"
+            cat /tmp/ulimit.err
+            echo "| sudo \`ulimit -n 4096\` | REJECTED: $(cat /tmp/ulimit.err) | |" \
+              >> "$GITHUB_STEP_SUMMARY"
+          fi
+          # And what the process actually ends up with if it is accepted.
+          sudo bash -c 'ulimit -n 4096 2>/dev/null; printf "after raise: soft=%s hard=%s\n" "$(ulimit -Sn)" "$(ulimit -Hn)"' || true
+
+      - name: What podman itself reports
+        run: |
+          echo "--- podman version ---"
+          podman --version || true
+          sudo podman --version || true
+          echo "--- num_locks in effect ---"
+          podman info --format '{{.Host.Slirp4netns}}' >/dev/null 2>&1 || true
+          echo "--- containers.conf files present ---"
+          for f in /etc/containers/containers.conf \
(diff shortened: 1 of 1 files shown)


hanthor:
Probe ran ([31704227929](https://github.com/tuna-os/tunaOS/actions/runs/31704227929)). Results, and they refute the leading hypothesis.

**`ubuntu-24.04-arm`**, image `20260810.90.1`, podman `5.8.4`:

| context | soft nofile | hard nofile |
| :--- | ---: | ---: |
| plain shell | 65536 | 65536 |
| **sudo shell** | **1024** | **1048576** |
| sudo, after `ulimit -n 4096` | 4096 | 4096 |

- `sudo bash -c 'ulimit -n 4096'` → **accepted**
- `sudo bash -c 'ulimit -n 4096 && podman info'` → **OK**
- `containers.conf`: `/usr/share/containers/containers.conf` has `#num_locks = 2048` **commented out**; `/etc/containers/containers.conf` has an `[engine]` section but sets no `num_locks`

## What this rules out

Two of the three candidate explanations are dead:

1. **"sudo/PAM caps the hard limit below 2048"** — no. The hard limit under sudo is **1048576**.
2. **"`ulimit -n 4096` is rejected, so the `&&` chain never reaches the login"** — no. It's accepted, and the process ends up at 4096.

## What's real, and what isn't explained

The asymmetry is genuine and worth recording: **sudo's soft `nofile` is 1024 against 65536 unsudoed.** That is precisely why the sudo path needs an explicit `ulimit` and the user path doesn't, and 1024 is below the 2048 locks podman wants. So the existing fix targets the right thing.

But **with that fix in place the failure does not reproduce.** `sudo podman info` succeeds here.

That means the production failure is image-version-dependent or context-dependent, not a static property of the action. Next step is to pull the runner image and podman versions from the header of the actual failing job in run 31663727643 and diff them against `20260810.90.1` / podman 5.8.4.

I'm not editing `ghcr-login` until that comparison explains the delta. It runs in every build in this repo, and "I couldn't reproduce it so here's a fix anyway" is how you find out at 04:00 that you broke all of them.

Holding this PR open as the record of the measurement.

---

hanthor:
Second run ([31704510288](https://github.com/tuna-os/tunaOS/actions/runs/31704510288)) — the control, with the `ulimit` removed. It kills the third hypothesis as well, and turns up something more useful.

```
sudo default soft limit: 1024
sudo podman info WITHOUT ulimit: OK -- descriptor story does not hold
```

**podman works fine at soft=1024.** So `ulimit -n 4096` is not what makes the sudo login work — at least not on this runner.

## The actual finding: the arm64 fleet is not homogeneous

The two probe runs, same `ubuntu-24.04-arm` label, ~3 minutes apart:

| Run | Runner image | podman | `/etc/containers/containers.conf` | `podman info` w/o ulimit |
| :--- | :--- | :--- | :--- | :--- |
| 31704227929 | `20260810.90.1` | **5.8.4** | present | not tested |
| 31704510288 | `20260719.67.1` | **4.9.3** | **absent** | **OK** |

Different image, different podman major version, different config files present. That is exactly the shape of a failure that hits ~6 cells intermittently and reads as a flake.

This repo already documented this same heterogeneity, for a *different* podman failure — `reusable-build-image.yml`:

> *"Newer ubuntu-latest images ship podman under /usr/local … so `apt-get install podman` happily adds Ubuntu's podman 4.9.3 on top of it … That killed sailfin niri's Gate in run 30594521039 while kde/xfce/gnome passed in the same run, purely because their runners already had podman apt-installed, making the install a no-op. It looks like a flake and is not: it fires whenever a job lands on the newer image."*

Same two versions, 4.9.3 and the newer bundled one. Strongly suggests the `libpod_lock` failure is a property of **podman 5.8.4** (or of the `/etc/containers/containers.conf` that ships alongside it), not of arm64 or of sudo.

## What's still missing

The one cell not yet filled: **podman 5.8.4 without the ulimit.** Run 1 had 5.8.4 but predated this control step; run 2 had the control step but landed on 4.9.3. Runner assignment is not selectable, so this needs re-runs until it lands on 5.8.4 — or a version check inside the probe that reports which combination it measured.

Until that cell is filled I still won't touch `ghcr-login`. Three hypotheses have now been measured and all three were wrong; a fourth guess is not worth a repo-wide outage.

---

hanthor:
Third run ([31704945484](https://github.com/tuna-os/tunaOS/actions/runs/31704945484)), six arm64 samples. The missing cell is filled and it kills the last hypothesis too.

| samples | runner image | podman | `sudo podman info` **without** ulimit |
| ---: | :--- | :--- | :--- |
| 3× | `20260707.563` | 4.9.3 | **OK** |
| 3× | `20260729.566` | **5.8.4** | **OK** |
| 1× (x64 control) | `20260707.563` | 4.9.3 | **OK** |

The fleet split is confirmed with a clean 3/3 sample — two runner images and two podman major versions behind one label. But **podman initialises `/libpod_lock` fine at soft=1024 on both.**

## Every hypothesis is now refuted

| # | Hypothesis | Verdict |
| --: | :--- | :--- |
| 1 | sudo/PAM caps the hard limit below 2048 | **No** — hard is 1048576 |
| 2 | `ulimit -n 4096` is rejected, `&&` short-circuits | **No** — accepted, lands at 4096 |
| 3 | 1024 descriptors is too few for 2048 locks | **No** — OK on 4.9.3 |
| 4 | podman 5.8.4 needs more descriptors than 4.9.3 | **No** — OK on 5.8.4 too |

So `ulimit -n 4096` in `ghcr-login` is **not** what makes the sudo login work, on either podman. That is worth knowing on its own: the comment in that action attributes the fix to a cause that doesn't reproduce.

## What must differ, then

The probe runs `sudo podman info` in a clean job. Production runs it inside `ghcr-login`, **after** `docker/login-action` and after `Login to Podman (user)` — and in build jobs, after `container-storage-action` has already set up storage and possibly started containers.

So the next experiment isn't another synthetic limit check. It's to invoke the **real action, in a real build job's context**, on arm64, and see whether the failure appears when the preceding steps have run. Candidate mechanisms worth instrumenting there: descriptors already consumed by the runner/containers at that point in the job, or rootless podman state created by the *user* login changing what the *sudo* login then does.

Four measured refutations is a good reason to keep not guessing. `ghcr-login` runs in every build in this repo.

---

PR #1539, opened 2026-08-14
ci: set TBOX_CUSTOMIZE_NETWORK=host for weekly desktop screenshots

### LINKED ISSUE #1412: [ci-maintainer] Weekly Desktop Screenshots red every week since 06-22: DNS resolution fails inside tacklebox live-customize container (github.com / mirrors)
## CI Issue

The `Weekly Desktop Screenshots` workflow (weekly-desktop-screenshots.yml, Mon 02:00 UTC) has failed on every scheduled run since at least 06-22 — no desktop screenshot has been published to R2 (`download.tunaos.org/screenshots/*`) or committed to docs/images/desktops for ~2 months. This is currently a silent gap: no bead or issue tracks it.

## Evidence

All jobs fail at `Build live ISO (from published GHCR image)` → `live customize` step, where the tacklebox podman run cannot resolve external hosts:

| run | date | error |
|---|---|---|
| 31353454671 | 08-10 | `curl: (6) Could not resolve host: github.com` (albacore-gnome etc., all variants) |
| 30787528848 | 08-03 | `curl: (6) Could not resolve host: github.com` (yellowfin-gnome) |
| 30240462131 | 07-27 | `Could not resolve host: kitten.mirrors.almalinux.org` (yellowfin-cosmic) |
| 29226444083 | 07-13 | failure (same step) |
| 28772022617 | 07-06 | failure |
| 28353460544 | 06-29 | failure |
| 27935967968 | 06-22 | failure |

Varying hosts (github.com, kitten.mirrors.almalinux.org) point to DNS failing inside the tacklebox `podman run ... --entrypoint /bin/bash <image> -c set -eu; cd /run/tbox-customize/1 && bash ./customize-live.sh` container on ubuntu-24.04 runners, not a single broken upstream.

## Recommendation

- Investigate DNS inside the tacklebox customize container: verify /etc/resolv.conf propagation, consider `--network host` for the customize run, and add retries to the curl calls in customize-live.sh.
- Until fixed, consider splitting the matrix so one failing variant does not red the whole weekly run (fail-fast is already false; the workflow still reports red and publishes nothing).

---
---
🐝 **Hive Agent**: `ci-maintainer` | **Instance:** `hive-keen-mink` | **SHA:** `unknown`

hanthor:
Fixes #1412.

Weekly Desktop Screenshots has been red every run since 06-22, always at the `Build live ISO` step, always with a DNS resolution failure inside the tacklebox live-customize container — but the *host* it fails to resolve varies run to run (`github.com` on 08-10/08-03, `kitten.mirrors.almalinux.org` on 07-27). A varying-host pattern like that points at something structurally wrong with the container's network path rather than one flaky remote.

Traced the actual code path: this workflow builds via `just iso` → `scripts/build-iso-tacklebox.sh` → the separate `tuna-os/tacklebox` Go tool, which runs `live_customize` scripts (this repo's `customize-live.sh`) inside a podman container. tacklebox's `internal/install/customize.go` already documents this exact failure mode:

```go
// The customize scripts need outbound network (flatpak install above all),
// and podman's default bridge is not always usable: on a host whose
// netavark/firewalld rules have gone stale, root-context containers get an
// interface with no route out while rootless ones work fine. The symptom is
// remote — "Could not resolve hostname tunaos.org" from inside flatpak —
// and the build then produces an ISO with no installer on it.
// TBOX_CUSTOMIZE_NETWORK=host is the escape hatch; unset keeps podman's
// default, so nothing changes for hosts that are fine.
if net := strings.TrimSpace(os.Getenv("TBOX_CUSTOMIZE_NETWORK")); net != "" {
    runArgs = append(runArgs, "--network", net)
}
```

That's an exact match for the symptom in this issue. I grepped every workflow in this repo that calls `just iso` (`publish-iso-groups.yml`, `installer-screenshots.yml`, `live-iso-bootc.yml`, `installer-smoke.yml`, `luks-e2e.yml`) and none of them set `TBOX_CUSTOMIZE_NETWORK` — the escape hatch exists upstream but nothing here was flipping it on.

## Change

Set `TBOX_CUSTOMIZE_NETWORK=host` alongside the existing `GITHUB_REPOSITORY_OWNER` env on the `sudo ... just iso ...` invocation in `weekly-desktop-screenshots.yml` only — scoped to the workflow that's actually failing, not a blanket change across every `just iso` caller.

## Verification

- Confirmed via `grep` there's no existing `TBOX_CUSTOMIZE_NETWORK` usage anywhere in this repo (workflows, scripts, Justfile) — this wires in an unused-but-already-shipped upstream option, not new tacklebox behavior.
- YAML validated with `yaml.safe_load`.
- Could not run the actual weekly build in this environment (needs nested KVM + real GHCR pull); this is a config-only change with a code-documented, direct mechanism, but the next scheduled/dispatched run of this workflow is the real verification. Flagging that honestly rather than claiming an end-to-end pass I couldn't produce.

---
🐝 **Hive Agent**: `contributor` | **SHA:** `58873ba5`

### DIFF
--- a/.github/workflows/weekly-desktop-screenshots.yml
+++ b/.github/workflows/weekly-desktop-screenshots.yml
@@ -107,7 +107,13 @@ jobs:
         env:
           GITHUB_REPOSITORY_OWNER: tuna-os
         run: |
-          sudo GITHUB_REPOSITORY_OWNER=tuna-os \
+          # TBOX_CUSTOMIZE_NETWORK=host works around tuna-os/tacklebox's documented
+          # podman-bridge DNS issue for root-context containers (see tacklebox's
+          # internal/install/customize.go): the live-customize container can end up
+          # with no route out on GitHub-hosted runners, surfacing as "Could not
+          # resolve host" for whatever host the customize scripts happen to hit
+          # first (github.com, mirrors.almalinux.org, ...) — tuna-os/tunaos#1412.
+          sudo GITHUB_REPOSITORY_OWNER=tuna-os TBOX_CUSTOMIZE_NETWORK=host \
             just iso "${{ matrix.variant }}" "${{ matrix.flavor }}" ghcr "${{ matrix.flavor }}"
 
       - name: Boot ISO and capture desktop screenshot


hanthor:
Closing as a duplicate of #1518, which is now merged.

Both branches make the same fix for #1412 — adding `TBOX_CUSTOMIZE_NETWORK=host` to the `sudo … just iso` invocation in `weekly-desktop-screenshots.yml`. #1518 was merged because it also adds a Bats guard (`tests/bats/test_customize_live.bats`) so the workaround cannot be silently dropped again.

One thing to be aware of: that guard greps for the literal `sudo TBOX_CUSTOMIZE_NETWORK=host`, so the variable has to stay in that position on the line. This branch's ordering (`sudo GITHUB_REPOSITORY_OWNER=tuna-os TBOX_CUSTOMIZE_NETWORK=host`) would not match it.

The fix is on `main`; #1412 can be verified against the next weekly run.

---

PR #1575, opened 2026-08-14
ci: skip doomed dnf retries on resolver-level failures (#1555)

hanthor:
## Summary

Issue #1555: Hummingbird has been red for 10 straight nightlies (08-05..08-14). Part of the failure is niri/kde dnf dependency resolution failures (`libjsoncpp.so.26`/`librhash.so.1` missing — `nothing provides X`), which currently burn all 4 retry attempts with backoff sleep before failing, wasting significant CI time on an error class that is deterministic and can never succeed on retry.

## Change

`build_scripts/lib.sh`'s `dnf_retry()` now detects resolver-level failures (`nothing provides`, `No match for argument`, `Problem: conflicting requests`) via `tee`-captured output + `${PIPESTATUS[0]}`, and short-circuits to immediate failure instead of retrying. Genuine transient failures (network/mirror issues) still retry with backoff as before.

## Testing

Verified locally with a mock `dnf` script covering 4 scenarios: unresolvable failure fails fast, transient failure still retries and succeeds, clean success passes through, and no temp files are leaked. All passed.

Refs #1555
---
🐝 **Hive Agent**: `contributor` | **SHA:** `9db41e87`

### DIFF
--- a/build_scripts/lib.sh
+++ b/build_scripts/lib.sh
@@ -417,19 +417,49 @@ pkg_clean() {
 # those fail identically on every attempt; the loop returns the last DNF
 # exit code so callers still see real errors.
 #
+# It used to retry those identically-failing transactions anyway, though:
+# `nothing provides X` / `No match for argument` are dnf's own resolver
+# telling you the *repo content* can't satisfy the request, which four
+# attempts and `dnf clean metadata` cannot change — clearing metadata just
+# re-downloads the same repodata that already says the provider is missing.
+# Measured on tuna-os/tunaos#1555 (Hummingbird's public-hummingbird repo
+# missing libjsoncpp.so.26/librhash.so.1 providers): ~50s of pure sleep
+# wasted per doomed call, three separate calls in one job. Detecting that
+# class of error and returning immediately doesn't change any caller's
+# pass/fail outcome (the transaction was always going to fail) — it just
+# stops paying for retries that can't work.
+#
 # Usage: dnf_retry install -y foo bar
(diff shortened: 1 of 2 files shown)


hanthor:
Good change — blocked only on a conflict. `mergeable_state` is `dirty` against current `main`; rebase and I'll merge it.

Fast-failing on `nothing provides` / `No match for argument` / `Problem: conflicting requests` is right: those are resolver verdicts, not transport failures, and no number of retries changes a repository's dependency graph. The `tee` + `${PIPESTATUS[0]}` approach preserves the exit status correctly, and keeping genuine transient failures on the existing backoff is the right split.

Two notes for the rebase:

**It saves time, it doesn't fix #1555.** Worth being explicit in the PR body, because the issue title is "Build Hummingbird red 10 straight nightlies" and this makes those nightlies fail *faster*. `libjsoncpp.so.26` and `librhash.so.1` still have no provider — the variant stays red until that's resolved upstream or the package set changes. That's a real win (10 nightlies × 4 attempts × backoff is a lot of runner time) but the issue shouldn't close on this.

**Worth a look while you're in `dnf_retry()`:** `docs/MATRIX-STATUS.md` notes that `install_available` skips unresolvable packages by design, so missing packages never fail a build — they only land in `/usr/share/tunaos/missing-on-*.txt` inside the image, which nothing reads on a schedule. If hummingbird's niri/kde sets go through that path rather than a hard `dnf install`, the fast-fail may not be reached at all for some of them. Not a blocker for this PR, just the thing to check before assuming the retry burn is gone.

---

Issue #1560, opened 2026-08-14
[ci] Sign+attest: cosign attest fails on rekor.sigstore.dev 502 — ~30s retry window vs 45+ min outage, blocks Promote for green builds on 7 variants


hanthor:
## Symptom

On the 2026-08-14 nightlies, **25 "Sign + attest" jobs failed across 7 variants** with the identical error, after their builds, manifests, and image signing all succeeded. This blocked Promote for every affected flavor — many otherwise fully-green images lost their published tag update:

- yellowfin (run [31766492172](https://github.com/tuna-os/tunaOS/actions/runs/31766492172)): base, cosmic-nvidia, niri-nvidia, cosmic-hwe, xfce-hwe, xfce-nvidia, gnome-hwe (7 jobs)
- albacore (run [31766397923](https://github.com/tuna-os/tunaOS/actions/runs/31766397923)): 9 jobs
- skipjack (run [31766408724](https://github.com/tuna-os/tunaOS/actions/runs/31766408724)): 4 jobs
- grouper (run [31766671286](https://github.com/tuna-os/tunaOS/actions/runs/31766671286)): base, gnome-asahi → cascaded into skipping **all** grouper stage-2 flavors
- flounder-sid (run [31766504862](https://github.com/tuna-os/tunaOS/actions/runs/31766504862)): base → skipped all stage-2
- bonito / bonito-rawhide (runs [31766548793](https://github.com/tuna-os/tunaOS/actions/runs/31766548793), [31766505527](https://github.com/tuna-os/tunaOS/actions/runs/31766505527)): gnome-hwe on each

```
Error: signing ghcr.io/tuna-os/yellowfin@sha256:88010b...: creating bundle:
signing bundle: error signing bundle: Post
"https://rekor.sigstore.dev/api/v1/log/entries": giving up after 2 attempt(s):
status 502: ...502 Bad Gateway
##[warning]cosign failed (attempt 3/3); retrying...
##[error]Process completed with exit code 1.
```

## Root cause

[...]

Issue #1567, opened 2026-08-14
[guppy] base amd64 runner dies during Syft SBOM scan — step timeout never fires, no logs uploaded, recurring Gentoo-specific resource exhaustion


hanthor:
## Symptom

guppy `base / linux-amd64` (2026-08-14 run [31766322789](https://github.com/tuna-os/tunaOS/actions/runs/31766322789), job 94662903686) built its image (03:18–03:35) and pushed to GHCR successfully (03:37:39), then died during **"Generate SBOM"**: the step stayed `in_progress` forever (no `completed_at`), the job concluded `failure` ~51 minutes later, and the runner never uploaded logs (`get_job_logs` → HTTP 404). Base Manifest then failed its "Exit on failure" fan-in and all guppy stage-2 flavors (gnome, kde, xfce) were skipped — a full 0/N night for the variant.

## Root cause (inferential — no logs exist, which is itself evidence)

The step has `timeout-minutes: 20` (`.github/workflows/reusable-build-image.yml:434`), but a step timeout is enforced by the runner agent; it can only fail to fire if the agent itself is wedged (memory/disk thrash), after which the Actions service kills the job for lost communication and no logs upload. This is the exact failure shape this variant already documented: the comment at `reusable-build-image.yml:431-433` cites #956 ("the Gentoo base job spends its remaining runner time in Syft"), previously mitigated with `--scope squashed` plus advisory soft limits `GOMEMLIMIT: 3GiB` / `GOGC: 25` (`:441-442`). GOMEMLIMIT is advisory — a huge Gentoo file inventory can still OOM-thrash the 7 GB runner.

## Suggested fix (in rough order of preference)

1. Wrap the syft invocation in a hard external bound that survives agent thrash: `sudo systemd-run --scope -p MemoryMax=4G ...` and/or shell `timeout 18m`, so the step fails fast with logs instead of killing the runner.
2. Generate the SPDX from the already-produced package manifest for Gentoo instead of re-scanning the image with syft.
3. Decouple SBOM generation from the build result (the image was already pushed!): `continue-on-error` + a separate sign/attest gate, so a dead SBOM scan doesn't turn a published build into a red variant.

⚠️ These lines are in `.github/workflows/reusable-build-image.yml` — per #1557 the hive App can't push workflow files, so this needs the maintainer merge path.

## Determinism

Mechanically a resource flake, but recurrent and Gentoo-specific (#956 precedent); expect it to repeat until bounded.

[...]

PR #1578, opened 2026-08-14
ci: extend cosign retry to survive a real Rekor outage (#1560)

hanthor:
## Summary

Issue #1560: on 2026-08-13 and 2026-08-14, `rekor.sigstore.dev` returned 502s for 45+ minutes each night. 25 "Sign + attest" jobs across 7 variants failed after their builds, manifests, and image signing had already succeeded — blocking Promote for otherwise fully-green flavors, including cascading skips of all stage-2 flavors for grouper and flounder-sid.

The existing `cosign_retry` wrapper does 3 attempts with 5/10/15s linear backoff (~30s total), sized for "a transparency-log blip that lasts minutes" per its own comment. That's two orders of magnitude short of the outage actually observed.

## Change

`.github/workflows/reusable-build-image.yml`, "Sign images and attest SBOMs" step:

- `cosign_retry` now uses exponential backoff capped at 90s between attempts, 11 attempts total — ~10 min per call before giving up, long enough to ride out the outages seen on both nights.
- Added an explicit 45-minute timeout to the `sign` job as a backstop, since the job makes several sequential cosign calls (index sign/verify, then per-platform sign/verify/attest/verify-attestation) and a pathological all-502 run could otherwise run against the implicit 360-minute GitHub Actions default.

This is fix option 1 from the issue (lengthen the retry window) — the lowest-risk of the three proposed. Options 2 (decouple SBOM attestation from the transparency log) and 3 (make Sign+attest independently re-runnable) change cosign's security posture / job graph in ways that deserve their own review, so this PR doesn't attempt them.

## Note

Per #1557, the hive App can't push `.github/workflows/` changes directly — this is a normal fork PR for the maintainer merge path.

Refs #1560
---
🐝 **Hive Agent**: `contributor` | **SHA:** `323e697b`

### DIFF
--- a/.github/workflows/reusable-build-image.yml
+++ b/.github/workflows/reusable-build-image.yml
@@ -810,6 +810,12 @@ jobs:
       inputs.publish && github.event_name != 'pull_request' &&
       needs.manifest.result == 'success'
     runs-on: ubuntu-latest
+    # A single Sign+attest job makes several sequential cosign calls (index
+    # sign/verify, then per-platform sign/verify/attest/verify-attestation).
+    # The retry budget below is ~10 min per call, worst case if every call
+    # in the job independently exhausts it (#1560) -- this is a backstop so
+    # a pathological all-502 run fails the job instead of hanging the workflow.
+    timeout-minutes: 45
     permissions:
       contents: read
       packages: write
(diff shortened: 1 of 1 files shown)


hanthor:
Superseded by #1572, which merged and already fixed this exact cosign_retry sizing (issue #1560, also referenced by #1567's syft-timeout fix in the same PR). Closing to avoid a duplicate/conflicting change on the same lines.

Issue #1563, opened 2026-08-14
[marlin] cachyos overlay keeps two kernels but builds one initramfs into a `find | tail -1` gamble — all 5 cachyos flavors fail Gate (bootc "initramfs not found" / boot never reaches graphical.target)


hanthor:
## Symptom

First nightly with the new overlay matrix (2026-08-14, run [31766586852](https://github.com/tuna-os/tunaOS/actions/runs/31766586852)): **all 5 `*-cachyos` flavors failed their Gate** while all 5 `*-nvidia` flavors passed. Two presentations, one cause:

**gnome-cachyos, cosmic-cachyos, kde-cachyos** — `just qcow2` (`reusable-build-image.yml:1011-1015`) dies during install:
```
Installing bootloader via systemd-boot
error: Installing to disk: Setting up composefs boot: Setting up BLS boot:
Computing boot digest: Computing boot digest: initramfs not found
```

**xfce-cachyos, niri-cachyos** — install succeeds, but the qcow2 boots to a painted screen and never emits the desktop contract marker; `scripts/iso-e2e.sh` times out after 900s:
```
ERROR: desktop experience contract marker was not emitted
```
(The marker comes from `tunaos-desktop-contract.service`, `WantedBy=graphical.target` — the guest never reached graphical.target. Serial log artifact for confirmation: [boot-gate-marlin-xfce-cachyos](https://github.com/tuna-os/tunaOS/actions/runs/31766586852/artifacts/9207968778).)

## Root cause

`build_scripts/overlay/cachyos.sh:41-43` installs `linux-cachyos` via `pacman -Syu` **without removing the stock `linux` kernel** (contrast `build_scripts/overlay/asahi.sh:179`, which does `pacman -Rdd --noconfirm linux-aarch64`). The full `-Syu` can additionally bump the stock kernel, so the image ends up with 2–3 kernel module trees.

Then `cachyos.sh:50` regenerates exactly **one** initramfs into:
```bash
"$(find /usr/lib/modules -maxdepth 1 -type d | grep -v '\.img' | tail -1)/initramfs.img"
```
`find` without `sort` returns readdir order — which module dir receives the initramfs is a per-image gamble. When bootc's BLS/composefs setup selects a kernel dir without `initramfs.img`, install fails ("initramfs not found"). When the gamble half-lands, the image installs but the booted kernel entry and module tree disagree, and boot stalls before graphical.target — matching the xfce/niri behavior.

[...]

PR #1580, opened 2026-08-14
fix(marlin): remove stock kernel in cachyos overlay, stop guessing initramfs target (#1563)

hanthor:
## Summary

All 5 `*-cachyos` flavors failed Gate on the first nightly with the new overlay matrix: gnome/cosmic/kde-cachyos died during install with `initramfs not found`; xfce/niri-cachyos installed but never reached `graphical.target`.

## Root cause

`cachyos.sh` installs `linux-cachyos` via `pacman -Syu` without removing the stock `linux` kernel that `Containerfile.arch`'s base stage already installed (the `[cachyos]` repo isn't registered in `pacman.conf` until this overlay runs, so the base build's own `KERNEL_PKG` detection always picks stock `linux` for this path). The image ends up with 2-3 kernel module trees, and the overlay then built exactly one initramfs into whichever directory `find /usr/lib/modules | tail -1` happened to return — readdir order is unspecified, so which tree got it was a per-image gamble.

## Changes

`build_scripts/overlay/cachyos.sh`:
1. Remove the stock kernel (`linux`, `linux-headers`) before installing `linux-cachyos`, mirroring the existing `asahi.sh` pattern for marlin's aarch64 stock kernel. Switched `pacman -Syu` to `-Sy` so a full system upgrade can't independently re-bump a kernel this step is trying to remove.
2. Build an initramfs for every remaining `/usr/lib/modules/*/` dir instead of guessing one via `find | tail -1`.

`Containerfile.arch`:
- Fixed the same `find | tail -1` idiom at its own dracut step. Currently harmless there (only one kernel exists at that build stage, per the issue), but fixed while in the area so a future change doesn't reintroduce this exact bug with different odds.

## Scope

Pure `build_scripts/` + `Containerfile.arch` change — no workflow-permission constraint (#1557).

Refs #1563
---
🐝 **Hive Agent**: `contributor` | **SHA:** `205c0af8`


hanthor:
Closing — the fix for #1563 is already on `main`, and this branch would take other things out with it.

`cca7549` ("fix(cachyos): keep one kernel and build its initramfs by name, not by readdir order") landed 08-14 15:41 UTC and does what this PR describes: removes the stock kernel and stops the `find | tail -1` guess. #1563 is closed as completed.

Merging this now would also revert five unrelated fixes, because the branch predates them and the diff is against that older base — `a641d20` (libpod_lock reset, #1556), `2a576f0` (syft timeout bound #1567, and the Rekor backoff #1560), `7863d46` (hardcoded maintainer email, #1649), `8e06c9b` (bootc-lifecycle jq precedence, #1430). Same pattern as #1579 and #1582, all three from the 08-14 06:3x batch off base `0c8af301`. `mergeable_state: dirty` is what has been holding it.

If any part of this branch is *not* in `cca7549` — the `Containerfile.arch` half of the `find | tail -1` cleanup looks like a candidate — it is worth a fresh PR off current `main` rather than a rebase of this one.

---

PR #1587, opened 2026-08-14
docs: connect installer contract to tacklebox roadmap

hanthor:
## Summary

- connect installer behavior contract issue #1197 to the Q4 Tacklebox decoupling roadmap row
- make the architectural exit criterion explicit: fisherman-driven frontends should consume one backend-owned contract/parity check instead of maintaining duplicated behavior

## Context

The behavior contract is already documented in `docs/INSTALLER-FRONTENDS.md`, and the GNOME scope correction is being handled separately in #1442. This change makes #1197 visible as the critical-path dependency for #1192 rather than leaving the architecture finding disconnected from the roadmap.

## Validation

- `git diff --check`
- `git diff --cached --check`

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -135,7 +135,7 @@ CI pipeline builds are green on amd64, amd64-v2, and arm64 for core variants.
 
 | Goal | Owner | Dependencies |
 |------|-------|--------------|
-| Tacklebox decoupling | architect | #1192 (tracker; #306 closed) |
+| Tacklebox decoupling | architect | #1192 (tracker; #306 closed), #1197 (installer behavior contract) — exit criterion: the four fisherman-driven frontends consume one backend-owned contract/parity check instead of maintaining five behavior copies |
 | Upstream snapshot automation | ci-maintainer | #1194 (tracker; #307 closed) |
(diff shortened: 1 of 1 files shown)


hanthor:
Closing: the ROADMAP row this PR edits was rewritten on `main`, and this one-line replacement would drop that rewrite.

The Tacklebox decoupling row now carries a 2026-08-14 audit result: 3 of #306's 4 recommendations already landed (`TACKLEBOX_SHA`/`TACKLEBOX_IMAGE` pinning via `scripts/lib/common.sh`; tacklebox running as a container image rather than a host binary; the `ghcr.io/hanthor/bluefin:lts` `iso.toml` reference gone), with the real remaining gap identified as the version pin having no single source of truth (`image-versions.yaml` vs hardcoded overrides in `publish-iso-groups.yml` and `luks-e2e.yml`).

This branch replaces that whole row with the pre-audit text plus a #1197 exit criterion, so applying it loses the audit.

The #1197 link is a reasonable thing to want in that row — but as an addition to the current text, and it is really a maintainer call whether "one backend-owned contract across the four fisherman frontends" is the exit criterion for Tacklebox decoupling or a separate goal. Worth raising on #1197 rather than re-running this branch.

---

PR #1595, opened 2026-08-14
ci: make ISO boot gate architecture-aware

### LINKED ISSUE #1593: [ci] scripts/iso-e2e.sh and ISO boot-gate QEMU setup are x86_64-only — arm64 ISO artifacts can't boot-verify
## Context

PR #1592 (fixing #1378) makes arm64 ISO/QCOW2 artifacts actually get scheduled and run on an arm64 runner. But two pieces of the pipeline remain x86_64-only:

- `scripts/iso-e2e.sh` looks specifically for `qemu-system-x86_64` (or `qemu-kvm`) and OVMF (x86 UEFI) firmware paths — no `qemu-system-aarch64` / AAVMF equivalent.
- `reusable-build-artifacts.yml`'s "Install QEMU for boot verification" step installs `qemu-system-x86 ovmf` unconditionally.

So once #1592 lands, arm64 artifact cells should get past ISO build (native arm64 runner, no cross-arch emulation needed there) but will fail at the boot-gate step.

## Proposed fix

- Install `qemu-system-arm`/AAVMF conditionally based on runner arch in `reusable-build-artifacts.yml`.
- Teach `scripts/iso-e2e.sh` to pick `qemu-system-aarch64` + AAVMF_CODE/AAVMF_VARS when running on aarch64, alongside the existing x86_64 path (probably a `machine`/`-cpu` selection difference too — `virt` machine type is typical for aarch64 QEMU, vs `q35`/`pc` for x86).

## Why filed separately

This needs verification against real aarch64 KVM boot behavior on an actual `ubuntu-24.04-arm` runner — not something safely guessable without CI access. Scoping it out of #1592 so that PR's mechanical, verified matrix/runner fix isn't blocked on unverified QEMU/firmware changes.

Refs #1378, #1592
---
🐝 **Hive Agent**: `contributor` | **SHA:** `5339b128`

hanthor:
## Summary

Fixes #1593 by making the ISO boot gate work on the native arm64 runners introduced by #1592.

- Install `qemu-system-arm` and `qemu-efi-aarch64` on ARM64 runners, retaining the existing x86 package path elsewhere.
- Select `qemu-system-aarch64` + AAVMF + the `virt` machine on aarch64; retain QEMU/KVM + OVMF + `pc` on x86_64.
- Use an ARM-compatible TCG CPU default and apply the selected machine/firmware to every live, installed-disk, TPM, and direct-disk boot path.
- Report the correct removable EFI fallback (`BOOTAA64.EFI` on ARM, `BOOTX64.EFI` on x86) in installed-ESP diagnostics.

## Root cause

The artifact matrix can now schedule arm64 ISO cells, but the reusable workflow installed only x86 QEMU/OVMF and `scripts/iso-e2e.sh` hardcoded the x86 emulator, firmware paths, and `pc` machine.

## Validation

- `bash -n scripts/iso-e2e.sh`
- `git diff --check`
- Confirmed Ubuntu Noble package metadata and AAVMF paths for `qemu-system-arm`/`qemu-efi-aarch64`.
- Local YAML parser and ShellCheck were unavailable; CI should provide the authoritative workflow/lint validation.

Real arm64 KVM/UEFI boot validation is intentionally left to the arm64 GitHub runner in CI, as called out by the issue.


hanthor:
I rebased this onto `main` (it was based on `a10ce08`, and its only red check had run during a window where CI's pytest step was broken repo-wide, so the result was unclassifiable). With a clean base the picture is now unambiguous.

## `Unit Tests` is a real failure, and it's a one-liner

pytest passes fully — **385 passed** — and every lint, CodeQL and E2E check is green. The job fails in the **BATS** step, on exactly one test:

```
not ok 62 installed boot: the ESP is dumped before the install VM is destroyed
# (in test file tests/bats/test_iso_e2e.bats, line 813)
#   `grep -q 'WARN: esp has NO EFI/BOOT/BOOTX64.EFI' "$SCRIPT"' failed
```

The assertion greps `scripts/iso-e2e.sh` for a hardcoded literal:

```bash
grep -q 'WARN: esp has NO EFI/BOOT/BOOTX64.EFI' "$SCRIPT"
```

This PR correctly parameterises that message:

```bash
echo "WARN: esp has NO EFI/BOOT/${fallback_efi}; firmware has no fallback to boot"
```

so the literal no longer exists. The test is pinning a string the change was right to make dynamic — the two sibling assertions on the same test (`ESP contents`, `esp: removable fallback present`) still pass because those strings are untouched. Updating line 813 to match the parameterised form should be all that's needed; asserting both concrete values would keep the test's intent while surviving the change:

```bash
grep -q 'WARN: esp has NO EFI/BOOT/' "$SCRIPT"
grep -q 'fallback_efi="BOOTX64.EFI"' "$SCRIPT"
grep -q 'fallback_efi="BOOTAA64.EFI"' "$SCRIPT"
```

## Two things I checked that are *not* problems

- The new `[[ ]]` sits in a block otherwise written in POSIX `test` syntax, which looked like a bashism. It isn't — that heredoc runs `sudo bash -s`. It also runs **inside the guest**, so `uname -m` reports the installed image's architecture rather than the runner's, which is more correct than a host-side check would be.
- `-machine virt` with `virtio-blk-pci` / `virtio-scsi-pci` is fine; arm `virt` has a PCIe bus.

## One risk worth a second opinion before merge

`scripts/iso-e2e.sh:378` sets `QEMU_GPU_ARGS=(-vga virtio -display none)` and this PR doesn't touch it. `-vga` is not reliably accepted on the aarch64 `virt` machine — it may need `-device virtio-gpu-pci` instead. I can't settle that without a real arm64 runner, which is the same limitation the PR description already calls out. Flagging it as the most likely next failure rather than as a blocker: without this PR arm64 ISO cells fail 100% at the x86 QEMU/OVMF step, so this is strictly an improvement either way.

---

PR #1605, opened 2026-08-14
fix(cachyos): build initramfs for deterministic kernel trees

### LINKED ISSUE #1563: [marlin] cachyos overlay keeps two kernels but builds one initramfs into a `find | tail -1` gamble — all 5 cachyos flavors fail Gate (bootc "initramfs not found" / boot never reaches graphical.target)

hanthor:
## Summary

Fixes #1563 by making CachyOS kernel/initramfs preparation deterministic.

- Install `linux-cachyos` without a mid-overlay full `pacman -Syu`.
- Remove the stock `linux` package when present, so bootc does not see competing kernel trees.
- Generate an initramfs for every remaining `/usr/lib/modules/<kver>` directory instead of selecting one through unstable `find | tail -1` ordering.
- Apply the same all-kernel initramfs rule to the Arch base Containerfile.
- Add focused Bats coverage for stock-kernel removal and the eliminated selection gamble.

## Root cause

The CachyOS overlay retained the Arch stock kernel while installing `linux-cachyos`, then generated one initramfs into whichever module directory happened to be returned last by unsorted `find`. Bootc/BLS could select a different kernel tree, producing either `initramfs not found` during installation or a guest that never reached the desktop contract.

## Validation

- `git diff --check`
- `bash -n build_scripts/overlay/cachyos.sh`
- Static checks confirm neither affected path contains the old `find | tail -1` selection.
- Focused Bats tests added; full Bats execution is delegated to CI.

### DIFF
--- a/Containerfile.arch
+++ b/Containerfile.arch
@@ -168,7 +168,11 @@ RUN --mount=type=bind,from=context,source=/,target=/run/context \
 # nowhere shared to live.
 RUN --mount=type=bind,from=context,source=/,target=/run/context \
     /run/context/build_scripts/bootc/dracut-config.sh && \
-    dracut --force "$(find /usr/lib/modules -maxdepth 1 -type d | grep -v '\.img' | tail -1)/initramfs.img"
+    for kernel_dir in /usr/lib/modules/*; do \
+        [ -d "${kernel_dir}" ] || continue; \
+        case "${kernel_dir##*/}" in *.img) continue ;; esac; \
+        dracut --force "${kernel_dir}/initramfs.img"; \
+    done
 
 # systemd-firstboot asks the CONSOLE for a timezone and blocks there forever.
 #
--- a/build_scripts/overlay/cachyos.sh
+++ b/build_scripts/overlay/cachyos.sh
@@ -38,14 +38,38 @@ Include = /etc/pacman.d/cachyos-mirrorlist
 EOF
 fi
 
-pacman -Syu --noconfirm --needed \
+# Install the CachyOS kernel without upgrading the entire base mid-overlay.
+# A full -Syu can update the stock kernel as a side effect, leaving multiple
+# module trees for the bootc/BLS generator to choose between.
+pacman -S --noconfirm --needed \
 	cachyos-keyring cachyos-mirrorlist cachyos-settings \
 	linux-cachyos linux-cachyos-headers tpm2-tss
+
+# The Arch base's stock linux package is not a usable fallback here: keeping
+# it alongside linux-cachyos gives bootc two kernels but only one reliably
+# prepared initramfs. Remove it before generating the CachyOS initramfs.
+if pacman -Qq linux >/dev/null 2>&1; then
+	pacman -Rdd --noconfirm linux
+fi
 
 # Mark as CachyOS-augmented for install-desktop.sh detection
 install -D /dev/null /etc/cachyos-release
 printf 'CachyOS\n' >/etc/cachyos-release
 
-# Rebuild initramfs via dracut
-dracut --force --omit "tpm2-tss" "$(find /usr/lib/modules -maxdepth 1 -type d | grep -v '\.img' | tail -1)/initramfs.img"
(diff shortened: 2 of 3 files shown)


hanthor:
Closing as superseded — the fix for #1563 already landed on `main`, in a stronger form than this draft.

`build_scripts/overlay/cachyos.sh` on `main` now:

- selects the kernel by **name** (`-name '*cachyos*'`), not by `find | tail -1` or by version sort;
- hard-fails when that selection is empty, printing the trees it did find;
- deletes every leftover module tree, because `initramfs.img` is generated rather than packaged, and then asserts exactly one remains **before** building the initramfs;
- calls `dracut --force --omit tpm2-tss "/usr/lib/modules/${KVER}/initramfs.img" "${KVER}"` with both the path and the kernel version.

`Containerfile.arch` got the same treatment, and `tests/bats/test_cachyos_overlay.bats` pins all of it (`no find|tail -1 feeds an initramfs path`, `the cachyos kernel is selected by name, not by version sort`, `exactly one remaining kernel is proven, not hoped for`, `the single-kernel assertion runs BEFORE the initramfs is built`).

This branch's approach — build an initramfs for *every* remaining tree — is the alternative that `main` deliberately did not take, and it also conflicts in both files. Nothing here is still missing; reopen if you disagree.

---

PR #1613, opened 2026-08-14
ci: gate installer smoke on GPU capability

### LINKED ISSUE #1396: [ci-maintainer] Installer Smoke now weekly-scheduled and red across board: installer runs but never reports a window (EGL dri2 failure, 22+ jobs)

hanthor:
## Summary

- detect whether the runner exposes the DRM render node required by EGL-dependent desktop frontends
- keep GNOME coverage in automatic hosted runs while filtering deterministic GPU-dependent failures from the generated matrix
- preserve explicit manual flavor selections for targeted diagnosis
- add a regression assertion for the capability gate

## Root cause

The installer smoke harness falls back to plain `-vga virtio` when `/dev/dri/renderD128` is unavailable. KDE, COSMIC, Niri, and XFCE require `EGL_EXT_device_drm`; they can remain alive without producing a readiness stamp, yielding the issue's misleading failure.

## Validation

- `git diff --check`
- matrix filter smoke test with `jq`
- focused installer-smoke assertions
- YAML parse was not run because PyYAML is unavailable in the checkout environment

Closes #1396

### DIFF
--- a/.github/workflows/installer-smoke.yml
+++ b/.github/workflows/installer-smoke.yml
@@ -38,7 +38,29 @@ permissions:
   packages: read
 
 jobs:
+  # Hosted Ubuntu runners do not expose a DRM render node.  The Smithay
+  # frontends (KDE, COSMIC, Niri and XFCE) need EGL_EXT_device_drm, so
+  # launching those legs on plain virtio only produces the misleading
+  # "process is alive, no readiness stamp" failure.  Detect the capability
+  # before expanding the matrix and keep the weekly hosted run honest.
+  host-capabilities:
+    runs-on: ubuntu-latest
+    outputs:
+      gpu_available: ${{ steps.detect.outputs.gpu_available }}
+    steps:
+      - id: detect
+        run: |
+          if [[ -e /dev/dri/renderD128 ]]; then
+            echo "gpu_available=true" >> "$GITHUB_OUTPUT"
+            echo "DRM render node detected; GPU-dependent installer legs enabled"
+          else
+            echo "gpu_available=false" >> "$GITHUB_OUTPUT"
+            echo "::notice::No DRM render node; GPU-dependent installer legs will be omitted"
+            echo "Explicit manual flavor selections remain available for targeted diagnosis."
+          fi
+
   generate-matrix:
+    needs: host-capabilities
     runs-on: ubuntu-latest
     outputs:
       matrix: ${{ steps.gen.outputs.matrix }}
@@ -56,6 +78,7 @@ jobs:
         env:
           FILTER_VARIANT: ${{ inputs.variant || '' }}
           FILTER_FLAVORS: ${{ inputs.flavors || '' }}
+          GPU_AVAILABLE: ${{ needs.host-capabilities.outputs.gpu_available }}
         run: |
           set -euo pipefail
           json="$(yq -o=json '.' .github/build-config.yml)"
@@ -71,6 +94,16 @@ jobs:
                  | select(.build_iso == true)
                  | select((.platforms // $variant.platforms // []) | index("linux/amd64"))
                  | {variant: $v, flavor: .id} ]}')
+          fi
+
+          # A plain hosted runner cannot provide the EGL/DRM surface required
+          # by the four non-GNOME frontends.  Keep GNOME as a useful weekly
+          # smoke signal, and reserve the other legs for a GPU-capable manual
+          # run instead of reporting a deterministic infrastructure failure
+          # as an installer regression.
+          if [[ "$GPU_AVAILABLE" != "true" && -z "$FILTER_FLAVORS" ]]; then
+            echo "::notice::Filtering GPU-dependent installer flavors (no DRM render node)"
+            M=$(echo "$M" | jq -c '{include: [.include[] | select(.flavor == "gnome")]}' )
           fi
           echo "matrix=$M" >> "$GITHUB_OUTPUT"
           echo "$M" | jq .
--- a/tests/bats/test_live_installer_launch.bats
+++ b/tests/bats/test_live_installer_launch.bats
@@ -95,6 +95,14 @@ launcher_app() {
   grep -Fq 'grep -Fx' "$WORKFLOW"
 }
 
+@test "installer smoke gates GPU-dependent flavors on runner capability" {
+  [ -f "$WORKFLOW" ]
+  grep -Fq 'host-capabilities:' "$WORKFLOW"
+  grep -Fq 'gpu_available' "$WORKFLOW"
+  grep -Fq 'No DRM render node' "$WORKFLOW"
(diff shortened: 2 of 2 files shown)


hanthor:
I wrote the readiness-stamp assertion this PR's root cause refers to (#1173), so flagging two things from that side rather than opening anything competing.

**The capability probe is half of the condition `iso-e2e.sh` actually uses.**

`host-capabilities` tests `-e /dev/dri/renderD128`. The harness gate is two-part (`scripts/iso-e2e.sh:404-405`):

```bash
if [[ "$_gpu_mode" != "plain" ]] && { [[ "$_gpu_mode" == "virgl" ]] || [[ -e /dev/dri/renderD128 ]]; } &&
	"$QEMU" -device help 2>/dev/null | grep -q "virtio-vga-gl"; then
```

A render node **and** a QEMU build that has `virtio-vga-gl`. On a runner with the node but a QEMU lacking the device, this PR would keep the four legs in the matrix while `iso-e2e.sh` silently falls back to `-vga virtio` — reproducing exactly the "alive, no readiness stamp" failure the gate exists to prevent, just less often and harder to spot.

Worth mirroring the second clause, or better, having both read one helper so they cannot drift. Two independent definitions of "this runner can do GL" is the kind of thing that agrees on day one and stops agreeing later.

Also note `_gpu_mode` (`TBOX_E2E_GPU=virgl`) can force virgl *without* a render node, so an operator override is invisible to a probe that only stats the device.

**Second, a consequence worth making explicit rather than discovering later.**

With the filter active, the weekly hosted run becomes GNOME-only — which means the four non-GNOME frontends' readiness stamps are never exercised in CI at all. Those stamps just landed across all five frontends specifically so a windowless frontend fails on a GPU-less runner; filtering those legs out means that assertion can no longer fire for them, and "cannot fire" reads as green forever. `docs/INSTALLER-FRONTENDS.md` makes the same point about the old `pgrep -af` check.

That may well be the right trade — a deterministic infra failure *is* noise, and the `::notice::` keeps it visible in the log. But the stamp check is cheap and GPU-independent by design: it is readable over SSH with no compositor rendering at all. So there's a middle option worth considering — keep the four legs, and gate only the **pixel** checks (blank/advance/OCR) on `gpu_available`, letting the stamp assertion still run everywhere. That preserves real coverage on hosted runners for the failure mode that motivated the stamp, while still not reporting missing EGL as an installer regression.

One process note: this PR's base predates the `generate-matrix` rewrite now on `main` (the `yq`/`build-config.yml` form), and `main` has moved a long way since — worth a rebase before it leaves draft.

---

hanthor:
Closing as obsolete — `main` resolved #1396 in the opposite direction, deliberately.

`installer-smoke.yml` was split by #1929 into `build-iso` (builds and uploads the ISO) and `smoke` (downloads it, boots it, asserts). The DRM render node belongs to the second half, which runs on `ubuntu-latest` and drives `scripts/iso-e2e.sh`, not `iso-e2e-gpu.sh`. `main` then removed the `gpu` routing entirely, and `tests/test_installer_smoke_needs_no_gpu.py` now pins that outcome:

- `test_no_cell_is_routed_to_the_gpu_group` — `'"gpu"'` must not appear in the workflow;
- `test_the_boot_half_runs_where_the_render_node_question_lives` — `smoke` must stay on `ubuntu-latest`;
- `test_the_boot_half_uses_the_non_gpu_harness`.

This PR's `host-capabilities` job probes `/dev/dri/renderD128` on the matrix-generation runner (`ubuntu-latest`), where it is never present, so in practice it would permanently narrow the smoke matrix to `gnome` — the opposite of what `main` now asserts is unnecessary. Its own **Unit Tests** check is also red.

Reopen if the EGL readiness question comes back, but it needs rewriting against the post-#1929 workflow shape.

---

Issue #1570, opened 2026-08-14
[tracking] Fully green build matrix — consolidated root causes from the 2026-08-14 nightly sweep (all 13 variants diagnosed)


hanthor:
## What this is

Every variant nightly failed on 2026-08-14 (and most since ~08-08). All 13 variant workflows were diagnosed job-by-job from the latest runs (main @ `fed4749f`): yellowfin, albacore, skipjack, bonito, bonito-rawhide, sailfin, guppy, grouper, marlin, flounder, flounder-sid, gurnard, hummingbird. Every failed job maps to one of the root causes below — there are **no unexplained failures**. This issue tracks the set; each item is independently fixable.

## Deterministic bugs (red every night until fixed)

| Issue | What | Costs |
| :--- | :--- | :--- |
| #1561 | `verify-nvidia.sh` initramfs parity false-FAIL on the fc43 7.1.4 kernel | ~9 nvidia cells on bonito, rawhide, albacore, skipjack, yellowfin — **largest single win** |
| #1563 | marlin cachyos overlay: dual kernels, one `find\|tail -1` initramfs | all 5 cachyos cells |
| #1564 | flounder trixie nvidia: `libnvidia-allocator1` missing → no `nvidia-drm_gbm.so` | 3 flounder nvidia cells |
| #1565 | flounder-sid nvidia: headers/kernel skew (cached base 7.1.7 vs sid archive 7.1.8) | 3 sid nvidia cells |
| #1566 | rawhide base: openh264 fc45 RPM vs F46 key (upstream branch window) | whole rawhide variant |
| #1569 | EL10 asahi verify gate structurally unpassable (4 known #777 gaps as hard FAILs) | albacore+skipjack gnome-asahi |
| #1555 | hummingbird: repo dep breakage (libjsoncpp/librhash providers, KDE/Qt6 devel) | whole variant *(pre-existing issue)* |

## Infra / flake amplifiers (kill different cells each night)

| Issue | What | Costs |
| :--- | :--- | :--- |
| #1560 | Rekor 502 during `cosign attest`; ~30s retry vs 45+ min outage; blocks Promote for green builds | 25 Sign+attest jobs on 08-14 alone (7 variants) |
| #1556 | arm64 `libpod_lock` podman failure at sudo login; in-place retry (#1511) proven useless | ~12 arm64 cells/night across 8 variants *(pre-existing issue — consolidated evidence + fix proposal added there)* |
[...]

PR #1644, opened 2026-08-14
fix(flounder): install the packages that actually ship the Wayland GBM path

### LINKED ISSUE #1564: [flounder] trixie nvidia overlay omits libnvidia-allocator1 → nvidia-drm_gbm.so missing → Debian contract check deterministically fails all nvidia flavors

hanthor:
Closes #1564.

## Confirmed from the archive, not inferred

The contract demands `/usr/lib/x86_64-linux-gnu/nvidia/current/nvidia-drm_gbm.so` and no package in the install list ships it. It comes from `libnvidia-allocator1`, which `nvidia-driver-libs` only **Recommends** — and `--no-install-recommends` drops it:

```
libnvidia-allocator1 (trixie 550.163.01-2) file list:
  /usr/lib/x86_64-linux-gnu/nvidia/current/libnvidia-allocator.so.1
  /usr/lib/x86_64-linux-gnu/nvidia/current/nvidia-drm_gbm.so

nvidia-driver-libs:  rec: libnvidia-allocator1 (= 550.163.01-2)
```

`--no-install-recommends` is deliberate and stays, so the Recommends has to be named.

## Why this isn't a one-line list edit

`nvidia-drm_gbm.so` is loaded **through** an EGL external-platform registration, and that registration is in a different package entirely:

```
libnvidia-egl-gbm1:
  /usr/lib/x86_64-linux-gnu/libnvidia-egl-gbm.so.1
  /usr/share/egl/egl_external_platform.d/15_nvidia_gbm.json
```

Nothing in the nvidia chain depends on **or** recommends it — I checked `nvidia-driver` and `nvidia-driver-libs` directly — so it was never going to arrive on its own, even with Recommends enabled.

That matters: adding `libnvidia-allocator1` alone would have turned the check green while EGL still never looked for the GBM platform. The contract would have certified *"Wayland compositors can use the driver"* in a state where they could not. So the check is extended to assert `15_nvidia_gbm.json` alongside the `.so` — a check that guards half a mechanism is how this stayed scoped to one filename for as long as it did.

`libnvidia-egl-wayland1` is the same component the Arch sibling already installs as `egl-wayland` (`overrides/nvidia-arch/20-nvidia.sh`), so this also brings the two families level. It's worth noting it does **not** have the same split problem — it ships its own registration, so no fourth package is hiding behind it:

```
libnvidia-egl-wayland1 (trixie 1:1.1.18-1) file list:
  /usr/lib/x86_64-linux-gnu/libnvidia-egl-wayland.so.1
  /usr/share/egl/egl_external_platform.d/10_nvidia_wayland.json
```

The allocator/egl-gbm split is the unusual one, which is exactly why it went unnoticed.

## Suite safety

All three are in **stable and unstable**, amd64 and arm64 (`madison`, 2026-08-14), so this is safe for flounder (trixie) and flounder-sid alike. `libnvidia-egl-gbm1` is in contrib, the rest non-free/main — all enabled by the component block already in this script.

These are Debian package names, and this overlay dispatches on `IS_DEBIAN`, so I checked whether Ubuntu-based Grouper/Gurnard could hit them: `.github/build-config.yml` declares no `grouper`/`gurnard` nvidia flavor, so nothing else reaches this code path.

## Verification

**30/30 on this branch, 26/30 against `upstream/main`**, with exactly the 4 new tests failing there. The good-root fixture gains the platform JSON, which is why the pre-existing "passes on a complete driver install" test still passes — the fixture and the contract moved together.

`bats` isn't installed in my environment, so I ran the real `.bats` file through a shim; CI is the real runner.

**Boundary:** verified against the live Debian archive and fixture-tested — not built. First real proof is the next flounder nightly.

## Note for whoever merges

This touches the same `apt-get install` block as **#1642** (flounder-sid header pinning, issue #1565). Both are based on `main`, so they'll conflict textually. The resolution for that block is to keep #1642's split (exact-ABI `linux-headers-${KVER}` in its own transaction first) and add this PR's three packages to the driver transaction.

They also both touch `tests/bats/test_nvidia_overlay_arch_debian.bats` — #1642 adds 5 tests and rewrites `guards on IS_DEBIAN`; this one adds 4 and edits `make_good_debian_root`. Different regions, but expect a second conflict there, not just the one in the install list. They're independent fixes for different variants and neither depends on the other.

### DIFF
--- a/build_scripts/checks/verify-nvidia-debian.sh
+++ b/build_scripts/checks/verify-nvidia-debian.sh
@@ -148,6 +148,19 @@ if compgen -G "${NVLIB}/nvidia-drm_gbm.so" >/dev/null; then
 else
 	fail "missing /usr/lib/x86_64-linux-gnu/nvidia/current/nvidia-drm_gbm.so — Wayland compositors cannot use the driver"
 fi
+# The backend above is loaded THROUGH the EGL external-platform registration,
+# not directly: without 15_nvidia_gbm.json, EGL never looks for the GBM
+# platform and nvidia-drm_gbm.so sits on disk doing nothing. Checking only the
+# .so made "Wayland compositors can use the driver" provable in a state where
+# they still could not — and, unlike the .so, nothing in the nvidia-driver
+# dependency chain pulls this in even with Recommends enabled, so it can only
+# ever arrive by being named in the install list (tunaOS#1564). Path read from
+# libnvidia-egl-gbm1's file list (trixie 1.1.2.1-1; same version in unstable).
+if compgen -G "${NV_ROOT}/usr/share/egl/egl_external_platform.d/15_nvidia_gbm.json" >/dev/null; then
+	pass "EGL GBM external platform registered"
+else
+	fail "missing /usr/share/egl/egl_external_platform.d/15_nvidia_gbm.json (libnvidia-egl-gbm1) — EGL will not load the GBM backend"
+fi
 
 echo "== initramfs policy =="
 DRACUT_CONF="${NV_ROOT}/usr/lib/dracut/dracut.conf.d/99-nvidia.conf"
--- a/build_scripts/overlay/overrides/nvidia-debian/20-nvidia.sh
+++ b/build_scripts/overlay/overrides/nvidia-debian/20-nvidia.sh
@@ -87,6 +87,40 @@ if ! apt-cache show "$HEADERS_PKG" >/dev/null 2>&1; then
 	HEADERS_PKG="linux-headers-generic"
 fi
 
+# linux-headers-generic is a virtual package resolving to the arch-specific
+# real one (linux-headers-amd64 here) — see header. dkms needs it present
+# under /usr/lib/modules/${KVER}/build before autoinstall below has
+# anything to build against.
+#
+# --no-install-recommends is deliberate, but it means every Recommends this
+# stack needs has to be named here. nvidia-driver-libs Recommends
+# libnvidia-allocator1 (verified on packages.debian.org, trixie 550.163.01-2),
+# and that is the package shipping
+# /usr/lib/x86_64-linux-gnu/nvidia/current/nvidia-drm_gbm.so — read straight
+# off its file list, not inferred. Dropping the Recommends dropped the file,
+# and verify-nvidia-debian.sh failed every flounder nvidia build on it
+# (tunaOS#1564).
+#
+# The two egl-* packages are separate source packages that nothing in the
+# nvidia-driver chain depends on OR recommends, so they were never going to
+# arrive on their own. They are what makes the GBM backend reachable rather
+# than merely present:
+#
+#   libnvidia-egl-gbm1     libnvidia-egl-gbm.so.1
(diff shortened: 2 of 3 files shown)


hanthor:
Rebased on `main` — conflict resolved and pushed.

**The conflict was complementary, not competing.** `main` had gained the sid/rolling headers-version fallback (`HEADERS_PKG` + `apt-cache show` probe, from #1565); this branch adds the missing GBM packages. I kept **both**, and keeping main's block is not optional — the install list on this branch already references `"${HEADERS_PKG}"`, so dropping it would have left an unbound variable under `set -u`.

Order is main's code first, then this branch's comment block immediately above the `apt-get install`.

**Tests: 30/30 pass** on `tests/bats/test_nvidia_overlay_arch_debian.bats`, including the three this PR adds:

```
ok 26 nvidia-debian 20-nvidia.sh installs libnvidia-allocator1 for nvidia-drm_gbm.so
ok 27 nvidia-debian 20-nvidia.sh installs the EGL external platform packages
ok 28 the contract check and the install list agree about the GBM path
```

Worth noting #1564 is now the **last open deterministic per-variant bug** on the #1570 tracker — everything else there is either closed with a merged fix or is one of the three CI-structure issues (#1729/#1730/#1731). So this is the one that still costs flavor cells every night.

Also: **#1579 targets the same issue and is superseded by this PR** — I verified that rather than assuming it, and left the evidence there.

---

PR #1645, opened 2026-08-14
fix(chunkah): use the digest pin that already existed, prune /sysroot/, validate before load

### LINKED ISSUE #1568: [bonito] chunkah rechunk emitted a corrupt oci-archive ("invalid tar header") and CHUNKAH_IMAGE is unpinned :latest — pin it, prune /sysroot, validate before podman load
## Symptom

bonito `base / linux-amd64` (2026-08-14 run [31766548793](https://github.com/tuna-os/tunaOS/actions/runs/31766548793), job 94663563425) failed at the rechunk→load handoff, exit 125:

```
INFO build complete            (chunkah)
...
Error: payload does not match any of the supported image formats:
 * oci-archive: ... adding layer with blob "sha256:85439f79..." :
   unpacking failed (error: exit status 1; output: archive/tar: invalid tar header)
exhausting input failed (error: archive/tar: invalid tar header)
+ LOADED_ID=
error: recipe `_build` failed with exit code 125
```

chunkah reported `build complete`, but the produced `out.ociarchive` was corrupt/truncated at a mid-archive blob. This killed bonito's base → Manifest → everything downstream (together with the arm64 lock flake on the other leg).

## Analysis

Single occurrence so far — gnome-hwe amd64 rechunked and loaded fine in the same run — so treat as environment/upstream drift until reproduced. Two candidate causes, both worth eliminating regardless:

1. **`CHUNKAH_IMAGE` is unpinned** at `quay.io/coreos/chunkah:latest` (`scripts/build-image-inner.sh:195-197`) — an upstream regression lands in our builds the hour it's published, unreviewed. Every other build input in this repo is digest-pinned (see `image-versions.yaml`'s whole design); chunkah should be too.
2. **Disk pressure**: the uncompressed ~5.4 GiB archive is written before the `podman system prune -af` that freed 94 GB later in the job — a truncated write during low-disk is consistent with a mid-blob invalid tar header.

Also observed in the same log, independent of the failure: `rootfs contains sysroot/ostree which was not pruned ... Use --prune /sysroot/` — wasted bytes in every rechunked image.

## Suggested fix

In `scripts/build-image-inner.sh:193-252`:
1. Pin `CHUNKAH_IMAGE` to a digest (and let Renovate bump it, like every other pin).
2. Add `--prune /sysroot` to the chunkah invocation for bootc images.
3. Validate the archive before `podman load` (`tar tf` or a size sanity check) and/or check free disk before the rechunk, retrying once on corruption so a truncated write doesn't cost the whole variant night.

Pure `scripts/` change — no workflow-permission constraint (#1557).

## Determinism

Unclear (1 occurrence on 08-14); the hardening is warranted either way.

hanthor:
Closes #1568.

The corruption itself is one occurrence and unreproduced. This eliminates both candidate causes and — more usefully — makes the next occurrence say what happened instead of surfacing as an opaque tar error two steps downstream.

## The pin already existed, and had never worked

The issue asks for `CHUNKAH_IMAGE` to be digest-pinned. `registry-map.yaml` has carried a `coreos-chunkah` digest all along, under the comment *"SECURITY: Pinned to digest to prevent supply-chain attacks via tag mutation"*. **No caller could use it:**

```bash
local digest_override_var="TUNA_IMAGE_DIGEST_${name}"
```

A hyphen isn't a legal shell identifier character, so for `coreos-chunkah` this didn't merely miss an override — bash raised `invalid variable name`, and under `_registry.sh`'s own `set -euo pipefail` that took `registry_ref` down with it. Not a wrong answer: **no answer**.

```
$ bash -c 'v="TUNA_IMAGE_DIGEST_coreos-chunkah"; echo "${!v:-FALLBACK}"'
bash: TUNA_IMAGE_DIGEST_coreos-chunkah: invalid variable name   (rc=1)
```

The path and tag overrides two lines above already do `${name//-/_}`, and the file header documents `TUNA_IMAGE_DIGEST_coreos_chunkah` with underscores. The tests missed it because the digest fixture had no hyphen and the hyphen fixture had no digest — neither covered the intersection, which is exactly what `coreos-chunkah` is.

So `build-image-inner.sh` hardcoded `:latest`, and an upstream chunkah regression reached our builds the hour it was published.

**The pinned digest was also gone.** Quay returns 404 for `sha256:33ab77f2…` — garbage-collected upstream while nothing bumped it. Refreshed to `sha256:ff8b8b46…` (verified: manifest list covering amd64 + arm64), and added a `renovate.json` manager for `registry-map.yaml`, scoped to `registry: quay` entries so `registryUrlTemplate` is correct by construction. A pin nothing maintains is a slow-motion build break — this one already broke.

## `--prune /sysroot/` — the trailing slash is not cosmetic

The issue suggests `--prune /sysroot`. From chunkah's scanner:

> Paths must be absolute. A trailing `/` means prune children only, keeping the directory itself. — `src/scan.rs`

Without the slash, chunkah drops `/sysroot` **itself** — the mountpoint a bootc image needs. Upstream's own warning, the one bonito logged, spells it `--prune /sysroot/`.

Applied to **both** branches (`prune` is a `Vec`, so it's repeatable). The rpmdb branch — every RPM variant, bonito included — passed no `--prune` at all, so the ostree object store has been packed into every rechunked RPM image.

## Validation before `podman load`

A zero exit from chunkah is not evidence the archive is loadable; that's the whole lesson here. `validate_ociarchive` walks every tar header (the read that failed) and confirms an `oci-layout` entry; the rechunk retries once on a fresh output directory; a corrupt archive fails *here*. Plus `df` around the rechunk, so the disk-pressure hypothesis gets confirmed or killed next time rather than re-argued.

## Two bugs found by running the validator, not reviewing it

I built real tarballs and ran the actual function against them. Both of these would have broken every build:

| Bug | Effect |
|---|---|
| `tar tf … \| grep -q` | `grep -q` exits at first match → tar takes SIGPIPE → under `set -o pipefail` the pipeline fails → **a valid archive is rejected** |
| `^\./?oci-layout$` | makes the *slash* optional, not the `./` → **rejects archives written without the `./` prefix** |

Final behavior, against real files:

```
valid.ociarchive      (./ prefix)   VALID
valid2.ociarchive     (bare)        VALID
truncated.ociarchive                REJECTED
empty.ociarchive                    REJECTED
notoci.tar                          REJECTED
missing.ociarchive                  REJECTED
script survived set -euxo pipefail
```

I also had a third "bug" in a draft comment claiming `[[ -n "$X" ]] && rm` would abort under `set -e`. It doesn't — errexit exempts it when another statement follows. I kept the `if` form (it's only safe by line order) but corrected the comment rather than shipping a confident wrong explanation.

## Verification

| Suite | This branch | `upstream/main` |
|---|---|---|
| `test_chunkah_rechunk.bats` (new, 12) | 12/12 | 4/12 |
| `test_registry_ref.bats` (+3) | 21/21 | 18/21 |
| `test_registry.bats` (untouched) | 15/15 | 15/15 |

`bats` isn't installed here, so I ran the real `.bats` files through a shim; CI is the real runner.

## One observation, deliberately not fixed

`.github/workflows/validate-renovate.yaml` watches `.github/renovate.json5`, which doesn't exist — the live config is `renovate.json` at the repo root. So renovate config changes, including this one, get **no CI validation**. I mitigated by modelling the new manager on the existing `image-versions.yaml` one, checking the JSON parses, and running the regex against the real file (1 match, correct captures). Fixing the workflow path is a separate change and a workflow file.

## Two things a reviewer will reasonably ask

**Does `registry_ref` add a new dependency to the build path?** No. `build-image-inner.sh` already sources `_registry.sh` at line 40, whose export block calls `registry_ref common` at source time — so `yq` has always been a hard requirement of this script, and the build job installs it (`reusable-build-image.yml`, "Install just and yq", verified with `yq --version`). Unlike the existing `registry_ref akmods` call, mine is unprotected, which is correct: a chunkah ref that can't be resolved should stop the build rather than silently fall back to an unpinned tag.

**Does CI green here prove the new call works?** No — and this is worth being blunt about. The rechunk path only runs when `SKIP_RECHUNK != 1`, and PR builds skip it. Lint and Unit Tests passing on this PR tell you nothing about the `registry_ref coreos-chunkah` line itself; only a real variant build exercises it. That's what the resolver tests are for — including one that resolves the actual shipped `registry-map.yaml` rather than a fixture.

**Boundary:** the `registry_ref` fix and the validator are behaviorally tested; the chunkah invocation itself is shape-tested and read from upstream source — not built. First real proof is the next bonito nightly.

### DIFF
--- a/registry-map.yaml
+++ b/registry-map.yaml
@@ -68,7 +68,16 @@ images:
     path: coreos/chunkah
     # SECURITY: Pinned to digest to prevent supply-chain attacks via tag mutation.
     # tag: latest is retained as fallback for setups without the pinned image.
-    digest: sha256:33ab77f29c90fb3ec7a3a6b6c84ee2f2bcff1846b3c34a3231092aebd39e673b
+    #
+    # Refreshed 2026-08-14 (tunaOS#1568). The previous pin
+    # (sha256:33ab77f2…) had been garbage-collected upstream — quay returns
+    # 404 for it — so anything resolving this entry would have fallen through
+    # to building chunkah from source on every build. Nothing noticed because
+    # registry_ref could not read this entry at all (see scripts/_registry.sh)
+    # and the only consumer hardcoded :latest. A pin that nothing bumps and
+    # nothing reads rots; renovate.json now has a manager for this file.
+    # Verified a manifest list covering amd64 + arm64, the two arches built here.
+    digest: sha256:ff8b8b466a942ec6000445d4001fc661e2fc5a952ad9ee29b4de9ab09d1d1708
     tag: latest
 
   # --- Tool images (build infra) ---
--- a/renovate.json
+++ b/renovate.json
@@ -1,23 +1,23 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-  "extends": ["config:best-practices"],
-
+  "extends": [
+    "config:best-practices"
+  ],
   "platformAutomerge": true,
   "automergeType": "pr",
   "automergeStrategy": "squash",
   "automerge": true,
-
   "rebaseWhen": "conflicted",
-
   "git-submodules": {
     "enabled": true
   },
-
   "customManagers": [
     {
       "customType": "regex",
       "description": "Update image digests in image-versions.yaml",
-      "managerFilePatterns": ["/^image-versions\\.yaml$/"],
+      "managerFilePatterns": [
+        "/^image-versions\\.yaml$/"
+      ],
       "matchStrings": [
         "- name: (?<depName>[\\w-]+)\\n\\s+image: (?<packageName>[^\\n]+)\\n\\s+tag: (?<currentValue>[^\\n]+)\\n\\s+digest: (?<currentDigest>sha256:[a-f0-9]+)"
       ],
@@ -27,7 +27,9 @@
     {
       "customType": "regex",
       "description": "Update pinned downloads in image-versions.yaml (GitHub releases)",
-      "managerFilePatterns": ["/^image-versions\\.yaml$/"],
+      "managerFilePatterns": [
+        "/^image-versions\\.yaml$/"
+      ],
       "matchStrings": [
         "#\\s*renovate:\\s*datasource=(?<datasource>[^\\s]+)\\s+depName=(?<depName>[^\\s]+)\\n\\s+[\\w-]+:\\s*\"(?<currentValue>[^\"]+)\""
       ]
@@ -35,7 +37,9 @@
     {
       "customType": "regex",
       "description": "Update GitHub Actions pinned SHAs",
-      "managerFilePatterns": ["/\\.github/workflows/.+\\.ya?ml$/"],
+      "managerFilePatterns": [
+        "/\\.github/workflows/.+\\.ya?ml$/"
+      ],
       "matchStrings": [
         "uses:\\s+(?<depName>[\\w-]+/[\\w-]+)@(?<currentDigest>[a-f0-9]{40})\\s*#\\s*(?<currentValue>v?\\d[^\\s]*)"
       ],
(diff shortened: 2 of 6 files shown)


hanthor:
Reviewed this against the tree rather than the description — **the load-bearing claims all hold up**:

| Claim | Check | Result |
|---|---|---|
| Hyphen breaks `${!var}` | `bash -c 'v="TUNA_IMAGE_DIGEST_coreos-chunkah"; echo "${!v:-X}"'` | ✅ `invalid variable name`, rc=1 |
| …and it's fatal | `scripts/_registry.sh:23` | ✅ `set -euo pipefail` present |
| Digest path missed the sanitize | `_registry.sh:97` vs `:102` | ✅ `${name}` vs `${name//-/_}` — exactly the asymmetry |
| Old pin was GC'd | `quay.io/v2/coreos/chunkah/manifests/sha256:33ab77f2…` | ✅ **404** |
| New pin is valid | `…/sha256:ff8b8b46…` | ✅ **200**, OCI index, `linux/amd64` + `linux/arm64` |

The "a pin that nothing bumps and nothing reads rots" framing is the right diagnosis, and the trailing-slash note on `--prune /sysroot/` is correct and worth keeping.

Two things to fix or answer:

### 1. `coreos-chunkah` is not the only hyphenated entry with a digest

The new comment in `_registry.sh` says:

> coreos-chunkah is the only hyphenated image in registry-map.yaml that carries a digest

`bluefin-iso` (registry-map.yaml:93) also has both — hyphen in the name, `digest: sha256:fa7d24b6…`, `tag: lts`. So `registry_ref bluefin-iso` died the same way.

The **fix is unaffected** — `${name//-/_}` covers both. But the comment is the account a future maintainer will trust, and it currently understates the blast radius to one entry when it was two.

Worth considering a scope question this raises: the Renovate manager this PR adds is scoped to `registry: quay` entries by construction (deliberately, per its own description). `bluefin-iso` is `registry: ghcr`, so it stays in exactly the unbumped state that let the chunkah pin rot. I tried to check whether that digest is also already GC'd and **could not get a conclusive answer** — my anonymous ghcr token request failed and the authenticated attempt 404'd on the `lts` control tag too, so the path/auth was wrong rather than the image missing. Flagging it as unverified, not as a finding.

### 2. Dropping the fallback is a behavior change — intended?

```bash
CHUNKAH_IMAGE="$(registry_ref coreos-chunkah)"
```

Under `build-image-inner.sh`'s `set -euxo pipefail`, a failed command substitution now kills the build. Every other `registry_ref` caller in the repo falls back instead — including **line 77 of this same file**:

```bash
AKMODS_REGISTRY_BASE="$(registry_ref akmods 2>/dev/null || echo "ghcr.io/${AKMODS_ORG}")"
```

…plus `scripts/rechunk.sh:21` and `scripts/resolve-image.sh:54`.

Failing loud rather than silently unpinning is a defensible call — arguably the better one, given this issue is *about* an unnoticed unpin. Two consequences worth confirming are intended:

- If `yq` is unavailable or `registry-map.yaml` is malformed, the build dies here instead of degrading.
- The `./scripts/build-chunkah.sh` build-from-source path just below becomes unreachable in that case — the script exits before `podman image inspect` runs. That fallback still works for "digest not pullable", just not for "resolution failed".

If fail-loud is intended, it's probably worth saying so in the comment and considering whether line 77 should follow suit, so the file doesn't carry both conventions.

---

Context, since I picked up #1568 and found it already covered: **#1574 also implements this issue** and conflicts with this PR. I've posted a comparison there rather than opening a third PR. FWIW both PRs get the `--prune /sysroot/` trailing slash right — I checked #1574 specifically expecting that bug and it isn't there.

---

PR #1661, opened 2026-08-14
fix(ci): stop piping a remote installer into a root shell, and verify what we download

hanthor:
Refs #1327. **Not the rclone fix** — that already landed in the right repo.

## Where the issue actually stands

#1327 is filed against `tuna-os/tunaos-packages`, a different repo. It's been fixed there: **tunaos-packages#373** replaced all 19 occurrences with a shared composite action that downloads a pinned rclone `.deb` and verifies SHA-256, auto-detecting amd64/arm64. Nothing left to do on that.

## What the issue missed

It was filed **per-vendor** — "the rclone installer" — so nothing went looking for the same shape elsewhere. This repo had it, with a different vendor, in `reusable-build-image.yml`'s *"Install just and yq"* step:

```bash
|| curl -fsSL https://just.systems/install.sh | sudo bash -s -- --to /usr/local/bin
```

Worse than the rclone one in one respect: it was a **fallback** behind a pinned download, so the unpinned root shell ran *precisely when the pinned path had already failed* — and that curl's stderr went to `/dev/null`, so nobody saw why. Same runner, same blast radius: R2 keys and GHCR push credentials.

## Two more holes in the same block

Both survive merely deleting the pipe:

- **Neither download was checksummed.** Versions were pinned, which still trusts whatever bytes the CDN returns.
- **`sudo wget -qO /usr/bin/yq …`** wrote unverified bytes straight to a root-owned system path — there was no moment at which a check *could* have objected.

## The fix

Pinned SHA-256 for both, downloaded to `/tmp`, verified, and only then installed. Unknown architecture is a loud failure instead of a silent continue. The retry loop for the CDN's transient 403s is kept — that was a real documented problem — and now covers both downloads rather than just `yq`.

**Hashes verified two independent ways** on 2026-08-14 — downloaded and hashed here, then compared against each project's own published checksums (`just`'s `SHA256SUMS`; `yq`'s `checksums` + `checksums_hashes_order`, sha256 being column 18). All four agree:

| asset | sha256 |
|---|---|
| `yq_linux_amd64` | `fa52a4e7…eded4` |
| `yq_linux_arm64` | `578648e4…945ea` |
| `just-1.55.1-x86_64-…musl.tar.gz` | `b0ef600f…16c778` |
| `just-1.55.1-aarch64-…musl.tar.gz` | `b0ee814c…1ad173` |

## The regression guard is the generalisation the issue lacked

`tests/bats/test_no_curl_pipe_root_shell.bats` matches the **shape** — `curl`/`wget` piped into `bash`/`sh`, with or without `sudo` — across `.github/workflows`, `scripts/` and `build_scripts/`, rather than one vendor's URL. Plus three assertions that the replacement is genuinely *verified* and not merely *pinned*.

**5/5 on this branch, 1/5 against `upstream/main`** — the shape test catches the `just.systems` pipe.

## Verification

I can't run a build, but I did exercise the parts that are runnable:

```
x86_64   -> YQ_ARCH=amd64 JUST_ARCH=x86_64 YQ_SHA=fa52a4e7
aarch64  -> YQ_ARCH=arm64 JUST_ARCH=aarch64 YQ_SHA=578648e4
riscv64  -> ::error::no pinned yq/just checksum for riscv64
```

YAML parses, and the extracted `run` block passes `bash -n`. (My first attempt at that check selected the wrong YAML path and syntax-checked an empty file — worth saying, since a vacuous green is exactly what this PR is about.)

**Boundary:** the step itself has not run on a real runner. The download URLs and hashes are confirmed against the live releases; that the step executes correctly in Actions is proven by the next build that uses it, not by this PR.

### DIFF
--- a/.github/workflows/reusable-build-image.yml
+++ b/.github/workflows/reusable-build-image.yml
@@ -135,27 +135,62 @@ jobs:
 
       - name: Install just and yq
         run: |
+          set -euo pipefail
           # Pinned versions — /latest/download/ redirects trigger 403
-          # rate-limiting on some runner types (e.g. linux/amd64/v2).
+          # rate-limiting on some runner types (e.g. linux/amd64/v2) — AND
+          # pinned SHA-256. A version pin alone still trusts whatever bytes the
+          # CDN hands back, and these land as root-owned binaries in /usr/bin
+          # and /usr/local/bin on a runner holding R2 keys and GHCR push
+          # credentials (tunaOS#1327).
+          #
+          # Hashes verified two ways on 2026-08-14: downloaded and hashed
+          # independently, and compared against each project's own published
+          # checksums (just's SHA256SUMS, yq's checksums + checksums_hashes_order).
           YQ_VERSION="v4.53.3"
           JUST_VERSION="1.55.1"
-          ARCH=$(uname -m | sed 's/x86_64/amd64/;s/aarch64/arm64/')
-          JUST_ARCH=$(uname -m | sed 's/amd64/x86_64/')
+          case "$(uname -m)" in
+            x86_64)
+              YQ_ARCH="amd64"; JUST_ARCH="x86_64"
+              YQ_SHA256="fa52a4e758c63d38299163fbdd1edfb4c4963247918bf9c1c5d31d84789eded4"
+              JUST_SHA256="b0ef600f0df20d5ae91ae931627c499fc52b477ffe5f5ea7b7b3ec616b16c778"
+              ;;
+            aarch64)
+              YQ_ARCH="arm64"; JUST_ARCH="aarch64"
+              YQ_SHA256="578648e463a11c1b6db6010cbf41eafed6bee79466fcffa1bb446672cf7945ea"
+              JUST_SHA256="b0ee814c9656427408e339893541e30d9027828686839499b2a2a34dd61ad173"
+              ;;
+            *)
+              echo "::error::no pinned yq/just checksum for $(uname -m)" >&2
+              exit 1
+              ;;
+          esac
 
-          # Retry loop — GitHub CDN transient 403s resolve after a short wait
-          for i in 1 2 3; do
-            sudo wget -qO /usr/bin/yq \
-              "https://github.com/mikefarah/yq/releases/download/${YQ_VERSION}/yq_linux_${ARCH}" \
-              && break || sleep $((i * 2))
-          done
+          # Retry loop — GitHub CDN transient 403s resolve after a short wait.
+          fetch() { # url dest
+            local i
+            for i in 1 2 3; do
+              curl -fsSL -o "$2" "$1" && return 0
+              sleep $((i * 2))
+            done
+            echo "::error::failed to download $1 after 3 attempts" >&2
+            return 1
+          }
 
-          curl -fsSL \
-            "https://github.com/casey/just/releases/download/${JUST_VERSION}/just-${JUST_VERSION}-${JUST_ARCH}-unknown-linux-musl.tar.gz" \
-            -o /tmp/just.tgz 2>/dev/null \
-            || curl -fsSL https://just.systems/install.sh | sudo bash -s -- --to /usr/local/bin
+          # Downloaded to /tmp and verified BEFORE anything root-owned is
+          # written. The previous version piped straight to /usr/bin/yq under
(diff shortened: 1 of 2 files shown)


hanthor:
Keeping this open — most of it is still wanted, but it needs a rebase, and I should be straight about why.

I merged #1467 a few minutes ago, which removes the same `just.systems | sudo bash` fallback. I had not read this PR at that point. #1467 is the narrower change: it deletes the pipe and retries the pinned URL, and nothing more. So the *pipe removal* here is now redundant and will conflict.

**Everything else here is strictly better than what landed, and is not on `main`:**

- **Checksum verification for both downloads.** #1467 leaves the gap this PR names precisely — a pinned *version* still trusts whatever bytes the CDN returns, and `sudo wget -qO /usr/bin/yq` wrote unverified bytes to a root-owned path with no point at which a check could object. Pinned SHA-256 for both, verified in `/tmp` before install, is the actual fix.
- **`tests/bats/test_no_curl_pipe_root_shell.bats`.** Matching the *shape* across `.github/workflows`, `scripts/` and `build_scripts/` rather than one vendor's URL is the generalisation #1327 lacked, and it is what stops this recurring with a fourth vendor. #1467 leaves a comment where this PR leaves a test.
- **Loud failure on unknown architecture** instead of silent continue.

So: rebase onto current `main`, drop the pipe-removal hunk as already-done, keep the checksums, the arch guard and the bats test. That is a clean, small PR and I would take it.

Two notes on the rebase. `reusable-build-image.yml` is moving underneath this branch — #1467 just touched this exact block, and #1748 restructures the signing jobs further down the file — so rebase before doing anything else. And your base is `0c8af301`, the same base as the 08-14 06:3x batch (#1579/#1580/#1582), whose diffs turned out to revert five merged fixes; worth checking the rebased diff contains only what you intend.

The "vacuous green" note in your own test plan is the right instinct, and the reason the bats test is the most valuable part of this PR.

---

hanthor:
Triage 2026-09-02 — leaving this open: the change is still wanted and I re-verified its evidence, but it needs a rebase I cannot push (the branch lives on the `hanthor/tunaos-999` fork).

**Half of #1327 already landed.** `main`'s `Install just and yq` step no longer has the `|| curl -fsSL https://just.systems/install.sh | sudo bash` fallback — that was removed under tunaos#1357. What is still missing on `main` is the other half this PR adds: nothing verifies the bytes. `main` still does `sudo wget -qO /usr/bin/yq ...`, writing an unverified download straight into `/usr/bin` on a runner holding R2 keys and GHCR push credentials.

**All four pinned checksums re-verified today** against the upstream release assets:

| asset | expected | downloaded |
|---|---|---|
| `yq_linux_amd64` (v4.53.3) | `fa52a4e7…9eded4` | ✅ match |
| `just-1.55.1-x86_64-unknown-linux-musl.tar.gz` | `b0ef600f…16c778` | ✅ match |
| `yq_linux_arm64` (v4.53.3) | `578648e4…7945ea` | ✅ match |
| `just-1.55.1-aarch64-unknown-linux-musl.tar.gz` | `b0ee814c…1ad173` | ✅ match |

**The conflict is one hunk.** `.github/workflows/reusable-build-image.yml`, the `Install just and yq` step: this branch replaced the download block while `main` separately removed the `just.systems` fallback from the same lines. The resolution is to keep **this branch's** version — it already removes that fallback and adds the verification on top. I checked the merged result locally: it parses as YAML, yamllint reports only pre-existing line-length warnings, and every assertion in the new `tests/bats/test_no_curl_pipe_root_shell.bats` passes against it (no `curl|shell` anywhere in `.github/workflows`, `scripts/`, `build_scripts/` or the `Justfile`; no `sudo wget|curl`; two `sha256sum -c -` calls; the unknown-arch guard present).

Worth carrying that one hunk forward rather than reopening the finding later.

---

hanthor:
Ported into #2281. Main had already dropped the `curl https://just.systems/install.sh | sudo bash` fallback under a different issue number, but the second half of this PR was still missing: yq and just were installed from unverified bytes. Both are now fetched to /tmp, checked against the pinned SHA-256, and only then installed to root-owned paths, and an architecture with no pinned checksum fails instead of guessing. `tests/bats/test_no_curl_pipe_root_shell.bats` comes with it and passes.

---

PR #1544, opened 2026-08-14
docs: add Linux tech press pitch pack

hanthor:
## Summary

- add four maintainer-ready pitches for It's FOSS, OMG! Linux, Linux Magazine, and Phoronix
- add a compact media kit with canonical links, current story facts, screenshots, and technical references
- add a handoff checklist covering fact checks, maintainer approval, follow-up, and response tracking
- index the pack from `docs/README.md`

This prepares the campaign requested by #1534. External outreach remains a maintainer action; no outlet was contacted by this PR.

## Validation

- `git diff --cached --check`

Closes #1534

PR #1667, opened 2026-08-14
docs: fact-check the press materials before they reach editors

hanthor:
Refs #1534. **Not a fifth press document** — both writable actions already have PRs in flight.

## Where the issue stands

| Action | Status |
|---|---|
| 1. Per-outlet pitches | PR **#1544** (`docs/TECH-PRESS-PITCHES.md`, draft) |
| 2. Media kit | PR **#1653** (`docs/PRESSKIT.md`) |
| 3–4. Send pitches, time follow-ups | maintainer actions on the org's behalf |

Plus #1545 (YouTuber review kit) and #1594 in the same area. Adding another document would be waste.

## What isn't covered: whether the materials are true

There's a pattern. **#1594 exists solely** to remove a "weekly release cadence" claim from these files. The maintainer's review of the Gurnard launch post found `bootc container copy` (not a real subcommand) and a "CKI ARK kernel" claim with zero hits anywhere in the repo. I found "56 stars, 40+ repos" wrong in the FOSDEM CFP earlier today.

These are the documents most likely to embarrass the project, because they go to people whose job is to check.

## The DistroWatch draft contradicted the table ROADMAP calls canonical

ROADMAP states the rule in its own words: its variant table *"is the canonical per-variant status; tunaos.org wiki and blog copy must track it."* The draft had drifted five ways:

- **Skipjack** listed `Stable` — canonical says **Beta**
- **Marlin** listed `Alpha` — canonical says **Beta**
- **Gurnard** listed as a headline new variant — canonical says **Experimental** (#1341: no named owner, staff-vs-descope decided at the Q3 checkpoint)
- **COSMIC, Niri, XFCE listed as separate variants.** They're desktop *flavors* available across variants — this would have told DistroWatch the project ships distributions it does not have. Tromsø is a separate repo.
- **Sailfin, Guppy, Grouper, Flounder, Hummingbird** missing entirely

The matrix is now *copied* from ROADMAP rather than hand-authored, since hand-authoring is how it drifted.

## Two capability claims were ahead of the code

**aarch64 ISOs don't exist.** The draft said "All published images are multi-arch (x86_64 and aarch64)" beside "Download: ISOs for all variants" and "Architectures: x86_64, aarch64". Images *are* multi-arch; ISOs aren't — `build-variant.yml`'s artifact matrix hardcodes `platform: "linux/amd64"` (#1378, fix open as #1592). An aarch64 listing promises a download that doesn't exist.

**Apple Silicon isn't shipped.** "Support via an Asahi-based installer" reads as done. `build-config.yml`'s own comment on the asahi flavor says *"EXPERIMENTAL: kernel + glue only; boot payloads pending #777"*, and the macOS installer app's README says it still needs real hardware to prove it launches, drives a backend, or partitions a disk.

I left the release-cadence lines alone — **#1594 owns those**, no point conflicting.

## The guard

`tests/bats/test_press_claims_match_roadmap.bats` enforces the rule ROADMAP already states: every variant press copy names must exist in the canonical table, with the canonical status, and the arch claim must not promise aarch64 ISOs.

**Statuses only — wording is deliberately not pinned.** A test that fails on every prose edit gets deleted, and then it protects nothing.

**4/4 here, 1/4 against `upstream/main`**, naming all eight drifts individually rather than just going red.

One thing running it caught: my first matcher flagged **Bonito** as a nonexistent variant, because ROADMAP names that row "Bonito / Bonito Rawhide". It now accepts either half, so a listing can say "Bonito" without the test forcing compound names into press copy.

### DIFF
--- a/docs/DISTROWATCH-SUBMISSION.md
+++ b/docs/DISTROWATCH-SUBMISSION.md
@@ -18,9 +18,13 @@ Variants cover the desktop spectrum on a common base: GNOME with backports
 onto Enterprise Linux lifecycles (AlmaLinux 10 and CentOS Stream 10),
 KDE Plasma 6, the Rust-built COSMIC, the scrollable-tiling Niri compositor,
 XFCE 4.20, plus Ubuntu-, Debian-, Gentoo-, Arch-, and Fedora-based flavors.
-All published images are multi-arch (x86_64 and aarch64), with Apple Silicon
-(M1/M2) support via an Asahi-based installer and Snapdragon X Elite laptop
-support documented per variant.
+Published container images are multi-arch (x86_64 and aarch64). Downloadable
+ISOs are x86_64 today — the artifact matrix that builds them is still
+amd64-only (#1378), so an aarch64 listing would promise a download that does
+not exist. Apple Silicon (M1/M2) work is in progress rather than shipped: the
+build config marks the Asahi flavor "EXPERIMENTAL: kernel + glue only; boot
+payloads pending #777", and the macOS installer app has not yet been proven on
+real hardware.
 
 Under the hood, TunaOS is manifest-driven: a YAML file defines a variant,
 the pipeline turns it into a published, signed, multi-arch image, and CI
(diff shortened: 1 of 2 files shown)


hanthor:
Rebased on `main` — now conflict-free. Two things worth flagging, because neither was a mechanical merge.

**1. The conflicting variant table had two defects, both fixed.** `main` and this branch had each rewritten it. This branch's version is the right approach — ROADMAP-derived, and it fixes the category error where COSMIC/Niri/XFCE were listed as separate *variants* rather than desktop flavors. But as written it:

- **dropped Yellowfin** — the flagship `Stable` variant, and ROADMAP's first row. A DistroWatch listing missing it would be a worse factual error than the ones this PR set out to fix.
- carried a stray `| Goal | Owner | Tracking | Status |` header row pasted in from a different ROADMAP table.

Resolved by taking this branch's table with both corrected; it now matches ROADMAP's 11 rows exactly.

**2. Your new guard started failing, and it was right to.** `YOUTUBER-REVIEW-KIT.md` (#1545) landed after this branch was written. The guard's own comment says it should be scanned — it is — but it failed on four false positives:

```
NOT A VARIANT: YOUTUBER-REVIEW-KIT.md lists 'Enterprise Linux desktop, stable'
NOT A VARIANT: YOUTUBER-REVIEW-KIT.md lists 'Snapdragon X Elite / ARM laptop'
```

That doc heads its download table `| Story | Variant | Download | Notes |` — the variant is in **column 2**. `variant_rows` matched any four-column row starting with a capital, so it read column 1 as a name and reported prose as undeclared variants.

Fixed in a follow-up commit: identify a variant table by its **header** (first column literally `Variant`) rather than by row shape. This file already makes the argument — *"a false positive is how a guard gets deleted, so it matters more than the true positives"* — and four of them on a doc the guard deliberately scans is exactly that.

**Verified both directions**, not just green: setting Skipjack to `Stable` fails the status test, and adding an invented variant fails the existence test. 5/5 pass otherwise.

---

PR #1675, opened 2026-08-14
ci: bound upstream snapshot refresh size

### LINKED ISSUE #1647: [architect] _upstream-snapshots/ vendored wholesale: 241 files / 1.9MB committed every refresh with no size gate — git bloat
## Architecture Finding

**Type**: tech-debt
**Affected area**: `_upstream-snapshots/{aurora,bluefin-lts,zirconium}/`

`_upstream-snapshots/` is vendored wholesale into git: **241 files, 1.9MB**, and growing (measured 920KB at first finding). The refresh workflow (`snapshot-upstreams.yml`) pins each snapshot to an upstream commit via `.snapshot.json`, but it commits the **entire tree** every refresh with **no size gate** — every refresh permanently bloats git history with content-level copies of three upstream trees.

## Impact

- `.git` grows monotonically on every upstream refresh; history bloat compounds (a 1.9MB tree × many refreshes).
- The "diff is the review surface" design means reviewers must eyeball wholesale tree diffs — the review surface is the whole tree, not the deltas.
- `lint.yml` and `Justfile` must special-case `-not -path './_upstream-snapshots/*'` exclusions, and the directory is effectively invisible to quality gates.

## Recommendation

Add a size/content gate to `snapshot-upstreams.yml` (fail or prune when a refresh exceeds a threshold, e.g. >N files or >X MB of new content). Consider snapshotting only the *changed* files as diffs, or moving the full snapshots to an artifact store (GHCR/R2) and committing only `.snapshot.json` manifests.

---
*Filed by architect agent (ACMM L6 — full mode)*

hanthor:
## Summary

- Add a fail-closed size/content budget to the upstream snapshot refresh workflow.
- Reject refreshes over 300 files, 2 MiB, or 150 changed paths before they are staged or committed.
- Document the helper alongside the existing snapshot sync script.

## Why

`_upstream-snapshots/` is intentionally kept in git as a filtered review surface, but the refresh workflow previously admitted any tree size. A broad upstream path change could silently add a wholesale vendor tree and permanently increase repository history. The new check keeps normal refreshes working while requiring an explicit follow-up for an over-budget change.

## Validation

- `bash -n scripts/check-upstream-snapshot-size.sh`
- Snapshot budget passes on the current tree: 241 files, 229,587 bytes, 0 changed paths.
- Forced low file and byte limits both fail with actionable GitHub error messages.
- `git diff --check`

Fixes #1647.

### DIFF
--- a/.github/workflows/snapshot-upstreams.yml
+++ b/.github/workflows/snapshot-upstreams.yml
@@ -61,6 +61,16 @@ jobs:
         run: |
           ./scripts/sync-upstream-snapshots.sh "${UPSTREAM}"
 
+      - name: Enforce snapshot size budget
+        # Keep the vendored review surface bounded. A refresh over budget must
+        # be split, narrowed, or explicitly re-designed before it can create
+        # another large git snapshot commit (tuna-os/tunaos#1647).
+        env:
+          MAX_SNAPSHOT_FILES: '300'
+          MAX_SNAPSHOT_BYTES: '2097152'
+          MAX_SNAPSHOT_CHANGED_FILES: '150'
+        run: ./scripts/check-upstream-snapshot-size.sh
+
       - name: Detect changes
         id: diff
         run: |
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -18,6 +18,7 @@ Invoked by:
 | `resolve-image.sh` | Resolves image references (base, common, brew, akmods) |
 | `build-image-inner.sh` | The build engine (env-var driven) |
 | `sync-upstream-snapshots.sh` | Syncs and drift-checks the `_upstream-snapshots/` tree |
+| `check-upstream-snapshot-size.sh` | Fails refreshes that exceed the snapshot size or change budget |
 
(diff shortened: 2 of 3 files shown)


hanthor:
Triage note before merging: the red **Unit Tests** check on this PR is pre-existing and unrelated.

The only failing assertion in [run 31834604858](https://github.com/tuna-os/tunaOS/actions/runs/31834604858) is `not ok 5 Justfile contains custom recipes` — a bats test that was stale against `main` at the time (the custom recipes had moved to `just/custom-overlay.just`; `main` now carries `tests/bats/test_custom_overlay.bats` instead). No pytest test failed in that run, and this PR touches only `.github/workflows/snapshot-upstreams.yml`, `scripts/check-upstream-snapshot-size.sh`, and `scripts/README.md`.

Budget check against current `main`: `_upstream-snapshots/` is 239 files / 203,170 bytes, comfortably inside the 300-file / 2 MiB caps, so the new gate does not trip on the existing tree.

Merging.

---

PR #1689, opened 2026-08-14
[sec-check] harden GitHub context interpolation in workflows

### LINKED ISSUE #1181: [sec-check] github.* context interpolated into run: blocks (Actions script-injection hardening)
## Security Finding

**Severity**: low
**Type**: unsafe-pattern (hardening)

\~40 spots across the org interpolate `github.*` context directly into `run:` shell commands instead of passing it through `env:`. Examples:

- `bootc-installer/.github/workflows/flatpak.yml:164` — **unquoted**: `gh release create ${{ github.ref_name }}`
- `tromso/.github/workflows/lint.yml:120-121` and `xfce-linux/.github/workflows/lint.yml:106-107` — `git fetch origin "${{ github.event.pull_request.base.ref }}"`
- `tunaos-packages/.github/workflows/validate-package-factory.yml:68` — `git fetch ... origin "${{ github.base_ref }}"`
- Multiple publish-flatpak workflows — `git commit -m "chore(flatpak): update ... [${{ github.ref_name }}]"`
- `corral/.github/workflows/release-brew.yml:62-63` — release tag interpolated into `download-url`

## Impact

GitHub's security hardening guidance explicitly warns against inline context interpolation in `run:`: if the interpolated value ever carries attacker-controlled text (e.g. a PR base ref, branch or tag name crafted with shell metacharacters), the expression is expanded into the command string and executed by the shell before the script runs. Today the org's branch/tag creation requires write access, so exploitability is limited — but this is a classic footgun (a malicious branch name like `main$(curl ...)` would execute) and the standard fix is mechanical.

## Recommendation

Pass context values through `env:` and reference them inside the script:

```yaml
- name: Release
  env:
    REF_NAME: ${{ github.ref_name }}
  run: |
    gh release create "$REF_NAME"
```

Also quote all interpolated values (note the unquoted `gh release create` in bootc-installer).

---
*Filed by sec-check agent (ACMM L6 — full mode)*

hanthor:
## Summary

- pass GitHub context values through step environment variables before using them in shell scripts
- quote actor, repository, ref, event, and token values in affected workflow commands
- remove all direct `${{ github.* }}` interpolations from `run:` bodies in this repository

## Security

This prevents shell parsing of attacker-controlled branch, tag, event, and repository context values during Actions execution, addressing tuna-os/tunaos#1181.

## Validation

- `git diff --check`
- custom workflow scanner: zero direct `github.*` expressions in `run:` bodies
- pushed from `agent/issue-1181-context-hardening`

Closes #1181

### DIFF
--- a/.github/workflows/archive/build.yml
+++ b/.github/workflows/archive/build.yml
@@ -98,6 +98,7 @@ jobs:
         env:
           PR_HWE_GDX: ${{ needs.detect_pr_scope.outputs.hwe_gdx_related || 'false' }}
           EVENT_NAME: ${{ github.event_name }}
+          RECHUNK: ${{ github.event_name != 'pull_request' }}
           INPUT_IMAGE_VARIANT: ${{ inputs.image_variant }}
           INPUT_PLATFORMS: ${{ inputs.platforms }}
           INPUT_BUILD_VARIANT: ${{ inputs.build_variant }}
@@ -225,7 +226,7 @@ jobs:
               \"run_kde\": ${RUN_KDE},
               \"run_kde_hwe_gdx\": ${RUN_KDE_HWE_GDX},
               \"run_niri_hwe_gdx\": ${RUN_NIRI_HWE_GDX},
-              \"rechunk\": ${{ github.event_name != 'pull_request' }},
+              \"rechunk\": ${RECHUNK},
               \"sbom\": false,
               \"publish\": true
             }]")"
--- a/.github/workflows/archive/generate-release.yml
+++ b/.github/workflows/archive/generate-release.yml
@@ -57,8 +57,9 @@ jobs:
         id: target
         env:
           TARGET: ${{ inputs.target }}
+          EVENT_NAME: ${{ github.event_name }}
         run: |
-          if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
+          if [ "${EVENT_NAME}" = "workflow_dispatch" ]; then
             echo "target=${TARGET}" >> $GITHUB_OUTPUT
           else
             echo "target=lts" >> $GITHUB_OUTPUT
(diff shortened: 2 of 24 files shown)


hanthor:
Triage 2026-09-02 — leaving this open. Still not superseded, but it needs a rebase I cannot push (the branch is on the `hanthor/tunaos-999` fork).

Spot-checked against current `main`: `bootc-lifecycle.yml` still has, twice,

```yaml
run: |
  echo "${{ env.GITHUB_TOKEN }}" | \
    podman login ghcr.io -u "${{ github.actor }}" --password-stdin
```

so the interpolation-into-`run` shape this PR replaces with `env:` indirection is still there. The change is right and mostly mechanical across 24 workflows.

What a rebase has to deal with:

- **`.github/workflows/archive/build.yml` and `archive/generate-release.yml` were deleted on `main`.** Drop those two hunks rather than resurrecting the files.
- Content conflicts in `installer-smoke.yml` (rewritten by the #1929 build/smoke split), `iso-e2e.yml` and `just-fix.yml`.
- Everything else in the branch auto-merges.

One nit to fix while rebasing: the `build-toolchain.yml` hunk switches to `${GITHUB_ACTOR}` without adding it to that step's `env:`. It happens to work because the runner exports `GITHUB_ACTOR`, but it is inconsistent with the explicit `env:` block the other 23 files use.

---

hanthor:
Cross-reference for whoever picks this up: #1433 ("ci: harden Actions script-injection surface (`github.*` interpolation in `run:` blocks)") is open and covers the same class of change. Worth deciding which branch to carry forward before rebasing either — they will conflict with each other.

---

hanthor:
Ported into #2281 across 22 workflows. Confirmed not superseded: 115 direct `${{ github.* }}` interpolations remained inside `run:` blocks on main.

Three hunks needed a judgement call rather than a mechanical resolve, and I did not take this branch's side on them:

- `iso-e2e.yml` — main deleted that dispatch filter on purpose. With its inputs blanked it set `skip=true` on every cell, and run 32791464847 swept six cells and reported success having booted nothing. Main's side kept.
- `installer-smoke.yml` — `sudo -n podman` stays, because rootless re-exec is broken on that runner; the env-var form for the actor is taken, which is this PR's actual point.
- `just-fix.yml` — both `IS_FORK` and `REF_NAME` are read further down, so both are kept.

The two `.github/workflows/archive/` hunks are dropped: those files no longer exist on main.

---

PR #1690, opened 2026-08-14
fix(cachyos): require one kernel tree before rebuilding initramfs

### LINKED ISSUE #1563: [marlin] cachyos overlay keeps two kernels but builds one initramfs into a `find | tail -1` gamble — all 5 cachyos flavors fail Gate (bootc "initramfs not found" / boot never reaches graphical.target)

hanthor:
## Summary

- stop running a full `pacman -Syu` in the CachyOS overlay
- remove the stock `linux` package when present
- select and validate the sole remaining module tree before rebuilding its initramfs

## Validation

- `bash -n build_scripts/overlay/cachyos.sh`
- `git diff --check`
- static regression assertions in `tests/bats/test_cachyos_overlay.bats` (Bats is unavailable in this checkout)

Fixes #1563
Related to #1570

### DIFF
--- a/build_scripts/overlay/cachyos.sh
+++ b/build_scripts/overlay/cachyos.sh
@@ -38,14 +38,35 @@ Include = /etc/pacman.d/cachyos-mirrorlist
 EOF
 fi
 
-pacman -Syu --noconfirm --needed \
+# Do not upgrade the base image here: a full -Syu can install or update the
+# stock linux package alongside linux-cachyos, leaving bootc with multiple
+# kernel trees.  The image must have one kernel whose initramfs is unambiguous.
+pacman -S --noconfirm --needed \
 	cachyos-keyring cachyos-mirrorlist cachyos-settings \
 	linux-cachyos linux-cachyos-headers tpm2-tss
+
+# linux-cachyos is the kernel selected for CachyOS images.  Remove Arch's
+# stock kernel after the replacement is installed; -Rdd is intentional here
+# because the stock kernel's dependencies are also provided by the CachyOS
+# kernel package, but pacman cannot infer that relationship from package
+# metadata.
+if pacman -Qq linux >/dev/null 2>&1; then
+	pacman -Rdd --noconfirm linux
+fi
 
 # Mark as CachyOS-augmented for install-desktop.sh detection
 install -D /dev/null /etc/cachyos-release
 printf 'CachyOS\n' >/etc/cachyos-release
 
-# Rebuild initramfs via dracut
-dracut --force --omit "tpm2-tss" "$(find /usr/lib/modules -maxdepth 1 -type d | grep -v '\.img' | tail -1)/initramfs.img"
+# Rebuild initramfs via dracut.  Do not use find | tail -1: directory order is
+# unspecified, and selecting the wrong tree leaves the bootloader with a
+# kernel that has no matching initramfs.
+mapfile -t _module_dirs < <(find /usr/lib/modules -maxdepth 1 -mindepth 1 -type d ! -name '*.img' | sort -V)
+if [[ "${#_module_dirs[@]}" -ne 1 ]]; then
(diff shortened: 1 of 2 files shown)


hanthor:
Closing as superseded — `main` already requires exactly one kernel tree before rebuilding the initramfs.

`build_scripts/overlay/cachyos.sh` on `main` selects `KVER` by name, hard-fails on an empty selection, deletes leftover module trees, and then asserts `REMAINING_COUNT == 1` **before** the `dracut` call — which now passes both the output path and the kernel version positionally. `tests/bats/test_cachyos_overlay.bats` pins each of those (`exactly one remaining kernel is proven, not hoped for`, `the single-kernel assertion runs BEFORE the initramfs is built`, `cachyos.sh passes both the output path and the kernel version to dracut`).

That is this PR's premise, implemented and tested. The branch also conflicts in `build_scripts/overlay/cachyos.sh` and `tests/bats/test_cachyos_overlay.bats` (add/add against the file `main` now has). See also #1605, closed for the same reason.

---

PR #1698, opened 2026-08-14
build(just): extract image pipeline recipes into a module

hanthor:
## Summary

This continues the local modular Justfile pattern established by `just/custom-overlay.just`, `just/utilities.just`, and `just/vm-pipeline.just`.

- moves `build`, lifecycle/corral helpers, and ISO recipes into `just/image-pipeline.just`
- keeps the existing recipe names and flat imported namespace
- updates Justfile tests to find recipes in imported modules
- reduces the root Justfile from 432 to 305 lines on this base

## Why

`just` resolves imports from local files; it does not fetch Git URLs. This keeps the change within the working mechanism already used by tunaOS while reducing root-file inflation and establishing another reusable module boundary.

## Validation

- `git diff --check`
- static import and recipe-presence assertions passed
- `bats` and `just` are unavailable in this checkout, so runtime parsing and Bats execution could not be run locally

Refs #508

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -12,6 +12,7 @@ export platform := env("PLATFORM", if arch == "x86_64" { if `rpm -q kernel 2>/de
 import 'just/utilities.just'
 import 'just/custom-overlay.just'
 import 'just/vm-pipeline.just'
+import 'just/image-pipeline.just'
 
 # ==============================================================================
 #  BUILD PIPELINE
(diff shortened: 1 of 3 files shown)


hanthor:
Closing as stale — this is a pure move with no behavior change, and the file it moves from has since changed underneath it.

`Justfile` on `main` now imports four modules (`just/utilities.just`, `just/custom-overlay.just`, `just/vm-pipeline.just`, `just/qcow2-build.just`) and the recipes this branch lifts (`build`, `iso`, `iso-group`, `lifecycle-test`, `corral-build`) have been edited in place since 2026-08-14. The result is a content conflict in `Justfile`, and because the diff is "delete 129 lines here, add 77 there", every future edit to those recipes re-conflicts.

There is nothing to preserve in the branch beyond the mechanical move, so redoing the extraction against current `main` is strictly cheaper than resolving it. #508 stays open for that.

---

PR #1720, opened 2026-08-14
ci: stop new bare error suppressions

### LINKED ISSUE #1652: [architect] Systematic || true error suppression — 75 workflow + 20 just occurrences mask build failures
## Architecture Finding

**Type**: structural-risk (silent failure)
**Affected area**: `.github/workflows/` (75 occurrences) + `Justfile`/`just/*.just` (20 occurrences)

Systematic `|| true` error suppression masks real failures across the build pipeline: **75 occurrences in workflows, 20 in just modules**. Examples: `just/custom-overlay.just` uses `python3 -c ... 2>/dev/null || echo "default"` fallbacks that hide parse errors; multiple workflow steps end in `|| true` around verification steps.

## Impact

- CI can report green while the actual work silently failed (cf. tunaos#1533 — the boot-verify E2E was swallowed by `|| true` for weeks; this is the same pattern at scale).
- Each `|| true` is a potential maintenance trap: the fallback becomes the de-facto behavior and the error path is never exercised.
- The pattern is a *systemic* cause, not a one-off bug — fixing one instance doesn't fix the class.

## Recommendation

Audit and triage every `|| true`:
1. Replace with explicit `|| { echo "…"; exit 1; }` where the step is required.
2. Where failure is legitimately non-fatal, replace with a named guard (e.g. `if ! cmd; then echo "skip reason"; fi`) that emits a visible warning.
3. Add a lint rule to forbid bare `|| true` in workflows and just modules (shellcheck/Stylelint), requiring a comment justifying each suppression.

---
*Filed by architect agent (ACMM L6 — full mode)*

hanthor:
## Summary

- Make the reusable image boot-verification gate fail when `iso-e2e.sh` fails instead of swallowing the result.
- Propagate `yq` parse errors in the custom overlay pipeline instead of silently selecting defaults.
- Add a lint guard that rejects newly introduced bare `|| true` suppressions in workflows and Just files.

## Why

Issue #1652 identifies systematic error suppression as a source of silent build failures. This patch removes the highest-risk existing suppression and prevents the inventory from growing while the remaining best-effort cleanup and diagnostic cases receive deliberate follow-up triage.

## Validation

- `bash -n scripts/check-new-error-suppressions.sh`
- `bash scripts/check-new-error-suppressions.sh`
- `git diff --check`
- `just`, `yamllint`, `actionlint`, and `shellcheck` were unavailable locally; the corresponding CI jobs should validate them.

Closes #1652

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -44,6 +44,7 @@ jobs:
           - json-validate
           - actionlint
           - justfmt
+          - error-suppression
     steps:
       - name: Checkout
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
@@ -111,6 +112,10 @@ jobs:
             -ignore "save-always" \
             -ignore "cannot be filtered" \
             .github/workflows/*.yml .github/workflows/*.yaml
+
+      - name: Check new error suppressions
+        if: matrix.check == 'error-suppression'
+        run: bash scripts/check-new-error-suppressions.sh
 
       - name: Check Justfile formatting
         if: matrix.check == 'justfmt'
--- a/.github/workflows/reusable-build-image.yml
+++ b/.github/workflows/reusable-build-image.yml
@@ -265,7 +265,7 @@ jobs:
           mkdir -p verify-out
           IMAGE="localhost/${IMAGE_VARIANT}:${DEFAULT_TAG}"
           sudo -E corral create gate --bootc "$IMAGE" --wait-ssh --timeout 900
-          sudo ./scripts/iso-e2e.sh "$IMAGE" --disk --output verify-out --timeout 300 || true
+          sudo ./scripts/iso-e2e.sh "$IMAGE" --disk --output verify-out --timeout 300
 
       - name: Upload PR verification evidence
         if: always() && github.event_name == 'pull_request' && contains(matrix.platform, 'amd64')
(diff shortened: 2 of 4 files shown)


hanthor:
Closing — the new guard cannot fire as written, and part of the branch has already landed.

**The lint leg is a permanent no-op.** `scripts/check-new-error-suppressions.sh` starts with:

```bash
if ! git rev-parse --verify HEAD^ >/dev/null 2>&1; then
    echo "No parent commit available; skipping new suppression check."
    exit 0
fi
```

`lint.yml`'s `actions/checkout` step sets no `fetch-depth`, so it clones at depth 1 and `HEAD^` never resolves — on pull_request *or* push. The `error-suppression` matrix leg would therefore report green for every PR without ever reading a diff, which is worse than not having it: it is exactly the silent pass #1652 is about.

**One hunk is already on `main`.** `reusable-build-image.yml` line 332 now reads `sudo ./scripts/iso-e2e.sh "$IMAGE" --disk --output verify-out --timeout 300` with no `|| true`.

**What is still worth doing** is the `just/custom-overlay.just` half: lines 22, 27 and 78 still read `$({{ yq }} -r '...' custom/image.yaml 2>/dev/null || true)` directly beneath a comment explaining that this exact suppression turned an AttributeError into a silently-wrong default (tunaOS#1651). Dropping the suppression there contradicts nothing on `main` and needs no CI plumbing. Line 44 has the same shape and should go with them.

A follow-up that does that, plus (if the guard is still wanted) `fetch-depth: 2` on the lint checkout so `HEAD^` exists, would land cleanly. Reopening this branch is not the shortest path — it is three weeks behind `main`.

---

Issue #1729, opened 2026-08-15
[ci] One arch's runner fault zeroes a whole variant — stage-2 is hard-gated on a fully-green multi-arch base (14 flavors never built on 08-14)


hanthor:
## What

`build_stage2` is skipped whenever `build_base` reports `failure`:

```yaml
# .github/workflows/build-variant.yml:220
needs.build_base.result != 'failure' && needs.build_base.result != 'cancelled' &&
```

and `build_base` reports `failure` when **any single platform leg** of the base fails — the base's `manifest` job exits on the first non-success `needs` result:

```yaml
# .github/workflows/reusable-build-image.yml:589  ("Exit on failure")
for i in $(echo "${JOBS}" | jq -r 'to_entries[] | .value.result'); do
  if [ "$i" != "success" ] && [ "$i" != "skipped" ]; then exit 1; fi
done
```

So a single hosted-runner fault on the **arm64** leg of `base` stops every stage-2 desktop flavor from being built at all — including flavors declared **amd64-only**, which never consume an arm64 base.

Stage 3 was already decoupled from this on purpose (`build-variant.yml:250`, comment: *"stage-3 cells whose specific base is missing fail on their own without taking the rest down"*). Stage 2 was not, and it is the stage that carries the five desktops.

## Evidence — 2026-08-14 nightly sweep

In four variants the *only* base failure was `sudo podman login` on the arm64 leg (the `libpod_lock` fault tracked in #1556). The amd64 leg built, pushed to GHCR, published its package manifest and its SBOM. The manifest job then failed at "Exit on failure", and stage 2 never ran:

| Variant | Run | base amd64 | base arm64 | stage-2 flavors never attempted |
| :--- | :--- | :--- | :--- | :--- |
| `sailfin` | [31766602231](https://github.com/tuna-os/tunaOS/actions/runs/31766602231) | ✅ pushed | ❌ login | gnome, kde, niri, xfce, cosmic |
| `marlin` | [31766586852](https://github.com/tuna-os/tunaOS/actions/runs/31766586852) | ✅ pushed | ❌ login | gnome, kde, cosmic, niri, xfce |
| `flounder` | [31766676694](https://github.com/tuna-os/tunaOS/actions/runs/31766676694) | ✅ pushed | ❌ login | gnome, kde, xfce |
| `gurnard` | [31766667186](https://github.com/tuna-os/tunaOS/actions/runs/31766667186) | ✅ pushed | ❌ login | pantheon |

[...]

Issue #1731, opened 2026-08-15
[ci] No recovery path between nightlies — a transient infra fault costs a full day of matrix state, and every #1570 fix waits ~22h for its first evidence


hanthor:
## What

Each variant build is a single daily cron with no automated re-run:

```yaml
# .github/workflows/build-sailfin.yml (same shape in all 13 build-<variant>.yml)
on:
  schedule:
    - cron: "0 1 * * *"
  workflow_dispatch:
```

Nothing in `.github/workflows/` re-runs a failed variant run. Recovery is a human dispatching the workflow by hand. Combined with a build graph where one failing leg takes the variant with it (#1729), the practical unit of failure is *one variant-day*, and the unit of evidence is *one variant-night*.

## Why this blocks green specifically

**1. Transient faults cost a whole day.** The 08-14 sweep lost 13 of 13 variants, and the two largest contributors were both transient and both external:

- `rekor.sigstore.dev` returning 502 for 45+ minutes, failing 25 `Sign + attest` jobs across 7 variants (#1560). Every one of those images had already built, pushed and produced an SBOM — e.g. [grouper base](https://github.com/tuna-os/tunaOS/actions/runs/31766671286), which signed the index and one platform image successfully before the log went down mid-run.
- `sudo podman login` dying on `ubuntu-24.04-arm` runners (#1556), which on that night hit eight arm64 legs across `flounder`, `gurnard`, `sailfin`, `marlin`, `bonito` and `yellowfin`.

#1570 notes the remedy — *"A plain re-run after Rekor recovers should clear most of the 08-14 flake damage"* — but nothing performs it, so the README matrix sat at 6/142 for the rest of the day.

**2. Fix latency is a full cycle.** All nine merged fixes for #1570 landed **after** that morning's nightly:

| Commit | Landed (UTC) | Issue |
| :--- | :--- | :--- |
| `a641d20` | 08-14 05:33 | #1556 podman lock reset |
| `2a576f0` | 08-14 05:39 | #1560 Rekor backoff, #1567 syft bound |
| `507f510` | 08-14 06:22 | #1562 SELinux copy-up |
| `966cd80` | 08-14 15:05 | #1566 rawhide openh264 |
| `582ee71` | 08-14 15:28 | #1561 initramfs parity |
| `4dbef50` | 08-14 15:40 | #1565 sid nvidia headers |
| `cca7549` | 08-14 15:41 | #1563 cachyos initramfs |
[...]

Issue #1730, opened 2026-08-15
[ci] README live build matrix reports "never built" as "failing" — 14 of the 08-14 red cells produced no job at all


hanthor:
## What

`.github/scripts/update-build-status.sh` decides a cell's colour from the conclusion of its `<flavor> / Promote` job, and sorts every non-`success` value into one bucket:

```bash
# .github/scripts/update-build-status.sh:57-65
promotion=${promotion:-missing}
if [[ "$promotion" == success ]]; then
  green=$((green + 1))
else
  failed+=("$flavor")     # column header: "Blocked or failing tags"
fi
```

`missing` (no Promote job ever existed), `skipped` (an upstream job stopped it), `cancelled` and a real `failure` all land in the same list. The run-level icon has the same problem: the script takes the latest run with `--status completed`, which includes **cancelled** runs, and renders anything that is not `success` as ❌.

## Why it matters here

This is the exact conflation `docs/MATRIX-STATUS.md` was written to prevent — *"Before it existed, 'no result' and 'passing' looked identical from the outside, and three separate defects survived for weeks because the absence of evidence read as evidence of absence."* — and `docs/MATRIX-STATUS.md` even lists the cancelled-run misreading in *Failures that look like successes*: *"`gh pr checks` renders a cancelled run as `fail`. A superseded run is not a broken build."* The README generator makes both mistakes, on the page that is the project's front door.

## Evidence — current README snapshot

The live matrix reads `sailfin` **0/7** with `base,gnome,gnome-asahi,kde,niri,xfce,cosmic` under "Blocked or failing tags". In [run 31766602231](https://github.com/tuna-os/tunaOS/actions/runs/31766602231), `kde`, `niri`, `xfce` and `cosmic` have no jobs at all — the whole stage-2 group is one collapsed `matrix.flavor` placeholder, because the base manifest failed first (see #1729). Nothing about those four flavors was tested on 08-14; the README asserts they are blocked or failing.

Across the same sweep the count is 14 never-attempted flavors presented as red cells (`sailfin` 5, `marlin` 5, `flounder` 3, `gurnard` 1). `gnome-asahi` on several variants is a third case again: `skipped` because its Promote was gated, not because it failed.

Practical cost: "**Current image coverage: 6/142 cells (4%)**" cannot be read as "136 broken cells", and there is no way to tell from the page which of the 136 have a defect behind them and which were never reached. That is precisely the triage question #1570 is trying to answer.

## Suggested shape

[...]

PR #1748, opened 2026-08-15
fix(ci): stop a Sigstore outage and one arch's runner fault from zeroing the matrix

### LINKED ISSUE #1564: [flounder] trixie nvidia overlay omits libnvidia-allocator1 → nvidia-drm_gbm.so missing → Debian contract check deterministically fails all nvidia flavors

hanthor:
## The run that makes the case

[Run 31858324517](https://github.com/tuna-os/tunaOS/actions/runs/31858324517) (albacore, 2026-08-15). Every flavor built, pushed, generated an SBOM and passed its gates. **All twelve Promote jobs were skipped** — because `cosign attest` could not reach Rekor:

```
Error: signing ghcr.io/tuna-os/albacore@sha256:45d949df...: creating bundle:
  Post "https://rekor.sigstore.dev/api/v1/log/entries": status 502 Bad Gateway
##[warning]cosign failed (attempt 6/6); retrying...
##[error]Process completed with exit code 1.          ← 16m19s in
```

Nothing was wrong with the images. The nine deterministic fixes merged for #1570 on 08-14 all worked — the only build failure left in that whole run was one Gate. What is left standing between the matrix and green is not defects in the images; it is the pipeline turning transient, external events into a lost day.

## Changes

**SBOM attestation leaves the critical path.** Signing and attesting shared one job and Promote required all of it. Split into `sign` (image + index signatures, still a hard promotion precondition) and `attest_sbom` (`continue-on-error`, absent from Promote's `needs`).

This is [projectbluefin/dakota](https://github.com/projectbluefin/dakota)'s shape: `publish-image` signs, `publish-sbom` attaches and signs with `continue-on-error`, and `promote` needs only `publish-image` — *"Runs outside the critical path to :testing. Failures here must not block publish/promotion."* An image signature is a supply-chain invariant; an attestation that a transparency-log outage delayed by an hour is not. On the run above this change alone promotes 12 cells.

The per-platform `cosign verify` calls are gone too — one verify of the index proves the same identity and issuer, and each of the others was another Rekor round trip that could block a promotion.

**The retry budget becomes a deadline.** Six attempts with a doubling sleep *reads* like a long budget and delivered 16m19s against a longer outage. It also slept 480s after the final attempt, for nothing — every failing job ran 8 minutes longer than it had reason to. `cosign_retry` moves to `.github/scripts/cosign-retry.sh`:

| | before | after |
|---|---|---|
| budget | 6 attempts (~15m45s if instant) | 40-minute wall clock |
| backoff | doubling to 8m | doubling, capped at 2m — a long outage is probed 20× instead of 6× |
| after last attempt | sleeps 480s | returns |
| bad identity / registry 403 | full budget, then fails | fails immediately |
| exhaustion | generic error | `SIGSTORE_OUTAGE` marker for the re-run classifier |

**Stage 2 stops depending on a fully-green multi-arch base** (#1729). `build_stage2` gated on `build_base`'s conclusion, which is red if *any* platform leg failed — so one arm64 `libpod_lock` fault skipped every desktop flavor, including flavors declaring only amd64 that never touch the arm64 base. On 08-14 that cost 14 flavors across `sailfin`/`marlin`/`flounder`/`gurnard`, none of which produced a job at all.

The manifest job now tolerates a partial platform set (warns, omits the missing arch) and fails only when *nothing* built, exposing `manifest_published` for callers to gate on. Completeness is still enforced where it belongs — `all_platforms_built` stays false and the bare `:<flavor>` tag keeps its last good version. **A partial build reaches `:<flavor>-testing` and can never reach `:<flavor>`.**

**Infra failures get one automatic re-run** (#1731). `rerun-infra-failures.yml` classifies a failed nightly against signatures anchored to named runs — `SIGSTORE_OUTAGE`, `libpod_lock`, runner death (which uploads no logs, so the 404 *is* the signature) — waits 45 minutes, and re-runs the failed jobs once. `run_attempt == 1` only, so never a loop. Any failure it cannot positively identify as infra leaves the run red: a re-run loop over a real defect burns minutes while hiding it.

**`libnvidia-allocator1` on flounder** (#1564). It ships `nvidia-drm_gbm.so` and is only a *Recommends* of `nvidia-driver-libs`, so `--no-install-recommends` dropped it and every flounder `*-nvidia` image failed the Debian contract on a missing GBM backend while the rest of the driver stack was correct. Named explicitly rather than dropping `--no-install-recommends`, whose recommends set pulls in Xorg.

**"Never built" stops reading as "failing"** (#1730). The README generator bucketed `missing`/`skipped`/`cancelled` with `failure`, so 14 cells that produced no job at all read as broken, and a superseded (cancelled) run rendered as ❌ — the exact misreading `docs/MATRIX-STATUS.md` exists to prevent, on the front page. Now two columns, counted separately, and the run picker skips cancelled runs.

## Testing

`tests/test_sign_step_retries_cosign.py` rewritten — 20 passing. It drives the real helper rather than a reimplementation, and pins the promotion contract in both directions: Promote **must** require `sign`, and **must not** wait on `attest_sbom`.

```
tests/test_sign_step_retries_cosign.py tests/test_readme_build_status_refresh.py  20 passed
tests/test_boot_gate_ownership.py tests/test_generate_workflows.py tests/test_image_aliases.py  29 passed
```

Verified directly: non-transient error fails in 0s (was 15m45s), a transient one exhausts its deadline with no trailing sleep and emits the marker, and a flake clearing on attempt 2 succeeds.

## Not in this PR

Razorfin signs with a password-less key pair (`cosign sign --key env://COSIGN_PRIVATE_KEY`). With a key pair, `--tlog-upload=false` becomes possible and sigstore.dev leaves the critical path entirely — keyless can't do that, because the ephemeral Fulcio cert needs the log to establish validity. tunaOS already carries a `cosign.pub` at the repo root from the bluefin port.

That is a real answer to this class of outage, but it trades transparency-log auditability and no-long-lived-key for availability, and `SECURITY.md` plus #1187 are both written around the keyless model. It's a maintainer call, not a drive-by change — worth its own issue if you want it.

## Verification

Next nightly: a variant whose Sign job hits Rekor should still promote, and a variant that loses one platform leg should still build the flavors that declare the other one.

Fixes #1564, #1729, #1730, #1731
Refs #1560, #1556, #1570

---


hanthor:
Rebased onto current `main`. Two of the six changes here landed upstream while this was open, so the scope has narrowed — updating the record rather than leaving the description overstating it.

**Dropped, because `main` already has them:**

- **#1564** (flounder `libnvidia-allocator1`). Resolved in favour of `main`'s version, which is the better one: it explains that `libnvidia-egl-gbm1` ships `15_nvidia_gbm.json`, and that without that file EGL never loads the GBM external platform — so `nvidia-drm_gbm.so` would sit on disk unused and the contract check would pass while the thing it guards still did not work. My comment only said the packages were Recommends.
- **#1730** (README matrix classification). Same split, same reasoning, plus a `join_or_dash` helper. Took `main`'s.

**Still unique to this PR:**

- The `sign` / `attest_sbom` split — SBOM attestation off Promote's critical path (#1560).
- `.github/scripts/cosign-retry.sh` — deadline instead of attempt count, fail-fast on non-Sigstore errors, no trailing sleep.
- Stage-2 decoupling from a fully-green multi-arch base (#1729), via `manifest_published`.
- `rerun-infra-failures.yml` (#1731).

27 tests pass post-merge, including the promotion-contract assertions in both directions.

So this now **fixes #1729 and #1731** and refs #1560; #1564 and #1730 are closed by what is already on `main`.

Worth noting for #1694: three of the things I set out to fix this morning were fixed independently in parallel while I worked. That is the queue behaving as the staleness issue predicts, and it is cheaper to notice at rebase time than at merge time.

---

PR #1766, opened 2026-08-15
fix(ci): retry promotion copies, unbreak pytest collection, bound the non-blocking attestation

hanthor:
Three faults, all found by watching gurnard [run 31891742138](https://github.com/tuna-os/tunaOS/actions/runs/31891742138) — the first run in weeks where Promote actually executed, after #1748 unblocked it. Each one was invisible until the one before it was fixed.

---

## 1. Promotion copies had no retry

`Manifest` ✅, `Sign` ✅ in 19s, then Promote:

```
level=fatal msg="copying image 2/2 from manifest list: writing blob: uploading layer to
https://ghcr.io/v2/tuna-os/elementary/blobs/upload/2.fbd191d4-...: blob upload unknown"
```

GHCR dropped the upload session on the **sixth** copy, after five had succeeded. Under `set -e` that leaves the flavor half-promoted — canonical tags written, one alias missing.

Promote is six to sixteen full blob-for-blob re-uploads back to back, and every one was a bare `skopeo copy`. The workflow already retries its two other network legs (base-image pull, GHCR push); this was the one bare path left. `copy_retry()` now wraps all six destinations — canonical, dated, per-arch, per-arch-dated, and both alias forms. `skopeo copy` to the same destination is idempotent, so a retry after a partial upload re-sends only the blob that failed.

Exhausting the budget stays a hard failure. A `copy_retry` that returned 0 after four failures would leave users pulling a bare `:latest` that silently still points at yesterday — worse than a red run.

## 2. No pytest test has run in CI since 2026-08-14

Not caused by this branch — `main` is red the same way, which is how I found it.

`tests/test_desktop_verify.py` (#1424, 08-13) and `tests/pytest/test_desktop_verify.py` (#1526, 08-14) are two independent suites for the same script that happen to share a filename. With no `__init__.py`, pytest names modules by basename, so the second import raises `import file mismatch` — a **collection** error, not a test failure. The run is interrupted and exits 2 before a single test executes.

Two consequences worth naming:

- Unit Tests has been red on `main` and on every PR for a day, so nothing merged in that window was actually tested.
- The coverage artifact silently never uploaded (`No files were found with the provided path: coverage-python.xml`), which is part of why the README reports low coverage.

`pytest.ini` sets `--import-mode=importlib`, which derives module names from the full path so no future pair can collide either. Collection goes from `367 items / 1 error` to **382 collected**. Deliberately *not* fixed by deleting one of the two files — they're different suites by different authors, and picking a winner is their call, not a side effect of unbreaking collection.

## 3. `attest_sbom` is off Promote's dependency path but not off its clock

#1560 split SBOM attestation out so a Rekor outage couldn't skip Promote. It still stalls everything downstream: `build-variant.yml`'s stage 2 waits for the whole reusable-workflow call, this job included.

On that run, `Sign` succeeded in **19 seconds** and `Attest SPDX SBOMs` was still running **57 minutes** later, with `pantheon` unable to start — for an artifact the job is explicitly allowed to fail to produce. (Sign passing while attest hangs is the same asymmetry seen on albacore in #1560: attest uploads a whole SBOM predicate, not just a signature. Attestation is systematically the fragile call, not randomly unlucky.)

Two bounds, because they bound different things:

| | value | governs |
|---|---|---|
| `SIGN_DEADLINE_MINUTES` | 40 → **10** | one cosign call |
| `timeout-minutes` | **20** (new) | the step |

The env deadline alone cannot bound the step: the loop runs `cosign attest` **plus** `cosign verify-attestation` per platform digest, and each call starts a fresh deadline — so the real worst case is `deadline × 2 × platforms`. That is why a "40 minute" budget was still running at 57. Image signing keeps its 40m: it gates promotion, so giving up early throws away a whole build.

## Tests

`tests/test_promote_retries_skopeo_copy.py` lifts `copy_retry` straight out of the workflow — exercising the shipped code, not a copy — and drives it against a fake `skopeo`: the gurnard error replayed and recovered; every attempt failing → non-zero exit, `::error::`, bare tag not claimed; a one-attempt budget not sleeping before it gives up; and static checks that no `skopeo copy` sits outside the wrapper. All five fail against `main`.

`tests/test_sign_step_retries_cosign.py`'s deadline test is split rather than loosened. The old flat `>= 30` couldn't express the distinction that matters, so it now pins three things separately: blocking cosign steps wait out a real outage, non-blocking ones give up sooner, and the non-blocking step carries a ceiling its per-call deadline cannot give it.

### DIFF
--- a/.github/workflows/reusable-build-image.yml
+++ b/.github/workflows/reusable-build-image.yml
@@ -1001,8 +1001,30 @@ jobs:
           path: supply-chain/sboms
 
       - name: Attest SPDX SBOMs
+        # A hard ceiling on the whole step, because SIGN_DEADLINE_MINUTES is a
+        # budget PER cosign INVOCATION, not per step: the loop below runs
+        # `cosign attest` + `cosign verify-attestation` for every platform
+        # digest, and each call starts its own deadline. Slow-but-eventually-
+        # succeeding calls therefore multiply -- on gurnard run 31891742138
+        # this step was still going 48 minutes into a nominal "40 minute"
+        # budget, and a three-platform variant could reach 40m x 6. The env
+        # deadline governs one call; this governs the step.
+        timeout-minutes: 20
         env:
-          SIGN_DEADLINE_MINUTES: "40"
+          # Deliberately much shorter than Sign's 40m. Splitting this job off
+          # (#1560) took SBOM attestation off Promote's critical path, but not
+          # off the *clock*: build-variant.yml's stage 2 waits for the whole
+          # reusable-workflow call to finish, and this job is part of it. On
+          # the run above `pantheon` had still not started 48 minutes after
+          # base's images were built, signed and manifested -- held up purely
+          # by an attestation the pipeline is explicitly allowed to do without.
+          #
+          # 10m is far longer than any successful attestation (image signing
+          # completed in 19s on that same run) so genuine flakes are absorbed;
+          # what it stops absorbing is a sustained outage, which is exactly the
+          # case where continuing without the attestation is the intended
+          # behaviour.
+          SIGN_DEADLINE_MINUTES: "10"
           SIGN_BACKOFF_CAP_SECONDS: "120"
         run: |
           set -euo pipefail
--- a/.github/workflows/reusable-build-image.yml
+++ b/.github/workflows/reusable-build-image.yml
@@ -1288,6 +1288,7 @@ jobs:
           MAJOR_VERSION: ${{ env.MAJOR_VERSION }}
           ALL_PLATFORMS_BUILT: ${{ needs.manifest.outputs.all_platforms_built }}
           INPUT_PLATFORMS: ${{ inputs.platforms }}
+          MAX_RETRIES: 4
         run: |
           set -ex
           # Promote the verified -testing manifest to the bare tags users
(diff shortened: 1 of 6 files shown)


hanthor:
## Correction to §3 of this PR's description

I wrote that `Attest SPDX SBOMs` on [run 31891742138](https://github.com/tuna-os/tunaOS/actions/runs/31891742138) "was still running **57 minutes** later", and used that as evidence that per-invocation deadlines were multiplying. That is wrong.

[The job log](https://github.com/tuna-os/tunaOS/actions/runs/31891742138/job/95033806873) shows 21 attempts, every one a `rekor.sigstore.dev` 502, ending at 16:27:58 with

```
::error::SIGSTORE_OUTAGE: attest still failing after 40m of Sigstore unavailability; giving up.
```

Job duration **39m19s** — one deadline, honoured exactly. `cosign-retry.sh` behaved as designed. My figures came from polling the jobs API, which kept reporting `in_progress` long after the job had finished; I should have read the log instead of trusting the status field.

#1768 corrects the two places that wording landed in the tree (a workflow comment and a test docstring). The `timeout-minutes: 20` ceiling stays, but as a structural bound rather than a fix for a measured overrun — under `set -e` a failing call aborts after one deadline, so multiplication needs calls that *succeed* slowly, which has not been observed.

**The 40m → 10m deadline change is unaffected and was measured correctly:** base's manifest was pushed at 15:47:58 and `pantheon` could not start until 16:27:59 — ~40 minutes of downstream stall for an artifact the job is explicitly allowed to fail to produce.

### Also worth recording: this run promoted

The same run produced the first fully green non-rolling cell:

| job | result |
|---|---|
| `pantheon / Gate` | ✅ |
| `pantheon / Sign` | ✅ |
| `pantheon / Promote` | ✅ 17:03:27 |
| `pantheon / Attest SBOM` | still running at the time |

Promote completed while attestation was still in flight — the #1560/#1748 split working exactly as intended. `base / Promote` failed on the `blob upload unknown` this PR fixes, and `pantheon` promoted anyway thanks to #1729.

---

Issue #1893, opened 2026-08-19
ISO build stalls in tacklebox's post-customize `podman commit` when the build runs rootless


hanthor:
## Symptom

After #1882/#1885 made the tacklebox `[customize]` phase stream and complete, the ISO build hangs **after** customize finishes: a long stretch of zero output, then a job-level cancellation with no actionable error.

## The stalling step is `podman commit`, not assembly

**This issue's two original suspects (xorriso/squashfs assembly, and a post-customize step in the `just iso` recipe) are both wrong.** Assembly never starts.

tacklebox now streams a marker at the commit, which names the step. From ISO E2E job [97638514487](https://github.com/tuna-os/tunaOS/actions/runs/32793052501/job/97638514487) (`skipjack:xfce`, `ubuntu-24.04`, `source: build`):

| time | event |
|---|---|
| 00:29:39 | `>>> [customize] (2/2) customize-live.sh` |
| 00:30:35 | `customize-live: done (desktop=xfce, installer=org.tunaos.InstallerXfce)` |
| 00:30:35 | `>>> [customize] committing tbox-customize-f946e9baece38773 -> localhost/tbox-live-custom:f946e9baece38773` |
| 00:40:35 | `Error: … commit customized localhost/skipjack:xfce: timeout --foreground 600 podman commit …: exit status 124` |

Customize itself takes 56 seconds. The commit runs for **exactly 600.0 seconds** and is killed. Exit 124 is `timeout`'s own kill status. Before that bound existed, the same stall read as ~50 minutes of dead air and a job cancellation — this issue's original report.

## It is not the runner. It is root vs rootless.

I said in an earlier revision of this comment that the commit does not complete on a GitHub-hosted runner. **That was wrong, and it pointed at an expensive non-fix (moving ISO builds to RunsOn).** The counter-evidence is in this repository's own CI.

installer-smoke run [32747410944](https://github.com/tuna-os/tunaOS/actions/runs/32747410944), every job on `ubuntu-latest`:

| job | step | result | duration |
|---|---|---|---|
| build yellowfin:gnome | Build dev ISO | **success** | 38m46s |
| build yellowfin:kde | Build dev ISO | **success** | 41m16s |
| yellowfin:gnome | boot + installer walkthrough | **success** | — |

Same tacklebox customize-and-commit, same runner image, twice, followed by a green end-to-end boot. The runner is not the variable.

The variable is which user tacklebox runs as:

[...]

Issue #1823, opened 2026-08-17
[bonito-rawhide] stage-2 builds fail with rpmdb "database disk image is malformed" on the fresh rawhide base


hanthor:
## What

Validation run [32011563812](https://github.com/tuna-os/tunaOS/actions/runs/32011563812) (2026-08-17, main with #1821): the base now survives the rpmfusion skew on **both arches** (the #1810 tolerance path fired exactly as designed — `::warning`, `--skip-broken`, "proceeding without: ffmpeg"), which unblocked stage-2 — and stage-2 immediately hit a **new** blocker.

Desktop builds fail in `20-packages.sh` with every RPM install erroring:

```
>>> [RPM] INSERT OR REPLACE INTO 'Packages' VALUES(?, ?): 11: database disk image is malformed
>>> [RPM] git-core-2.55.0-2.fc45.x86_64: install failed
...
Transaction failed: Rpm transaction failed.
```

It persists across the workflow's 4 dnf retries **and** all 3 full build attempts (`dnf clean metadata` between attempts doesn't help — the corruption is in the image's rpmdb, not the repo metadata).

## Distribution — size-correlated, not deterministic

| stage-2 job | result |
|---|---|
| base-hwe amd64 + arm64 (small overlay) | ✅ both passed |
| gnome amd64 + arm64 | ❌ both |
| niri amd64 + arm64 | ❌ both |
| cosmic arm64, xfce arm64, kde amd64, base-nvidia amd64 | ❌ |
| cosmic amd64, xfce amd64, kde arm64 | still running (later attempts) at time of writing |

Small stage-2 overlays succeed; large package transactions on the same base fail on both architectures. That pattern (plus sqlite error code 11) points at the rpmdb sqlite database in the freshly-built rawhide base being invalid — or becoming invalid when a large transaction rewrites it across the overlayfs boundary — rather than at any specific package.

## Why it was never seen before

Until today the bonito-rawhide **base itself** failed (the #1810 ffmpeg skew), so stage-2 never built on a fresh rawhide base — the nightly desktops either skipped or built from the last promoted base. #1821 fixed the base, and this is the next layer of the onion.

## Suspects worth checking

- rpm/sqlite version skew in rawhide (f45/f46 boundary visible in the log: base is rawhide, packages install as `.fc45`) — an rpmdb written by the base stage's rpm being invalid for the stage-2 rpm
[...]

PR #1953, opened 2026-08-21
Fetch and deploy the installer flatpak for the ISO's own architecture

hanthor:
The tunaOS half of tuna-os/bootc-installer#25. Paired with tuna-os/bootc-installer#26, which publishes the aarch64 bundle.

## The measurement

gurnard run [32495176056](https://github.com/tuna-os/tunaOS/actions/runs/32495176056), `iso:pantheon (linux-arm64)`:

```
+ flatpak build-import-bundle /tmp/installer-local-repo /tmp/bootc-installer.flatpak
Importing app/org.bootcinstaller.Installer/x86_64/master   ← on an aarch64 host
+ flatpak install --system --noninteractive installer-local org.bootcinstaller.Installer
error: Nothing matches org.bootcinstaller.Installer in remote installer-local
+ exit 1
```

A flatpak bundle carries refs for one architecture. The release asset is x86_64-only and its name carries no arch, so the arm64 ISO fetched a bundle it could not use. The same run's amd64 leg went the whole distance on the same commit — build → boot gate → sign → walkthrough → R2 upload — so this is the one thing standing between tunaOS and a green arm64 ISO.

That also means nothing downstream of the arm64 ISO has ever been exercisable: not the installer smoke gate, not the walkthrough.

## The change

Pick the asset from `$(uname -m)`, and keep x86_64 on the bare name:

| host | asset |
|---|---|
| x86_64 | `org.bootcinstaller.Installer.flatpak` |
| everything else | `org.bootcinstaller.Installer-$(uname -m).flatpak` |

Every existing consumer fetches that exact x86_64 URL, so the suffix is for the other arches only. The **fallback mirror uses the same resolved name** — fetching a bare name there would reintroduce the bug for anyone who reaches the fallback, and there's a test that both URLs agree.

When no asset exists for the arch it now fails naming the arch:

```
ERROR: no org.bootcinstaller.Installer-aarch64.flatpak published for aarch64
```

Without that, the next failure is `Nothing matches … in remote installer-local` three commands later, which reads as a corrupt download rather than an asset that was never published. That misreading is exactly why this took a whole ISO run to find.

## A second hardcoded arch in the same block

Reading that code turned up another one a few lines down:

```bash
_app_arch_dir="/var/lib/flatpak/app/${INSTALLER_APP}/x86_64"
```

flatpak deploys under the host arch, so on aarch64 that globbed a directory that does not exist and the active-symlink repair silently did nothing. Fixing only the download would have left the arm64 ISO booting to an installer flatpak cannot launch.

Both move to `$(uname -m)`, and `test_no_arch_is_hardcoded_anywhere_in_the_installer_block` sweeps the whole block for any literal that remains — a sweep rather than a spot check, because either hardcoding alone breaks the ISO. It strips comment lines first: the comments quote the incident, `x86_64/master` included, and scanning them would make the test fail on its own explanation.

## Ordering

This side is inert until #26 lands — no aarch64 asset exists yet. Until then arm64 fails with a message naming what is missing instead of one that looks like corruption, which is strictly better than today. x86_64 behaviour is byte-identical.

## Verification

- `bash -n` and `shellcheck --exclude=SC1091` clean (the repo's bats gate uses that exact invocation, so info-level findings count).
- 6 new tests, mutation-checked: reverting the download to the bare name fails `test_both_download_urls_use_the_resolved_asset`; re-hardcoding the deploy path fails `test_the_deploy_path_is_not_hardcoded_to_x86_64`.
- 77 assertions green across `test_customize_live.bats`, `test_live_installer_launch.bats` and `test_build_scripts.bats`, 0 failures.

---

### DIFF
--- a/live-iso/common/src/customize-live.sh
+++ b/live-iso/common/src/customize-live.sh
@@ -484,13 +484,36 @@ if [[ -n "${INSTALLER_APP}" ]]; then
 		# missing stamp. Fallback: tuna-os/tuna-installer, which mirrors
 		# the same app ID as a release asset.
 		INSTALLER_FLATPAK_FILE="/tmp/bootc-installer.flatpak"
+		# A flatpak bundle carries refs for ONE architecture, so the asset has
+		# to match the host. The release was x86_64-only and the name carried
+		# no arch, so an aarch64 ISO downloaded it, imported
+		# app/org.bootcinstaller.Installer/x86_64/master, and then:
+		#
+		#   error: Nothing matches org.bootcinstaller.Installer in remote
+		#          installer-local
+		#
+		# (gurnard run 32495176056, iso:pantheon linux-arm64.) x86_64 keeps
+		# the bare name — every existing consumer fetches that exact URL, and
+		# renaming it would break them for no gain. See
+		# tuna-os/bootc-installer#25.
+		_installer_asset="org.bootcinstaller.Installer.flatpak"
+		if [[ "$(uname -m)" != "x86_64" ]]; then
+			_installer_asset="org.bootcinstaller.Installer-$(uname -m).flatpak"
+		fi
 		if ! curl --retry 3 --fail --location --max-time 300 \
-			"https://github.com/tuna-os/bootc-installer/releases/latest/download/org.bootcinstaller.Installer.flatpak" \
+			"https://github.com/tuna-os/bootc-installer/releases/latest/download/${_installer_asset}" \
 			-o "${INSTALLER_FLATPAK_FILE}" 2>/dev/null; then
 			echo "tuna-os/bootc-installer unavailable, falling back to tuna-os/tuna-installer..."
-			curl --retry 3 --fail --location --max-time 300 \
-				"https://github.com/tuna-os/tuna-installer/releases/latest/download/org.bootcinstaller.Installer.flatpak" \
-				-o "${INSTALLER_FLATPAK_FILE}"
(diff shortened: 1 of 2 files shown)


hanthor:
Blocker report — not fixable in this PR, and I'm stopping the re-runs.

The `build` job failed twice on the same commit, on two different runners, identically:

```
error: SELECT hnum, blob FROM 'Packages': 11: database disk image is malformed
::error title=rpmdb data loss (tunaOS#1823)::the rebuilt rpmdb no longer provides
  system-release(releasever) — the salvage was lossy and dnf's $releasever
  detection is gone.
```

Two runners with the same corruption rules out an unlucky host, so re-running is not going to clear it. That's #1823, still open.

This PR cannot be the cause: it touches `live-iso/common/src/customize-live.sh` only, which runs at ISO-customize time inside tacklebox — nothing in it is reached during the image build's rpm transactions. The diff is byte-identical in behaviour on x86_64.

**One thing worth fixing while someone is in there.** The guard's own message sends the reader to the wrong issue:

> `fix the runner's container storage (see #1893)`

[#1893](https://github.com/tuna-os/tunaOS/issues/1893) is the amd64 ISO build hanging in tacklebox's post-customize assembly phase — unrelated. The correct reference is #1823, which the same message already names in its title. `build_scripts/lib.sh:813`.

I've left that alone rather than fold an unrelated one-line change into a PR about architecture handling.

---

PR #2010, opened 2026-08-24
Unblock hummingbird promotion (ext4 root), run the ISO gate that never ran, and capture why a session exits

hanthor:
## 0. The hummingbird desktops cannot be published at all, and this is why

Measured against the registry on 2026-08-25, not inferred:

```
ghcr.io/tuna-os/hummingbird:base    200   (dated tags through base-20260819)
ghcr.io/tuna-os/hummingbird:gnome   404
```

The full tag list carries `gnome-linux-amd64` and `gnome-testing` but **no promoted desktop tag for any flavor** — gnome, kde, cosmic and niri have never been published. `docs/MATRIX-STATUS.md` records the consequence one step further out: live-overlay tags are missing for `hummingbird-gnome`, `hummingbird-kde`, `hummingbird-cosmic` and `hummingbird-niri`, because `live-overlay.yml` builds each overlay **from the published flavor image**, and there isn&#39;t one. No overlay, no ISO.

The cause is a single abort in the Gate&#39;s `bootc install to-disk`, in last night&#39;s run ([32786338463](https://github.com/tuna-os/tunaOS/actions/runs/32786338463), gnome Gate job [97638679623](https://github.com/tuna-os/tunaOS/actions/runs/32786338463/job/97638679623)) and in the five nightlies before it:

```
Creating root filesystem (xfs) on device /dev/loop0p3
Bootloader: grub
Installing for i386-pc platform.
/usr/sbin/grub2-install: error: ../grub-core/kern/fs.c:123:unknown filesystem.
error: boot data installation failed: installing component BIOS to device /dev/loop0
error: Recipe `qcow2` failed with exit code 1
```

hummingbird probes `BACKEND=ostree SEALED=0`, so the qcow2 path does not pass `--filesystem ext4` (that is reserved for composefs-sealed images) and it lands on the `type = "xfs"` default in `00-tunaos.toml`. BIOS `grub2-install` cannot read the xfs this base&#39;s `mkfs.xfs` writes, and `bootupctl` installs the BIOS component unconditionally on a hybrid layout — so the failure aborts the **whole** install, not just its legacy half. The Gate then produces no serial log and no screenshots, Promote never runs, and nothing is published.

`system_files_overrides/hummingbird/usr/lib/bootc/install/50-hummingbird.toml` sets `type = "ext4"` for hummingbird only, which is exactly the mechanism `00-tunaos.toml` prescribes for a per-variant override (&#34;a higher-sorting drop-in … so they win over this file&#34;).

**Status of that claim: being measured now, not proven.** The run that was supposed to prove it ([32809685474](https://github.com/tuna-os/tunaOS/actions/runs/32809685474)) was **cancelled** mid-`Build Image` — this workflow&#39;s concurrency group is keyed on `github.ref` with `cancel-in-progress: true`, so later pushes to this branch evicted it before it ever reached the Gate. A clean `flavor=gnome` dispatch is in flight and this section will be updated with what the Gate actually says.

Scope, checked rather than assumed: this fixes `bootc install to-disk`, which is the path the Gate takes. It is **not** what makes an ISO install to a laptop work — that path runs fisherman, which always lays down a 3-partition GPT with a separate ext4 `/boot` precisely because GRUB cannot read modern XFS, so it never touches the xfs root. What this unblocks is **publishing**, which everything downstream is waiting on.

---

Run [32681262659](https://github.com/tuna-os/tunaOS/actions/runs/32681262659) is the **first time the gnome smoke cell has ever run**. It answered three separate questions.

## 1. A gate that examined nothing

`e2e-installer-gui-checks.sh` resolved its TAP helpers as `${TEST_LIB_DIR}/lib/e2e-assert.sh` — `/lib` **after** the expansion instead of inside the default. `iso-e2e.sh` uploads the helper flat beside the script and sets `TEST_LIB_DIR` to the guest home, so that names a directory the guest does not have:

```
/home/liveuser/e2e-installer-gui-checks.sh: line 25:
  /home/liveuser/lib/e2e-assert.sh: No such file or directory
... line 54: check: command not found
... line 159: print_summary: command not found
```

`source` failed, `check()` was never defined, every assertion evaporated onto stderr, and bash exited 127. The harness printed that as:

```
::warning::installer GUI checks reported 127 failure(s) for gnome
```

**127 is bash for command-not-found.** It is not a count of anything, and it was a *warning*, in a mode that tolerates warnings. The per-flavor compositor and installer-frontend assertions have never executed, on any flavor, in any smoke run — and the workflow stayed green through it.

Both halves are fixed:

- the path matches where the file is uploaded, and all three sibling check scripts now bail out loudly on an unreadable helper rather than silently disarming every assertion;
- the harness distinguishes *did not run* from *found problems*. The TAP summary is the discriminator — `# Results:` comes only from `print_summary`, the last statement of every check script — so its absence is a hard failure in every mode. `E2E_*_STRICT` governs failed assertions, not an absent gate.

## 2. The disk screen gnome reaches and the contract cannot see

```
not ok - gnome: reached &#39;disk&#39; screen (Disk / target selection)
  # not found on any state the installer advanced to
```

…while crediting encryption and summary, both of which sit *after* disk selection. `tests/installer-screens.yaml` was measured on 2026-08-07 against KDE, Niri, XFCE and COSMIC; gnome was absent from the matrix until #1039, so upstream `bootc-installer` was the one fork nobody OCRed. Its heading is **&#34;Install Location&#34;**, matching none of the six keywords — the same shape as the `install` list before it. Checked against all five frontends before admitting it: the string occurs exactly once in the whole set, and carries no product name.

## 3. Why a compositor exits, which nothing recorded

greetd reports only its own bookkeeping — `greeter exited without creating a session`, five times to `start-limit-hit` — and the compositor&#39;s reason went nowhere. That message is the entire evidence base on which four desktops have been written off as needing a DRM render node, and **this run disproves the attribution**. Read from inside the guest:

```
crw-rw-rw-. 1 root render 226, 128 renderD128
[drm] pci: virtio-vga detected at 0000:00:02.0
[drm] features: -virgl +edid -resource_blob -host_visible
```

The node is present; 3D is not. Those are different claims. The `eglinfo` output cited for kde was collected on the runner **host**, which says nothing about the guest&#39;s environment. And gnome starts on that same hardware through Mesa&#39;s software path, so a compositor *can* come up here.

`live-iso/common/src/desktop-{xfce,niri,cosmic}.sh` now run the session under `systemd-cat -t tunaos-live-session`, and the workflow prints that tag — so the next run says why instead of inviting a sixth guess. This is the same remedy that ended four wrong guesses at `gnome-desktop:languages` in the package factory: make the failing thing print its own log, then read it. An image predating the change says so rather than printing nothing.

## Docs

`docs/MATRIX-STATUS.md` and the generator now state the symptom and point at the evidence instead of restating a cause the measurement did not support: those cells are **undiagnosed**, not hardware-limited. gnome is recorded as demonstrated rather than assumed — it boots, both processes come up, and the walkthrough captures 9 non-blank frames across 8 visual states.

Two real defects that run surfaced are recorded there too. The second is fixed here; the first is **not**, and is left as a correctly-failing gate rather than a guessed fix:

- **The installer starts behind the GNOME Activities overview.** The walkthrough must press `esc` before it can drive anything, and records it as a failure precisely because a user booting this ISO sees the same thing. Reproduced on `32450214451` and `32681262659`.

## Testing

19 new tests, **every one mutation-verified**, including reintroducing the original helper path and unwrapping a single one of niri&#39;s four session commands. The check scripts are tested by building the guest&#39;s layout on disk and *running* them — reading the line is how a shell-expansion bug survives review in the first place.

730 passed.

— hive: backend=pi model=openai-codex/gpt-5.6-sol pi=0.84.1

### DIFF
--- a/.github/build-config.yml
+++ b/.github/build-config.yml
@@ -20,16 +20,34 @@ iso_groups:
   - suffix: ""
     flavors: [gnome-nvidia, gnome-nvidia-hwe]
     offline_flavors: [gnome, gnome-hwe, gnome-nvidia, gnome-nvidia-hwe]
-  # Additional desktop ISO group (KDE/COSMIC/Niri/XFCE). NOT published as a
-  # single grouped ISO — the browser ISO Builder (https://iso.tunaos.org) covers these on
-  # demand from the published images, so we don't carry the cost of a
-  # 5-flavor offline-store ISO for the long tail. The images still build
-  # and publish to GHCR; only the grouped ISO is skipped.
-  # (Manually publishable via workflow_dispatch group=community if needed.)
-  - suffix: community
-    publish: false
-    flavors: [kde-nvidia, cosmic-nvidia, niri-nvidia, xfce-nvidia]
-    offline_flavors: [kde, kde-hwe, cosmic, cosmic-hwe, niri, niri-hwe, xfce, kde-nvidia, cosmic-nvidia, niri-nvidia, xfce-nvidia]
+  # ONE ISO PER BASE+DESKTOP, each embedding its own NVIDIA and HWE flavors —
+  # the same shape as the GNOME group above, applied to the other four
+  # desktops. Published as `<variant>-<suffix>.iso`.
+  #
+  # This replaces a single `community` group that packed KDE, COSMIC, Niri and
+  # XFCE into ONE ISO and carried `publish: false`, on the reasoning that the
+  # browser ISO Builder covers the long tail on demand so a five-flavor
+  # offline store was not worth the storage. The cost argument was sound; the
+  # conclusion was not. A user who wants KDE should get a KDE ISO, not a
+  # four-desktop ISO or a browser build step, and one desktop's offline store
+  # is a fraction of four.
+  #
+  # Storage does not grow the way the old comment implied either: these are
+  # DEDUPLICATED groups over a shared squashfs, so a per-desktop ISO carries
+  # its own desktop plus the base layers it already shares with every other
+  # ISO of the same variant.
+  - suffix: kde
+    flavors: [kde-nvidia]
+    offline_flavors: [kde, kde-hwe, kde-nvidia]
+  - suffix: cosmic
+    flavors: [cosmic-nvidia]
+    offline_flavors: [cosmic, cosmic-hwe, cosmic-nvidia]
+  - suffix: niri
+    flavors: [niri-nvidia]
+    offline_flavors: [niri, niri-hwe, niri-nvidia]
+  - suffix: xfce
+    flavors: [xfce-nvidia]
+    offline_flavors: [xfce, xfce-hwe, xfce-nvidia]
 
 # Build stages (for reference — enforced by the workflow job graph):
 #
--- a/.github/build-config.yml
+++ b/.github/build-config.yml
@@ -739,7 +739,12 @@ variants:
   # (Konflux hermetic build model, CKI ARK kernel).
   - id: hummingbird
     emoji: "🐦"
-    description: "Fedora Hummingbird (bootc container-native OS)"
+    # Rolling, security-hardened fork tracking Fedora RAWHIDE -- not Fedora 43,
+    # whatever the .fc43 dist tags suggest -- and shipping NO desktop upstream.
+    # The desktop flavors below are therefore a PORT onto a desktop-less base,
+    # not a rebuild of an existing one. Read docs/HUMMINGBIRD.md before
+    # reasoning about what its package set should contain.
+    description: "Fedora Hummingbird (hardened rolling Rawhide fork, bootc)"
     base_image: "quay.io/hummingbird-community/bootc-os:latest@sha256:c5539f9ed4d93aab6bd41e4f5aef8ab83055f3f9e855a47b69fadb7420d0d1df"
     platforms: ["linux/amd64", "linux/arm64"]
     flavors:
--- a/.github/workflows/gdm-paint-benchmark.yml
+++ b/.github/workflows/gdm-paint-benchmark.yml
@@ -59,6 +59,8 @@ jobs:
         with:
           name: gdm-bench-base-qcow2
           path: base.qcow2
+          # Already-compressed disk image: level 6 spends minutes for ~0%.
+          compression-level: 0
           retention-days: 1
 
   benchmark:
--- a/.github/workflows/installer-screenshots.yml
+++ b/.github/workflows/installer-screenshots.yml
@@ -154,7 +154,18 @@ jobs:
           GITHUB_REPOSITORY_OWNER: tuna-os
           TACKLEBOX_FROM_SOURCE: "1"
         run: |
-          just iso "${VARIANT}" "${{ matrix.flavor }}" ghcr "${{ matrix.flavor }}"
+          # Root, not bare. tacklebox's post-customize `podman commit` is
+          # bounded at a hardcoded 600s, and how long it takes is a product
+          # of the storage path and the disk. Measured on the same commit:
+          #
+          #   ubuntu-24.04  + rootless (fuse-overlayfs)   >600s, killed (x3)
+          #   ubuntu-24.04  + root     (native overlay)    133s
+          #   RunsOn        + rootless (fuse-overlayfs)    150s
+          #
+          # This job is on a GitHub-hosted runner, which is the slow disk, so
+          # it needs the fast storage path. `sudo -E env "PATH=$PATH"` is the
+          # form installer-smoke.yml uses and the one measured green there.
+          sudo -E env "PATH=$PATH" just iso "${VARIANT}" "${{ matrix.flavor }}" ghcr "${{ matrix.flavor }}"
 
       - name: Run installer walkthrough
         timeout-minutes: 30
--- a/.github/workflows/installer-screenshots.yml
+++ b/.github/workflows/installer-screenshots.yml
@@ -91,6 +91,30 @@ jobs:
             tesseract-ocr imagemagick
           sudo wget -qO /usr/bin/yq https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64 && sudo chmod +x /usr/bin/yq
 
+      # The runner image ships podman 5.8.4, which wedges committing a
+      # live-customized rootfs (#1893). Every ISO build needs the Kubic one
+      # instead; this job ran `just iso` without it. Necessary, not
+      # sufficient -- iso-e2e.yml installs this and its commit still exceeded
+      # tacklebox's 600s bound on a GitHub-hosted runner (job 97638514487) --
+      # but a build without it cannot finish at all.
+      - name: setup working podman
+        run: |
+          set -euo pipefail
+          sudo install -d /etc/apt/keyrings
+          curl -fsSL https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/unstable/xUbuntu_24.04/Release.key \
+            | gpg --dearmor | sudo tee /etc/apt/keyrings/kubic.gpg >/dev/null
+          echo 'deb [signed-by=/etc/apt/keyrings/kubic.gpg] https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/unstable/xUbuntu_24.04/ /' \
+            | sudo tee /etc/apt/sources.list.d/kubic.list
+          sudo apt-get update -qq
+          sudo apt-get remove -y podman conmon crun || true
+          sudo apt-get install -y podman
+          for p in /usr/local/bin/podman /usr/local/sbin/podman; do
+            if [ -e "$p" ] && ! dpkg -S "$p" >/dev/null 2>&1; then
+              sudo rm -f "$p"
+            fi
+          done
+          hash -r
+
       - name: Log in to GHCR
         run: echo "${{ env.GITHUB_TOKEN }}" | podman login ghcr.io -u ${{ github.actor }} --password-stdin
         env:
--- a/.github/workflows/installer-screenshots.yml
+++ b/.github/workflows/installer-screenshots.yml
@@ -279,6 +279,8 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: write
+      # Needed to open the refresh PR; `main` only accepts merge-queue changes.
+      pull-requests: write
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
@@ -331,9 +333,20 @@ jobs:
         if: steps.stage.outputs.staged != '0'
         run: bash scripts/gen-screenshot-gallery.sh
 
-      - name: Commit and push
+      # Opens a PR; does NOT push to main. `main` is protected by a
+      # repository rule -- "Changes must be made through the merge queue" --
+      # so `git push origin HEAD:main` is rejected outright (GH013). Observed
+      # on installer-smoke.yml run 32704425971, where the identical step had
+      # simply never executed before: it is gated on `staged != 0` and no
+      # flavor had ever produced frames. All three screenshot writers carried
+      # the same line; all three are now the matrix-status.yml shape.
+      - name: Open the screenshot refresh PR
         if: steps.stage.outputs.staged != '0'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          BRANCH: automation/installer-screenshots-weekly
         run: |
+          set -euo pipefail
           git config user.name "github-actions[bot]"
           git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
           git add docs/images/installer docs/SCREENSHOTS.md
@@ -341,6 +354,25 @@ jobs:
             echo "No screenshot changes to commit"
             exit 0
           fi
-          git commit -m "docs: refresh installer walkthrough screenshots ($(date +%Y-%m-%d)) [skip ci]"
-          git pull --rebase origin main
-          git push origin HEAD:main
+          git checkout -b "$BRANCH"
+          git commit -m "docs: refresh installer walkthrough screenshots ($(date -u +%Y-%m-%d))"
+          git push --force origin "$BRANCH"
+
+          if [[ -z "$(gh pr list --head "$BRANCH" --state open --json number --jq '.[].number')" ]]; then
+            gh pr create --base main --head "$BRANCH" \
+              --title "docs: refresh installer walkthrough screenshots" \
+              --body "$(printf '%s\n\n%s\n' \
+                'Captured by the weekly installer walkthrough and staged by `scripts/gen-screenshot-gallery.sh`.' \
+                'Force-updated in place, so this branch always holds the newest capture.')"
+          else
+            echo "PR already open for $BRANCH — force-push refreshed it."
+          fi
+
+          # A PR opened with the Actions token starts no checks of its own;
+          # the merge queue runs the required checks on the merge group, so
+          # arming auto-merge is what actually lands the refresh.
+          pr=$(gh pr list --head "$BRANCH" --state open --json number --jq '.[0].number')
+          if [[ -n "$pr" ]]; then
+            gh pr merge --auto --merge "$pr" ||
+              echo "::warning::could not arm auto-merge on #$pr — the refresh will wait for a human"
+          fi
--- a/.github/workflows/installer-smoke.yml
+++ b/.github/workflows/installer-smoke.yml
@@ -20,24 +20,29 @@ on:
         default: "yellowfin"
         type: string
       build_runner:
-        # The experiment behind tunaOS#2010: is RunsOn actually required to
-        # build an ISO, or was it inherited from a podman diagnosis that the
-        # workflow itself contradicts three steps later?
+        # MEASURED 2026-08-24, run 32747410944 against RunsOn runs
+        # 32747410944/32735883406/32718219267, same commit, same steps:
         #
-        # `runs-on` (default) keeps every scheduled run and every other
-        # dispatch exactly as it was. `github` routes the build to
-        # ubuntu-latest so the two can be compared on the same commit.
+        #   Build dev ISO      ubuntu-latest      RunsOn (c6a, 8 vCPU)
+        #   gnome              38m46s             48m18s
+        #   kde                41m16s             59m53s
         #
-        # The `schedule` trigger passes no inputs, so this evaluates empty
-        # there and falls through to RunsOn -- the weekly run cannot be
-        # changed by adding this.
+        # GitHub-hosted is FASTER on both flavors, on half the cores, and the
+        # gnome leg went green end to end (walkthrough included). No disk
+        # failure: `free disk space` frees enough, taking 1m36s-5m25s to do
+        # it against ~11s on RunsOn, which is the whole of the setup cost.
+        # The build is evidently not CPU-bound.
+        #
+        # So `github` is now the default and `runs-on` is the escape hatch,
+        # kept because one afternoon of measurements is not a guarantee and
+        # reverting should not need a code change.
         description: "Runner for the ISO build"
         required: false
-        default: "runs-on"
+        default: "github"
         type: choice
         options:
-          - runs-on
           - github
+          - runs-on
       flavors:
         description: "Comma-separated flavors"
         required: false
@@ -118,24 +123,28 @@ jobs:
     # The boot Gate in reusable-build-image.yml is unaffected: it genuinely
     # runs iso-e2e-gpu.sh and still asks for `gpu`.
     #
-    # RUNNER, and the experiment: `build_runner: github` sends this job to
-    # ubuntu-latest instead. The stated reason for RunsOn is directly above --
-    # "cannot work on the GitHub runner's podman 5.8.4" -- but the `setup
-    # working podman` step below opens by saying the RunsOn image ships THE
-    # SAME podman 5.8.4 and installs the Kubic build over it. That step is a
-    # plain apt install with nothing runner-specific in it, so it should work
-    # identically on either. What is genuinely different is disk (80GB gp3 vs
-    # ~14GB free of ~65GB, which is why `free disk space` runs below) and
-    # cores (8 vs 4, against a 75-minute cap that a ~50-minute build already
-    # eats most of). Those are measurable, and this input is how.
-    runs-on: ${{ inputs.build_runner == 'github'
+    # RUNNER. The comment above is the history; the measurement retired it.
+    # The stated reason for RunsOn was "cannot work on the GitHub runner's
+    # podman 5.8.4" -- but the `setup working podman` step below always said
+    # the RunsOn image ships THE SAME podman 5.8.4 and installs the Kubic
+    # build over it, and that step is a plain apt install with nothing
+    # runner-specific in it. It succeeds on ubuntu-latest, as it always would
+    # have. tunaOS#1893's root cause was a `podman commit --quiet` wedge
+    # inside tacklebox, fixed in tacklebox#231 -- never a runner property.
+    #
+    # ubuntu-latest is the default now, on the numbers in the input comment
+    # above. `build_runner: runs-on` still routes back to RunsOn without a
+    # code change; the `schedule` trigger passes no inputs, so it takes the
+    # empty branch, which is why the comparison is written `!= 'runs-on'`
+    # rather than `== 'github'`.
+    runs-on: ${{ inputs.build_runner != 'runs-on'
       && 'ubuntu-latest'
       || format('runs-on={0}/runner={1}', github.run_id, matrix.runner) }}
-    # Raised from 75 only for the GitHub leg: half the cores on the same work
-    # would otherwise be cancelled at the cap and read as "GitHub runners
-    # cannot build an ISO", when the finding would really be "it needs longer
-    # than 75 minutes". A timeout is not a measurement.
-    timeout-minutes: ${{ inputs.build_runner == 'github' && 150 || 75 }}
+    # 150 for ubuntu-latest, where the measured builds are 38-41 minutes: a
+    # cap close to the observed time turns a slow build into a cancellation
+    # and reports the wrong thing. 75 stays for the RunsOn path it was set
+    # for. A timeout is not a measurement.
+    timeout-minutes: ${{ inputs.build_runner != 'runs-on' && 150 || 75 }}
     strategy:
       fail-fast: false
       matrix: ${{ fromJson(needs.generate-matrix.outputs.matrix) }}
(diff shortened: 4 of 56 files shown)


hanthor:
## `E2E yellowfin:gnome` — pre-existing on `main`, not caused by this PR, and not fixed here

The check fails at step 7, **before the harness this PR touches is ever invoked**:

```
==> Fetching yellowfin-gnome-latest.iso from R2...
INFO  : There was nothing to transfer
::error::No ISO found in R2 at live-isos/yellowfin-gnome-latest.iso
```

`iso-e2e.yml` triggers on PRs that touch `scripts/iso-e2e.sh`, which this PR does — but its own header says what it actually tests:

> the workflow currently downloads from R2, which has the published main-branch ISO, not the PR's

So it boots the last **published** ISO, not this branch's code. There is no published `yellowfin-gnome-latest.iso` to boot.

**Reproduces on `main` with no PR involved.** The workflow's weekly `schedule` run is the clean read, and the most recent one — [32003299083](https://github.com/tuna-os/tunaOS/actions/runs/32003299083), `main`, 2026-08-17 — is a failure that lasted 71 seconds, consistent with dying at the same download. The scheduled runs on 2026-08-10 and earlier passed, so the published object went away sometime in that week. That is a week before this branch existed.

### Why I am not fixing it here

The real remedy is for `publish-isos.yml` to publish an ISO again, which is the ISO axis itself — the thing this PR is working toward, not something it can do.

The tempting alternative is to convert "object missing in R2" into a **skip**, alongside the existing "no R2 credentials → report skipped" path. I deliberately have not, because this PR exists to fix a gate that reported success while examining nothing, and turning a red check green by teaching it to skip is the same move one level up. If that skip is wanted it should be its own change, with the skip loud and narrowly conditioned on the object being absent, so a genuine boot regression still fails.

### What is green

The change in this PR is exercised by the repo's own suite — 730 passing, 19 of them new and every one mutation-verified, including reintroducing the original broken helper path. Also pushed `6c75416`, which addresses the CodeQL implicit-concatenation comment on `scripts/gen-matrix-status.py`; the string is byte-identical, verified by comparison rather than by inspection.

---

hanthor:
## `build` (Live ISOs) — fixed in `715b59d8`, and it is a real find

Separate failure from the R2 one above, and this one **is** actionable.

`Sync image into rootless storage` runs `sudo podman save ... | podman load`. The second half is rootless, and on **Ubuntu 24.04** — the distro this very workflow pins podman against through the kubic repo — an unprivileged `clone(CLONE_NEWUSER)` is refused by default:

```
kernel.apparmor_restrict_unprivileged_userns = 1
```

podman reports that without naming any of it:

```
cannot clone: Permission denied
Error: cannot re-exec process
##[error]Process completed with exit code 125
```

Nothing there mentions namespaces, AppArmor, or a sysctl — and it arrives **after a ~12 minute image build**, because the first rootless command in the job is the one immediately following `Build OS image`.

### Not a regression from this branch

On run `32534198668` (2026-08-21) the same job died *earlier*, at `Build OS image`. Step 8 had simply never been reachable. It is reachable now, and this is what it hits — so the branch got further, it did not break something.

### The part worth more than the sysctl

The subuid/subgid step already existed to make rootless podman work. It set a precondition and **never verified the thing it was a precondition for** — the same shape as the `e2e-installer-gui-checks.sh` bug this PR is about: configuration present, behaviour absent, nobody looking, green.

So the step now proves it, using the cheapest possible instance of the exact operation everything downstream needs (`podman unshare true`, a fraction of a second). On failure it prints the subuid/subgid entries and all three namespace gates, and states which later steps will fail and with what uninformative message. It also prints the gate values **as found, before changing anything**, so a run records what the runner image shipped with rather than only what we did to it.

### Testing

Four tests. Two execute the step body with `sudo`/`podman`/`sysctl` shimmed — one proving it fails *and names the gate* when the namespace is refused, one proving it can still pass, because a check that cannot succeed is as useless as one that cannot fail. Removing the probe fails the first; that is the mutation that matters.

733 passed.

`E2E yellowfin:gnome` remains red for the unrelated, pre-existing R2 reason in my earlier comment.

---

hanthor:
## xfwl4's own words — and a correction to my earlier claim

### The compositor answered

Smoke run [32691426582](https://github.com/tuna-os/tunaOS/actions/runs/32691426582), dispatched on this branch so the dev ISO carries the session-logging change:

```
INFO  xfwl4: Starting xfwl4 on a tty using udev
INFO  xfwl4::backend::udev: Using renderD128 as primary GPU
INFO  smithay::wayland::socket: Created new socket name=Some("wayland-1")
INFO  xfwl4::core::state: Listening on wayland socket name="wayland-1"
WARN  smithay::backend::drm::device::fd: Unable to become drm master, assuming unprivileged mode
INFO  smithay::backend::drm::device: DrmDevice initializing
INFO  smithay::backend::egl::display: Successfully selected EGL platform: PLATFORM_GBM_KHR
INFO  smithay::backend::egl::display: EGL Initialized
INFO  smithay::backend::egl::display: EGL Version: (1, 5)
WARN  xfwl4::backend::udev::device: failed to initialize gpu err=NoRenderNode
ERROR xfwl4: Failed to initialize primary GPU node
```

The compositor **starts**. It **finds the render node and names it**. It opens a Wayland socket and listens. It initialises EGL on GBM. Then it dies at `err=NoRenderNode` — for a device it had already selected — having first failed to become DRM master.

So "cosmic/niri/xfwl4/kde need a DRM render node" is not merely unproven; it is contradicted by the compositor's own log. The open question is now specific and much smaller: **why Smithay reports `NoRenderNode` for a device it just picked, and whether the preceding drm-master failure causes that or is a separate symptom.** I have deliberately not guessed — guessing is what the previous five attempts did, and this is the first round with actual evidence.

A second defect sits in the same log, and it is packaging rather than graphics:

```
Failed to reload defaults: /usr/local/share/xfce4/xfwl4/defaults does not exist
WARN xfwl4::core::util::rc: Missing value for required setting box_move
... ~45 more required settings missing ...
```

`/usr/local/share` is not where a distribution package installs data. Recorded in the docs; not the fatal error.

### Correction: the userns claim was stated more confidently than the evidence supported

In [my earlier comment](https://github.com/tuna-os/tunaOS/pull/2010#issuecomment-5390722616) I said `kernel.apparmor_restrict_unprivileged_userns = 1` was the gate. **I could not verify that, and I should not have asserted it.**

What is solid: on run `32690010127` steps 6–9 all passed, and step 8 — `Sync image into rootless storage`, previously `exit 125` — succeeded, as did `Build Live ISO`. What is *not* verified is which gate was closed, because `Build OS image` emits tens of thousands of lines and **this job's log download comes back starting part-way through that step** — everything steps 1–6 printed is absent. The "as found" values I added specifically to answer this were captured and then unreadable.

That is the same defect as not capturing them, one level along, so `77342f23` also writes the gate values to `$GITHUB_STEP_SUMMARY`, which has its own budget and is not subject to that truncation. The next run of this workflow will say which gate it found, and I will confirm or retract accordingly.

(The suite caught a bug in that change before CI did: under `set -u`, an unset `GITHUB_STEP_SUMMARY` killed the step *after* the probe had already passed. Guarded, and there is now a test for it.)

### Unrelated: this run's `build` failure is not the podman path

It failed at step 12, `Boot ISO and capture desktop screenshot`:

```
Could not access KVM kernel module: No such file or directory
qemu-system-x86_64: failed to initialize kvm: No such file or directory
```

on a `c5.2xlarge` **spot** instance. The previous run was `c6a.2xlarge` on-demand. The ISO built fine; the boot verification needs `/dev/kvm` and this instance type did not provide one. Instance-type/capacity, not this diff — flagging rather than fixing, since runner sizing is not mine to choose.

---

hanthor:
## `build` — removed a gate that could never have run (`11f585bd`)

Per the maintainer's call, the boot/screenshot steps are gone from `live-iso-bootc.yml` rather than repaired.

### It was structurally impossible, not flaky

```
==> Booting ./skipjack-gnome-10-x86_64.iso in QEMU (headless, KVM)...
Could not access KVM kernel module: No such file or directory
qemu-system-x86_64: failed to initialize kvm: No such file or directory
```

Runs [32690010127](https://github.com/tuna-os/tunaOS/actions/runs/32690010127) and [32696543980](https://github.com/tuna-os/tunaOS/actions/runs/32696543980), **both after the ISO built successfully.**

The job runs on `runs-on=.../runner=build-amd64`, whose families in `.github/runs-on.yml` are `["c6a", "c6i", "c5a", "c5"]` — ordinary EC2 instances. EC2 exposes `/dev/kvm` only on `.metal` types. The step could not have passed on any run, on any commit, ever. No change in this repository alters that, so retrying or resizing was never going to help.

### The verification is not lost

`installer-smoke.yml` already does it properly: boots the ISO on `ubuntu-latest` (which does provide KVM), SSHes into the guest, asserts the compositor and installer frontend are running, drives the installer through its screens, and captures the session journal — across **all five flavors** rather than this job's single hardcoded one. It did exactly that tonight for gnome and for xfce, and the xfce leg is how xfwl4's exit was finally diagnosed.

Keeping a gate here that never once executed would report a verification nobody performed — the failure mode this branch exists to remove.

### The rule, not just the removal

The next person adding a QEMU step picks a runner the same way, so the invariant is now tested: **every job that reaches for KVM must run on a GitHub-hosted `ubuntu-*` runner.** Four do — `iso-e2e`, `luks-e2e`, and the two weekly screenshot workflows — and all four were already correct.

A fourth test pins the premise: if a `.metal` family is ever added to `runs-on.yml`, it fails, so the blanket rule gets revisited deliberately rather than silently over-applying.

The emptiness guard earned its keep immediately — my first runs-on parser read `cfg.values()` instead of descending into `runners:`, found no families, and would have passed while checking nothing. Same shape as the `lstrip` bug and the missing helper.

739 passed.

### Still outstanding

- **`E2E yellowfin:gnome`** — red for the pre-existing R2 reason (no published `yellowfin-gnome-latest.iso`; reproduces on `main`'s 2026-08-17 scheduled run). Unchanged, already explained above.
- **The userns gate value stays unverified.** Four separate mechanisms blocked reading it: in-progress logs 404; the completed job's log download begins part-way *through* `Build OS image`, so steps 1–6 are absent entirely; and `check_run.output` is empty even on a completed job. The step summary is written and **a human can read it on the run's summary page** — the limitation is the API I have, not the mechanism. My retraction in [5391418089](https://github.com/tuna-os/tunaOS/pull/2010#issuecomment-5391418089) stands: step 8 went from failing to passing across that change, and I cannot say from evidence which gate was closed.

---

hanthor:
## Withdrawing the seat claim — I measured the wreckage and described it as the crash

In an earlier comment I said **"EVERY logind session has an empty Seat and empty TTY"** and offered it as the unifying explanation for four red desktops. Run [32700275124](https://github.com/tuna-os/tunaOS/actions/runs/32700275124) shows that reading was wrong, and `8458b1de` withdraws it.

Here is what those sessions actually are:

```
session 59: VTNr=0 Seat= Type=unspecified Class=manager Active=yes
session 75: VTNr=0 Seat= Remote=yes       Type=tty      Class=user Active=yes
```

`Class=manager` is the user manager. `Remote=yes Type=tty` is **the harness's own SSH login.** Neither has any business having a seat — and the compositor's session is not in the list at all, because `loginctl` runs minutes after the compositor died and greetd hit `start-limit-hit`.

`seat0` also exists (`loginctl list-seats` reports it, `/run/systemd/seats/seat0` is on disk), which is close to the opposite of what I said.

### What the same run does establish

The greetd journal, captured for the first time, shows the shape plainly:

```
08:01:20 pam_unix(greetd:session): session opened for user liveuser
08:01:22 pam_unix(greetd-greeter:session): session opened ... session closed
08:01:22 error: check_children: greeter exited without creating a session
... restart, greeter-only, exits instantly, ×4 -> start-limit-hit
```

The **desktop** session opened once — that is the xfwl4 run that died at `NoRenderNode`. Every cycle after it is the **greeter**, which is gtkgreet under cage: another wlroots compositor, failing the same way. Two different compositors, one symptom, `Unable to become drm master` in front of both.

The PAM candidate from my previous commit is not supported either. `/etc/pam.d/greetd` carries `session include system-auth` — which on EL is where `pam_systemd.so` lives — and `/etc/pam.d/greetd-greeter` names `session optional pam_systemd.so` outright.

### The fix is to the instrument, not to a theory

Both readings failed for the same reason: taken after the thing being measured had exited. `systemd-logind`'s journal is the one source without that problem — it records `New session N of user liveuser` **with seat and VT at creation time**, and that entry outlives the session.

So this commit adds `-u systemd-logind` to the capture and commits **no new theory on purpose**. The next run says whether the desktop session was ever seated; until it does, I am not naming a cause. This axis has now cost five wrong guesses and one wrong measurement, and the difference between those two mistakes is worth keeping straight.

Mutation-verified: dropping `systemd-logind` fails the test, whose message says why post-mortem `loginctl` cannot answer this. 744 passed.

### Still standing

`renderD128` is present and world-readable in the guest, and xfwl4 selects it **by name** before dying for want of it. Whatever the cause turns out to be, "these desktops need a DRM render node" remains contradicted by the compositor's own log.

---

hanthor:
## The `iso` criterion is satisfiable. Two cells now demonstrate it.

`docs/matrix-provenance.json` had `iso` at 0-for-66 — 66 cells carrying an iso verdict, none passing. Two now pass, by the two different routes the harness supports:

| cell | source | evidence |
|---|---|---|
| **skipjack:kde** | `build` — built on the runner | [job 97648442043](https://github.com/tuna-os/tunaOS/actions/runs/32796371936/job/97648442043): build 19m54s, customize commit 133s, 5.1 GB ISO, boot to `TUNAOS_LIVE_READY` in 10m23s |
| **albacore:gnome** | `r2-latest` — the published ISO | [job 97648434657](https://github.com/tuna-os/tunaOS/actions/runs/32796370587/job/97648434657): 5.7 GB downloaded in 1m50s, booted in 69s |

The second is the release check — that is the artifact a user downloads, booting unmodified.

## What was actually wrong

Four independent defects, none of them in the ISO recipe:

**1. The job killed its own steps.** `iso-e2e`'s `timeout-minutes: 30` sat under a build step declaring 60 and a boot step declaring 20, so every `source: build` cell was cancelled mid-build with both steps inside their limits — `guppy:xfce` started building at 00:18:40 and was cancelled at 00:48:29. Actions reports that as `cancelled`, the same word it uses for a spot reclamation, so a six-cell sweep produced no verdict about any cell. Now conditional (90 on the build path), with `tests/test_a_job_budget_covers_its_own_steps.py` asserting no job budget is smaller than the sum of its steps'.

**2. The build ran rootless.** It built the image as root, copied it into *rootless* storage with `podman save | podman load`, then ran `just iso` bare — so tacklebox did its post-customize `podman commit` under `fuse-overlayfs`. That commit ran past tacklebox's hardcoded 600s bound, three times, always at exactly 600.0s. As root it takes 133s. Details and the full timeline are in #1893.

**3. The PR check could not pass.** The no-input cell list was `yellowfin:gnome` + `albacore:gnome`, and with no inputs the ISO must come from R2. Measured against the live bucket: `albacore-{gnome,kde,xfce}-latest.iso` are 200, `yellowfin-gnome-latest.iso` is 404 (#2027). Every PR touching this workflow went red on a publishing gap rather than a harness regression. The list is now the cell the bucket holds — which is why this check is green above.

**4. The grouped-ISO publisher threw away a working ISO.** `publish-iso-groups.yml` never installed the Kubic podman, and pinned `TACKLEBOX_SHA` to a2026-07-18 commit — an ancestor of the repository pin, so a month stale for a reason already satisfied. With those fixed it built the thing it exists to build:

```
>>> Tacklebox ISO complete: /mnt/tunaos-iso-build/albacore/albacore.iso
    offline-store    4m44.3s (34.7%)   17G raw -> 3.7G squashfs
    install:combined 4m8.7s  (30.3%)
    TOTAL           13m39.511s
```

then died on two `podman run` calls missing `--log-driver=k8s-file`, which exist only to read `VERSION_ID` and arch for the output filename. The Kubic conmon has no journald support. `build-iso-tacklebox.sh` carries the identical pair *with* the flag and a comment explaining exactly this; this was the copy that lost it. Fixed, with a mutation-verified guard across the ISO-build scripts.

Its dispatch options also offered `community` and `nvidia` — suffixes that stopped existing when `build-config.yml` went one-group-per-desktop — so the four groups that do exist could not be dispatched at all.

## Known-remaining

`skipjack:xfce` stalls at exactly 600.0s **even as root**, so root is not a universal fix and I am not claiming it is. The measured difference is the flatpak runtime each installer pulls: xfce's brings `org.gnome.Platform 49` + `codecs-extra` + `GL.default 25.08`, kde's brings `org.kde.Platform 6.9` + `GL.default 24.08`. The 600s bound is a hardcoded literal in tacklebox with no environment override, so nothing here can distinguish "slow" from "wedged". Tracked in #1893; xfce is last in the stated priority order, behind gnome/kde and then niri/cosmic.

`marlin gnome,kde` and a second grouped-ISO dry run (`skip_upload=true`, nothing written to R2) are in flight.

---

hanthor:
`🐦-hummingbird / gnome / Attest SBOM` is red on 08a85887. It is an upstream Sigstore outage, not this branch, and it does not block Promote. Not fixing it — there is nothing here to fix.

`rekor.sigstore.dev` is returning 502:

```
Post "https://rekor.sigstore.dev/api/v1/log/entries": giving up after 2 attempt(s):
status 502: <!doctype html>...502 Bad Gateway
```

The step's retry budget behaved exactly as designed — 6 attempts over 10 minutes (15s, 30s, 60s, 120s, 120s, 120s) — and then emitted its purpose-built terminal error rather than a generic one:

```
##[error]SIGSTORE_OUTAGE: attest still failing after 10m of Sigstore unavailability; giving up.
```

## Why it does not gate publication

`tag-image` (Promote) declares `needs: [manifest, sign, verify_boot, verify_asahi]`. `attest_sbom` is deliberately not among them — that decoupling is the fix from #1560, and this run is the first time I have watched it actually pay off. Current state of the four that do gate it:

| job | result |
|---|---|
| Manifest | success |
| Sign | success |
| Asahi manifest gate | skipped (amd64-only flavor) |
| Gate | in progress |

`Sign` is worth calling out because it talks to the same Rekor. It ran 05:51:07→05:51:21 and the first attest 502 was at 05:51:46 — it got in 25 seconds ahead of the outage. Had it landed the other way round this would have been a genuine blocker, so the green here is timing, not immunity.

## Unrelated, and the reason this run exists

The same run cleared the failure that has kept `hummingbird:gnome` unpublished for six consecutive nightlies. `Build qcow2 from testing image` succeeded in 1m54s; it has died every night since 2026-08-20 on `grub2-install: error: unknown filesystem` against xfs. Whether the boot verify passes and Promote runs is still open at the time of writing — I am not claiming the cell is green, only that it got past the step that has always stopped it.

---

hanthor:
`E2E hummingbird:gnome` is red. Diagnosed, not fixed, and the fix is not in this repo.

This is a **different** failure from the `Gate` one above — that was the boot gate on the qcow2; this is the ISO build.

## It is not a timeout, despite what it says

```
##[error]The action 'Build the ISO on this runner' has timed out after 60 minutes.
```

It failed after **seven** minutes and then sat there:

```
05:55:45  flatpak missing; installing it for the customize step
05:55:45  + dnf5 install -y flatpak
05:55:45  Failed to resolve the transaction:
05:55:45  No match for argument: flatpak
05:55:45  ERROR: flatpak not installed and could not be installed;
05:55:45    cannot pre-install org.bootcinstaller.Installer
05:55:45  + exit 1
06:48:48  === podman storage (repeated at exit): root=overlay rootless=overlay ===
06:48:53  ##[error]... has timed out after 60 minutes.
```

53 minutes of silence between a clean `exit 1` and the job cap. Filed as #2044 — the masking is its own defect, because "timed out after 60 minutes" is also the signature of #1893 and #2034, which are genuinely about slowness. Three causes, one red.

Ruled out rather than assumed: the tacklebox 600s cap did **not** fire (`exit status 124` and `600.0s` appear nowhere in the log — the only matches are this workflow's own comments quoted back), and assembly never started, so it is not #1893 either.

## Why I am not fixing it here

`flatpak` is not resolvable in hummingbird's repos. I checked whether this was a repo-set problem inside the customize container rather than a genuine absence, because those have very different fixes — it is not: `dnf5 install -y dbus-daemon` **succeeded in that same container two seconds earlier**. The repos resolve fine. `flatpak` specifically is absent, and absent from the built image, while `bubblewrap` and `ostree` are present.

`live-iso/common/src/customize-live.sh` needs it to pre-install `org.bootcinstaller.Installer`, so no hummingbird ISO can be built until that is available. Same class as the missing GNOME stack — hummingbird's repos are incomplete (tunaos-packages#519). Requested as a manual step in tunaos-packages#520.

## What this branch did establish

The ext4 drop-in worked. `Build qcow2 from testing image` passed in 1m54s after six consecutive nightlies died on `grub2-install: error: unknown filesystem`, confirmed by `Creating root filesystem (ext4)` in the Gate log rather than inferred from the step going green.

It just revealed a larger blocker underneath: `hummingbird:gnome` carries 410 packages and no GNOME. Neither red on this PR is caused by the changes in it.

---

hanthor:
**Correction to my previous comment on this PR.** The `flatpak` and missing-GNOME findings are correct and measured; the *cause* I gave for them is not.

I wrote that hummingbird's repos are incomplete "(tunaos-packages#519)". Hummingbird does not read the repo #519 is about. It reads `repo.tunaos.org/hummingbird/20251124-$basearch/` plus `public-hummingbird-x86_64-rpms`; #519 concerns `repo/10/x86_64`. I linked them on a shared symptom without checking the repository.

Measured against the live hummingbird index — 8,100 packages, 43 matching `gnome-*`:

| | |
|---|---|
| present | `gtk4`, `gdk-pixbuf2`, `mutter`, `gvfs` |
| **absent** | `gnome-shell`, `gdm`, `nautilus`, `harfbuzz`, `gnome-desktop3`, `flatpak` |

`gtk4` is served and requires the absent `harfbuzz`, so dnf reports it as **broken** rather than missing and `--skip-unavailable` drops it. Same shape as #519, different repository.

The framing was also wrong, and it is the root of the error: hummingbird is a rolling, security-hardened fork tracking **Fedora Rawhide**, running the ARK kernel, and it **ships no desktop environment by design**. The `.fc43` dist tags are Rawhide's numbering, not Fedora 43. So `hummingbird:gnome` is a port onto a deliberately desktop-less base — absent packages there may be intended, not lost.

`2e8d9709` on this branch adds [`docs/HUMMINGBIRD.md`](https://github.com/tuna-os/tunaOS/blob/claude/image-green-status-gaps-w6z6qk/docs/HUMMINGBIRD.md) and closes the check that let this hide: `check-package-repo-pins.py` verified only that a pin *resolves*, which is the wrong property for a rolling distribution's snapshot — it now fails past 180 days (`STALE ... 274d old`, exit 1). Every pin in the build that produced the desktop-less image resolved, and nothing was red.

Nothing changes about this PR's own reds: neither is caused by the changes in it, and the ext4 result stands — `Build qcow2 from testing image` passed in 1m54s with `Creating root filesystem (ext4)` confirmed in the log.

---

hanthor:
**The ext4 drop-in is proven.** Section 0 said this was being measured; it has now been measured, on run [32813037866](https://github.com/tuna-os/tunaOS/actions/runs/32813037866) (commit `08a8588`, gnome Gate job [97699576724](https://github.com/tuna-os/tunaOS/actions/runs/32813037866/job/97699576724)):

| step | before (run 32786338463) | with `50-hummingbird.toml` |
|---|---|---|
| Build qcow2 from testing image | **failed** — `grub2-install: unknown filesystem` | **success, 1m54s** (05:51:15 → 05:53:09) |

`bootc install to-disk` completes. The `grub2-install` abort that has failed six consecutive nightlies is gone, and `Manifest` and `Sign` both passed on the same run.

**The Gate still fails, but at the next step and for a different reason.** `Boot and verify` ran 05:53:09 → 06:08:28 and failed after 15m19s — a timeout, not an install abort. That is a genuinely new failure: the disk now builds and the VM now gets far enough to be waited on.

I am not going to name the cause from the symptom, since the evidence is uploaded (`boot-gate-hummingbird-gnome`, artifact 9551285788, serial.log + 1 frame) and can be read. What is worth stating is the arithmetic that makes a boot timeout unsurprising here: `hummingbird:gnome` is base + 118 packages, and `gdm`, `gnome-shell`, `gnome-session-wayland-session`, `nautilus`, `gnome-control-center` and 17 other manifest packages exist in **neither** our rebuild repo nor upstream yet. 202 of the chain's 673 packages remain unbuilt (471 are served). A VM with no display manager has nothing to reach.

So this PR does what it claimed and no more: it clears the **install** half of the Gate. The **boot** half is waiting on packages, which is what tuna-os/tunaos-packages#512 builds — that one is queued to merge ahead of its 12:00 UTC nightly, and 202 packages is roughly 2.5–4.7 h of chain against a ~13.5 h nightly budget.

Queuing this for merge on that basis: it is a proven fix for a live six-night failure, it cannot regress a Gate that already fails earlier, and landing it now means the 22:20 UTC nightly exercises it on `main` against whatever the package chain has by then.

One nit from `github-code-quality` is still open (an unused `sig = inspect.signature(...)` in `tests/test_a_rolling_snapshot_pin_cannot_go_stale_unnoticed.py`); I am holding that edit until the in-flight Build Hummingbird run on this branch finishes, because pushing here cancels it — that is what killed run 32809685474.

---

hanthor:
Read the boot evidence rather than leaving the cause named-by-inference. It settles the question, and in this direction the news is better than the failure looked.

```
==> Booting disk image: hummingbird.qcow2 (qcow2)
==> Waiting up to 900s for the desktop contract marker...
serial| [  OK  ] Reached target graphical.target - Graphical Interface.
...
ERROR: desktop contract marker was not emitted
==> Screenshot 10-ready painted on attempt 1 (stddev=0.0272045)
```

**The ext4 disk boots.** It is not an install that produces an unbootable image — the guest comes all the way up and reaches `graphical.target`. The Gate fails on one thing only: no desktop session ever announces itself, so the contract marker never appears and the 900 s wait expires. The screenshot is effectively blank (stddev 0.027), which is what a reached-graphical-target-with-nothing-on-it looks like.

That is exactly the shape the package arithmetic predicts, and it is now measured instead of assumed: `gdm` and `gnome-shell` are not in the image because they exist in neither repo. There is nothing for `graphical.target` to start.

So both halves of this PR's claim are now evidenced:

- **install** — `bootc install to-disk` completes (1m54s, was `grub2-install: unknown filesystem`);
- **boot** — the resulting disk boots to `graphical.target`.

The remaining Gate failure is not a tunaOS defect and no change in this repository fixes it. It is the 202 unbuilt packages of the 673-package chain, which tuna-os/tunaos-packages#512 builds.

One unrelated unit failure is visible in the same serial log and worth a follow-up issue rather than a scope creep here: `Failed to start rechunker-group-fix.service - Fix groups for Legacy rechunker`. It does not block the boot.

---

Issue #1802, opened 2026-08-16
[guppy] kde build cancelled at the 180-minute job ceiling with 245 of 252 emerge jobs complete


hanthor:
## What

On run [31942667550](https://github.com/tuna-os/tunaOS/actions/runs/31942667550), `guppy / kde / linux-amd64` ([job 95158650047](https://github.com/tuna-os/tunaOS/actions/runs/31942667550/job/95158650047)) was cancelled during **Build Image** after **2h57m24s** (11:35:49 → 14:33:13), hitting `timeout-minutes: 180` on the `build_push` job (`.github/workflows/reusable-build-image.yml:109`).

It was not hung. The last output before the cancellation:

```
>>> Completed (244 of 252) kde-plasma/kwin-6.6.5-r1::gentoo
>>> Jobs: 245 of 252 complete, 1 running          Load avg: 2.61, 3.85, 4.52
>>> Emerging (246 of 252) kde-plasma/plasma-workspace-6.6.5-r1::gentoo
##[error]The operation was canceled.
```

**245 of 252 emerge jobs complete — roughly 97% — and still compiling steadily** (load average 2.61, packages completing minutes apart). It ran out of clock, not progress.

## Consequence

`kde / Manifest` then failed at **Load Outputs** — the cancelled build wrote no digest artifact — and `Sign`, `Gate`, `Promote` and `Attest SBOM` were all skipped. One cell lost to the ceiling.

## Context

The other guppy flavors on this run, same runner class:

| job | Build Image duration |
| :--- | ---: |
| `base / linux-amd64` | 35m42s |
| `xfce / linux-amd64` | ~1h57m |
| `gnome / linux-amd64` | ~1h59m |
| `kde / linux-amd64` | **2h57m+, cancelled** |

KDE Plasma is simply the largest desktop to build from source, and it is the only one that crosses the line.

This is the same wall as #644, one ceiling higher. That issue was `timeout-minutes: 60` and offered two fixes: (1) use the Gentoo binhost so `emerge` installs prebuilt binaries, (2) raise the timeout. Raising it is what happened, and the comment at `reusable-build-image.yml:106-108` records why — *"to give the source-based experimental variants headroom … A max, not a fixed duration: RPM variants finish in ~20-40m unaffected."* Three months on, the same shape has recurred at 180.

## Options

[...]

Issue #1755, opened 2026-08-15
[hummingbird] Desktop flavors cannot work as configured — the rebuild repo is missing gnome-shell/gdm, is x86_64-only, and three of four desktops have no manifest section


hanthor:
Research issue for #1754's first item. `hummingbird` has been red for 10+ consecutive nightlies (1/5 cells green — only `base`). #1555 attributes it to unresolvable dnf deps; **that specific cause is already fixed and is no longer what fails**. The real blockers are structural, and this is what I found measuring them.

---

## TL;DR

| Blocker | Cells | Fixable how |
| :--- | ---: | :--- |
| Rebuild repo has no `gnome-shell`, `gdm`, `mutter` | gnome ×2 | Rebuild those packages |
| No aarch64 rebuild repo exists at all (404) | 4 arm64 | Build one, or drop arm64 |
| `kde` / `niri` / `cosmic` have no `hummingbird:` manifest section | 6 | Author sections — **cosmic is 1 package away** |
| Repo pinned to a 9-month-old datestamp | all | Publish a floating alias |

**Quickest real win: `cosmic` on amd64.** 22 of its 23 packages are already in the repo. It fails only because nothing tries.

---

## 1. #1555's stated cause is already handled

#1555 says `libjsoncpp.so.26` / `librhash.so.1` are unresolvable. That was real, and `build_scripts/desktop/kcm-ublue.sh:78-84` already fixed it — `cmake` resolved by name but its providers didn't, and that now degrades to a warning:

```
::warning title=kcm_ublue skipped::dnf install of build deps failed despite
passing the repoquery probe — an unresolvable transitive dependency
```

So kcm_ublue is skipped, not fatal. **Something else is failing the builds**, and #1555 will not close them.

## 2. What the rebuild repo actually contains

The `hummingbird:` manifest section points at `https://repo.tunaos.org/hummingbird/20251124-x86_64/`. I pulled its `primary.xml.gz` and diffed it against the manifests.

The repo is **live and has 7,168 packages**. It is also missing the desktop:

```
gnome-shell   NO      niri            yes
gdm           NO      cosmic-session  yes
mutter        NO      xfwm4           yes
plasma-workspace NO   sddm            yes
```

Against `manifests/desktops/gnome.yaml`'s `hummingbird:` list — **23 of 52 present, 29 missing**:

[...]

Issue #858, opened 2026-07-27
marlin:kde ships no desktop session — customize-live detects it as gnome


hanthor:
`ghcr.io/tuna-os/marlin:kde` contains **no `/usr/share/wayland-sessions/` directory at all**:

```
$ podman run --rm --entrypoint sh ghcr.io/tuna-os/marlin:kde -c 'ls /usr/share/wayland-sessions/'
NONE
```

So the live payload's desktop detection falls all the way through to its default:

```
+ [[ -f /usr/share/wayland-sessions/plasma.desktop ]]
+ [[ -f /usr/share/wayland-sessions/plasmawayland.desktop ]]
+ [[ -f /usr/share/wayland-sessions/niri.desktop ]]
+ [[ -f /usr/share/wayland-sessions/cosmic.desktop ]]
customize-live: detected desktop=gnome
+ INSTALLER_APP=org.bootcinstaller.Installer
```

An image tagged `kde` gets the **GNOME** installer frontend and GNOME autologin wiring.

## Why it matters

The tag doesn't describe the contents. Anything downstream that trusts `marlin:kde` to be a KDE image is wrong — the live overlay, the browser ISO builder, and anyone pulling it directly.

Note the detection is not at fault: with no session files present there is nothing to detect, and defaulting to gnome is a reasonable fallback. The image is what's wrong.

## How it surfaced

Found while diagnosing why the `marlin:kde` live-overlay build kept failing. That failure had a *different* cause (`useradd: cannot create directory /home`, fixed in #857) which masked this one — the build died before anything would have noticed the desktop mismatch. With #857 merged, `marlin:kde` will produce a **green overlay for the wrong desktop**, which is worse than a red one.

## What to check

- Does the marlin build actually install a KDE session for the `kde` flavor? Arch's `plasma-meta` / `plasma-workspace` ships `/usr/share/wayland-sessions/plasma.desktop`.
- `scripts/verify-image-packages.sh` has a `kde) SYSTEMD_SERVICES+=(sddm.service|plasmalogin.service)` gate and `build_scripts/checks/verify-desktop-experience.sh` requires `require_glob '/usr/share/wayland-sessions/*plasma*.desktop'` — worth understanding why neither caught this for marlin. If those checks don't run for Arch-based variants, that gap is the more valuable fix than the packaging itself.

[...]

PR #2081, opened 2026-08-26
hummingbird gnome fixes from run 32907940350 + Hive practices, first slice (#2250): CI contract, fork safety, randomized lane, regression policy, nudges, graduation

hanthor:
Both of these come out of one job: Build Hummingbird #66, [run 32907940350](https://github.com/tuna-os/tunaOS/actions/runs/32907940350), `🐦-hummingbird / gnome / linux-amd64`. The cosmic cell of the same run finished in 14 minutes and its ISO is building. The gnome cell ran 3h57m and published nothing.

---

## 1. `--skip-unavailable` was on the wrong side of `install`

`--skip-unavailable` is **command-scoped** in dnf5. `dnf -y --skip-unavailable install …` is not a slightly-odd ordering of the same command — it is a usage error that never reaches a repository:

```
Unknown argument "--skip-unavailable" for command "dnf5". Add "--help" for more information about the arguments.
The argument is available for commands: … install … (It has to be placed after the command.)
```

Two call sites in `build_scripts/desktop/install-desktop.sh` had it that way, both on the `IS_HUMMINGBIRD` branch. Both are written `dnf_retry … || install_available …`, so the `||` caught our own usage error and quietly re-ran the whole package set through the fallback.

That fallback is not equivalent to the transaction it replaced:

* it **drops the `-x` excludes** built from the manifest's `exclude:` list,
* it forces `--setopt=install_weak_deps=False`,
* it resolves package-by-package via `dnf repoquery`, one invocation each.

So the build composed a materially different image than the manifest describes — and paid four `dnf_retry` attempts plus 35s of backoff re-issuing a command line dnf had already rejected as malformed. All **52** gnome desktop packages went through it.

This is a different failure from the ones #1755 catalogued (missing packages in the rebuild repo). Here dnf never got as far as looking at a repo.

**Changes:** move the flag after `install` at both sites (the package transaction and the COPR block, which had the same shape). Teach `dnf_retry` to treat `Unknown argument` / `Unknown command` as final, the way it already treats unresolvable transactions, and emit a `::error::` annotation — because a caller's `||` otherwise swallows the exit code and the next one is invisible again, which is exactly how this one survived.

## 2. The build wedged for 3h48m and reported it as a cancellation

```
23:12:27  [13/30] pcre2-utf32-0:10.47-1.2.hum1 100% | 247.5 KiB | 00m00s
03:00:31  ##[error]The operation was canceled.
```

Seventeen packages still in flight, then not one further line until the job's 240-minute ceiling fired. A cancelled job writes no digest artifact, so `Manifest` failed and `Sign`, `Gate`, `Promote`, `Attest SBOM` and the gnome ISO were all skipped — and the log named nothing.

Raising the ceiling would make that worse, not better. Two bounds instead:

* **`/etc/dnf/dnf.conf`** gains `minrate=10240`, `timeout=120`, `retries=3`, appended in the same guarded block that already sets `max_parallel_downloads`, so every downstream `RUN` layer inherits it. librepo's defaults treat a connection trickling at 1000 B/s as healthy and wait indefinitely; these turn a stalled transfer into an error `dnf_retry` can act on. Deliberately generous — `repo.tunaos.org` is one small host, and the point is to catch a *wedge*, not to fail a slow mirror. The config drop-in is the right layer because the wedge was **not** in a call this repo wraps: it was a plain `dnf install` of build dependencies.

* **The build invocation** gains `timeout --kill-after=60s 210m`, *inside* sudo so the signals come from root — a non-root `timeout` cannot kill a root `buildah`. 210 against the job's 240 is not a guess at how long a build should take; the ceiling above already clears the slowest real one (guppy kde, 2h57m compiling plasma from source, per #1802). It exists so a build that was going to blow the cap anyway fails 30 minutes early with an exit code and an annotation.

The repo's own precedent — the `Generate SBOM` comment about `timeout-minutes: 20` being "enforced by the runner AGENT — useless precisely when the agent is the casualty" — is why this is an in-command wall clock rather than a step-level `timeout-minutes`.

## 3. Every live-ISO customize since 886444e3 died at the session bus (#2265)

Found because this PR's own `build` check was red. main's 886444e3 runs `dbus-daemon --fork --pidfile=…`, and dbus-daemon has no `--pidfile` option (only `--nopidfile`; the path is config syntax): usage error, exit 1, on a discarded stderr, so no session bus and the customize dies under `set -eu`. Every `live-iso-bootc.yml` run from 2026-08-25 on ended there (32875192246, 33594824101); the last green one is the commit before. The suite stayed green because the fake dbus-daemon in the fails-fast test honoured the option the real binary rejects.

`_start_bus` now uses `--nopidfile --print-pid=3 3>"$_pidfile"` (a descriptor on a file, not `$( )`, so nothing waits on a pipe the daemon holds). The fake bus rejects `--pidfile` like dbus 1.14 does, and `tests/regressions/test_issue_2265_*.py` runs the script's own helper under `set -eu` against the real dbus-daemon, and proves the shipped form fails against it. Verified locally with dbus 1.14.10; the `build` check went green on 5c9cb514.

## 4. Hive practices, first slice (#2250)

Covered in the comment below: CI contract self-audit (`scripts/check-ci-contract.py`, `just test-contract`), fork-safety static test, randomized nightly lane, incident-to-regression convention (`tests/regressions/`), advisory PR nudges, advisory-to-blocking graduation checker, contributor-contract recipes, and the legacy cruft removed.

## 5. `hummingbird:gnome` takes GNOME from projectbluefin/utah-packages, by digest (tunaos-packages#629)

Bluefin builds GNOME 51 for Hummingbird in Hummingbird's own root and ships it as an OCI image carrying a `createrepo_c` repository, cosign-signed. Building a second GNOME 51 for the same base in tunaos-packages had not resolved on the target (30 unresolved capabilities). Consume utah's instead, the way `common` and `brew` are consumed:

- `image-versions.yaml` pins `utah-packages` by digest (Renovate's existing image-versions block tracks it); `registry-map.yaml` names it; `resolve-image.sh` gains the role; `build-image-inner.sh` passes `UTAH_PACKAGES_IMAGE_REF`.
- `Containerfile.el10` declares the stage and **bind-mounts** `/repository` at `/run/utah-packages` for the gnome stage's `install-desktop.sh` RUN only. Never copied: ~500 MB.
- `manifests/desktops/gnome.yaml` hummingbird section: `utah-packages` repo at `file:///run/utah-packages`, priority 4 (beats the `tunaos-hummingbird` prefix at 5), `unsigned: true`.
- `install-desktop.sh`: `unsigned: true` writes `gpgcheck=0` and is **refused unless the baseurl is `file://`** — digest-pinned bind-mounted content is the one shape where the digest stands in for a signature; anything fetched over the network stays signed (#1655). The bats test that pinned the signature block now pins that rule.
- `tests/test_hummingbird_gnome_consumes_utah_packages.py` (10): the pin, the plumbing end to end, the manifest, and the repo loop run for real against a fake yq.
- `docs/HUMMINGBIRD.md` and `AGENTS.md` say where GNOME comes from now. The factory side (utah counted as had, container installability gate) is tunaos-packages#631.

## 6. tromso and xfce-linux are listed as sibling images, built elsewhere

They are TunaOS-family bootc images built in their own repositories with BuildStream on freedesktop-sdk: whole-OS products, not package sources, so they have no cells here and are not scored by green-criteria. `.github/build-config.yml` gains `sibling_images:`; `.github/scripts/sibling-images-status.sh` renders each one's latest **concluded** main-branch build into the README block after the variant matrix and outside its totals (fetch failure is ⬜, never a verdict, never stops the refresh); README "Choose your image" gains two rows. `tests/test_sibling_images_are_reported.py` (8) runs the renderer against a fake gh.

## 7. chezmoi v2.72.1 checksum pins (main's, ported)

main's 513f9524 bumped chezmoi and kept v2.72.0's sha256 pins, so `checksums` has been red on every PR since 2026-08-30. Regenerated with the script's own `--fix`.

---

## Tests

21 new for parts 1 and 2, in three places:

| File | Covers |
| :--- | :--- |
| `tests/test_dnf_flags_land_after_the_subcommand.py` (new, 10) | A tree-wide placement lint, plus behavioural tests that run the real `dnf_retry` from `lib.sh` against a stubbed `dnf` and count invocations. |
| `tests/test_a_stalled_dnf_download_cannot_wedge_the_build.py` (new, 5) | The transfer bounds exist, are generous enough for a small single-host mirror, are written before the first install in that script, and are idempotent. |
| `tests/test_build_ceiling_clears_the_slowest_variant.py` (+6) | The step bound exists, sits inside sudo, is below the job ceiling with room to report, still clears the slowest real build, annotates both timeout exit codes, and re-raises a normal build failure. |

The placement lint follows backslash continuations and splits on `&&`/`||`/`|`/`;` so a fallback on the right-hand side isn't parsed as dnf arguments. It covers `--skip-unavailable`, `--skip-broken`, `--nobest`, `--best`.

**Every change was mutation-checked**, and each mutation fails a test that names the specific consequence:

| Mutation | Fails |
| :--- | :--- |
| restore the misplaced flag | `test_no_shell_script_puts_a_command_scoped_flag_before_the_subcommand` |
| delete the usage-error branch from `dnf_retry` | `test_a_usage_error_is_not_retried`, `test_a_usage_error_is_annotated_…` |
| drop the `dnf.conf` bounds | all 5 in that file |
| drop the wall-clock `timeout` | `test_the_build_is_bounded_inside_the_step_not_only_by_the_job` |
| move `timeout` outside `sudo` | `test_the_timeout_runs_under_sudo_so_it_can_signal_a_root_buildah` |
| drop the `rc` re-raise | `test_a_normal_build_failure_still_fails_the_step` |
| restore `--pidfile=` in `_start_bus` | `test_the_fix_keeps_the_bus_off_the_build_pipe`, `test_the_shipped_helper_fails_against_the_real_daemon` |
| mark an `https://` repo `unsigned: true` | `test_unsigned_is_refused_for_a_network_repo` |
| `COPY --from=utah-packages` instead of a mount | `test_the_containerfile_declares_the_stage_and_mounts_it_on_gnome_only` |

`tests/bats/test_lib.bats` still passes 50/50.

## Not fixed here

`install_available` drops the `-x` excludes that the primary transaction passes. With the flag placement corrected the fallback should be rare, but it is still not equivalent to the transaction it stands in for, and that deserves its own change.

`tests/test_package_parity_is_scheduled.py::test_audit_roster_derives_from_build_config` is red on main and untouched here.


hanthor:
`🐦-hummingbird / gnome / Attest SBOM` is red on [run 32925587829](https://github.com/tuna-os/tunaOS/actions/runs/32925587829/job/98050649895). It is not this PR's, and it does not block anything.

**What failed.** Every one of six `cosign attest` attempts got the same response from the transparency log:

```
Post "https://rekor.sigstore.dev/api/v1/log/entries": giving up after 2 attempt(s):
status 502: 502 Bad Gateway
```

The step's own classifier reached the same conclusion and said so before giving up: `SIGSTORE_OUTAGE: attest still failing after 10m of Sigstore unavailability`.

**Why it isn't this PR's.** The diff touches `install-desktop.sh`, `lib.sh`, `10-base-packages.sh` and the `Build Image` step. It has no path to Rekor. `Sign` — which uses the same cosign and the same OIDC identity — passed at 03:28:35 in this run, a minute before Rekor started returning 502.

**Why nothing is blocked.** `promote` needs only `publish-image`. `reusable-build-image.yml` says why, in as many words:

> SBOM attestation deliberately does NOT happen here — it is its own job (attest_sbom) off the critical path… on the 08-15 nightly every albacore flavor built, pushed and passed its gates, and all twelve Promote jobs were skipped because `cosign attest` could not reach Rekor. Image signatures are a promotion precondition; an SBOM attestation that a transparency-log outage delayed by an hour is not (#1560).

This run is that design working: `Manifest` ✅, `Sign` ✅, `Gate` still booting the image, `Attest SBOM` ❌ and off to one side.

**What I did.** Re-probed Rekor at 03:47 — `https://rekor.sigstore.dev/api/v1/log` now answers `200`, so the outage has recovered — and re-ran the failed job once. No fix to port: there is no change to this PR that would have prevented a 502 from an external transparency log.

For context on the rest of the run: the gnome cell built in **11m33s** and pushed to GHCR, against 3h57m-then-cancelled on the previous nightly. That is the stalled-download bound and the dnf flag placement in this PR doing their job.

---

hanthor:
Correction to the comment above: I said I had re-run the failed job. I hadn't — GitHub refused it with `403 This workflow is already running`, because `Gate` is still booting the image in the same run. Everything else in that comment stands; the re-run is queued behind the run finishing and I'll fire it then.

Current state of [run 32925587829](https://github.com/tuna-os/tunaOS/actions/runs/32925587829): `Manifest` ✅, `Sign` ✅, `Attest SBOM` ❌ (Rekor 502, off the critical path), `Gate` still in `Boot and verify`. Its `Build qcow2 from testing image` step passed in under two minutes — that is the step that failed six consecutive nightlies with `grub2-install: unknown filesystem` before the ext4 override.

---

hanthor:
Two more reds on [run 32925587829](https://github.com/tuna-os/tunaOS/actions/runs/32925587829), separate from the Sigstore one. One I have pushed a fix for; the other has no fix to push.

## `gnome / Gate` — fixed in `ecb67b51`

The image booted **fine**. It reached `graphical.target`, showed a login prompt, ran 15 minutes with no failed units and powered down cleanly. What it never did was print the marker the gate greps for, so `iso-e2e.sh` waited its full 900s and exited 2, "readiness marker not seen".

The serial log says why:

```
[ 7.686065] systemd[1]: Applying preset policy.
[ 7.989555] Removed '…/multi-user.target.wants/tunaos-base-contract.service'
[ 8.001120] Removed '…/graphical.target.wants/tunaos-desktop-contract.service'
```

`systemctl enable` at build time writes a symlink under `/etc/systemd/system/<target>.wants/`. On first boot systemd applies preset policy, Fedora's catch-all is `disable *`, and any unit no preset file names has that symlink deleted. The build is long over by then, so nothing in CI has ever seen it. Seven units went, including `tunaos-live-ready.service` — the same marker on the ISO path.

In the same pass systemd *created* symlinks for `brew-update.timer` and `flatpak-appstream-refresh.service`, which presets do name. The mechanism works; ours were not in it.

`50-tunaos.preset` already existed and did one thing — disable `tracker-extract-3`. It now names the seven. Because `Gate` is a `needs:` of `Promote`, this was blocking a healthy image from ever publishing.

I also shipped a stray `99-tunaos.preset` in that commit — `git mv` from a mutation test leaked through a directory-level `git add`. Removed in `33916c29`, and the test now walks every `*.preset` we ship rather than the one it was told about.

## `iso:gnome (linux-amd64)` — blocked, nothing to port

First hummingbird **gnome** ISO build ever attempted; it got there because the cell published a digest this run and the provenance gate passed. It failed after **one second**:

```
+ dnf5 install -y flatpak
No match for argument: flatpak
ERROR: flatpak not installed and could not be installed; cannot pre-install org.bootcinstaller.Installer
+ exit 1
+ _reap_live_buses
```

Everything above that line worked: the image pulled, `baseline.sh` ran, `dbus-daemon` installed from `public-hummingbird`, and the system bus started. `flatpak` is absent from both `public-hummingbird` and our rebuild snapshot, so `ensure_flatpak`'s dnf fallback cannot rescue it. It is `layer-07` of `build-order-hummingbird-desktops.yml` and unbuilt.

There is no change to this PR that would make an absent package present. Worth noting what did work: the failure took one second and the `EXIT` trap reaped the bus, against 57 minutes of silence for the same failure before #2066.

---

hanthor:
Update on `Attest SBOM`, and a correction: I said the Sigstore outage had recovered and re-ran the job. The re-run failed identically — seven attempts, 03:52:35 → 04:00:39, every one `502 Bad Gateway` from `rekor.sigstore.dev/api/v1/log/entries`.

My "recovered" call was based on a `GET https://rekor.sigstore.dev/api/v1/log` returning `200` from this machine. That was the wrong measurement twice over: it tested the read path, not the write path, and it tested a different network than the runners.

Probing properly, from here, at 03:59:

```
GET  /api/v1/log         : 200
POST /api/v1/log/entries : 422  {"code":602,"message":"kind in body is required"}
```

A `422` means the backend received the request and rejected my deliberately-empty payload — so the write path is alive **from this network**. The runners get `502` from the same endpoint in the same minutes, which is a gateway that cannot reach the backend. So this is a partial or regional Sigstore outage, still ongoing for GitHub Actions egress, not a recovered one.

Nothing else changes: the classification (not this PR's), the reasoning (`Sign` passes with the same cosign and OIDC identity; the diff has no path to Rekor), and the impact (`promote` needs only `publish-image`, per #1560) all stand. The one re-run allowed is now spent and I won't re-run it again. I'll keep the PR watched until it is green.

---

hanthor:
`gnome / Gate` is red again on the run that carries the preset fix ([32928686075](https://github.com/tuna-os/tunaOS/actions/runs/32928686075/job/98059997129)). **The preset fix worked; the Gate is failing for an honest reason that no change in this PR can address.**

## The preset fix is validated

Boot evidence, same image, before and after:

| | before (`d08af8e4`) | after (`33916c29`) |
| :--- | ---: | ---: |
| units stripped by `Applying preset policy` | **8** | **1** |
| `tunaos-live-ready.service` | removed | **Started → Finished** |
| `TUNAOS_LIVE_READY` on the console | absent | **`uptime=12.91`** |

The one remaining removal is `/etc/systemd/system/default.target`, which presets cannot control and which the commit already calls out as unfixed. The marker the ISO path depends on now prints — that is the fix doing exactly what it was for.

## Why Gate is still red

`tunaos-desktop-contract.service` has **zero** mentions in the serial log — it never started. Its own unit says why:

```ini
After=display-manager.service dconf-update.service
Wants=dconf-update.service
Requires=display-manager.service
```

`gdm` is one of the 38 packages missing from this image, so `display-manager.service` does not exist. Confirmed in the boot log: `display-manager` 0 hits, `gdm` 0 hits. A `Requires=` on a unit that is not installed can never be satisfied, so systemd never runs the contract, so `TUNAOS_DESKTOP_CONTRACT_*` is never printed, so `iso-e2e.sh --disk --contract desktop` waits its full 900s and exits 2.

`tunaos-base-contract.service` reaches `Starting` and never finishes, and `graphical.target` is reached with nothing behind it.

**This is the gate working, not a false negative.** It is refusing to pass a desktop image with no desktop — 14 of the 52 packages the manifest asks for, no `gnome-shell`, no `gdm`. Making it pass would require either installing a display manager that does not exist in either index, or loosening a `Requires=` that is correct as written. I am not going to do the second to get a green tick.

So `Promote` stays blocked, correctly, until `gdm` exists — layer-07 of `build-order-hummingbird-desktops.yml`, alongside `flatpak`, which is the same wall the `iso:gnome` job hits. Both are downstream of tuna-os/tunaos-packages#534, which is what stops the chain restarting at `bootstrap-00` on every publish.

Nothing to port and nothing to re-run: a re-run reproduces this exactly, since the package set is the same.

---

hanthor:
Rebased — the conflict is resolved and the PR is `mergeable_state: clean` again.

**What conflicted:** one file, `tests/bats/test_customize_live.bats`. This branch and `main` independently fixed the *same* stale assertion — the test grepped for the literal `dbus-daemon --session`, a string 886444e3 removed when it moved the call into the `_start_bus` helper. Two corrections of one problem, landing on overlapping lines.

**How I resolved it:** kept a single assertion pair rather than either side verbatim, tightening the regex to the intersection of what both sides were guarding against:

```bash
grep -qE '_start_bus [^#|]*--session' "${SCRIPT}"
grep -q 'dbus-daemon "$@"' "${SCRIPT}"
```

The required space rules out the helper's own definition line (`_start_bus() {`), which the branch's `_start_bus[^#]*--session` would have matched had the flag ever moved onto it; `[^#|]` rules out both a comment (this branch's guard) and a pipeline (main's guard) carrying the flag. Checked positively against `live-iso/common/src/customize-live.sh:514` and negatively against the definition line. Main's comment detail — that the stale assertion failed on every PR rather than only live-iso ones, because Unit Tests runs the whole bats suite — is folded into the surviving comment; the attribution to 886444e3 is kept, since that commit's subject is the wedge fix that actually introduced the helper.

Nothing from either side's intent was dropped, and none of this PR's own changes were touched by the merge.

**Verified:** `test_customize_live.bats` 26/26, this PR's three test files 24/24, full local pytest suite 977 passed. All 11 CI checks green on `7400a192`, Unit Tests included.

Clearing `needs-rebase`.

---

hanthor:
The reported unused `shutil` import is already removed on the current PR head (`4589a913`, `test: drop an unused shutil import`). No further code change is needed.
---
🐝 **Hive Agent**: `contributor` | **SHA:** `unknown`

hanthor:
## Part 2 — Hive practices, first slice (#2250)

This branch was still open with the hummingbird fixes above, so the first #2250 slice landed on it (commit 7b60f805) rather than opening a second PR against the same head. The original description stands for part 1; this is part 2.

### Summary

Each item is a test, or a script with a test — not a doc promise.

**CI contract (item 1).** Every criterion in `.github/green-criteria.yml` now carries a machine-readable `gates:` block (workflow → job → verdict step → scripts) and a `freshness_sla_days`. `scripts/check-ci-contract.py` walks it and `tests/test_ci_contract.py` runs it on every PR that touches a workflow or the criteria file (test.yml's paths were widened; before this a workflow edit could not trigger the test that checks workflows). It proves each gate exists, is reachable from an active trigger through `uses:` chains, cannot be `continue-on-error`'d when blocking, cannot skip on the flavor without a declared `scope`, and that the schedule fires at least as often as the SLA. First run found one drift: `arch_honesty`'s prose named only its script, not the workflow.

**Fork safety, static half (item 13).** `tests/test_fork_safe_workflows.py`: every `pull_request` workflow declares `permissions`, never uses `pull_request_target`, and every step that pushes/comments/reads a secret is fork-aware (following preflight-step indirection, the iso-e2e #1129 shape). Three real findings fixed: `just-fix.yml` did an unconditional `git push` on every PR (403 on a fork → red check; it now delivers the diff to the job summary and exits 0), `validate-renovate.yaml` had no permissions block, `live-iso-bootc.yml` exported four R2 secrets and an `upload_r2` input that nothing reads.

**Randomized lane (item 3).** `tests/conftest.py` shuffles test order from `TUNAOS_TEST_SEED` (module-level then within-module), prints the seed, and on failure prints the reproduction command. `randomized-tests.yml` runs nightly and records the seed as an artifact before anything can fail. `just test-random [seed]` reproduces locally.

**Incident → regression test (item 5).** `tests/regressions/` with the convention `test_issue_<N>_<what_must_not_recur>.py`, enforced by `tests/test_regression_convention.py`, seeded with #858: the test runs the desktop contract's real `require_glob` against a tree with no Plasma session and holds that it exits 1 naming the path — and that the hummingbird waiver is the only way through. Policy text in CONTRIBUTING.md and AGENTS.md.

**Advisory nudges (item 11).** `.github/scripts/pr-nudges.sh` + `pr-nudges.yml`, modelled on hive's changelog-reminder: reminders derived from touched paths. Always exits 0; job summary + `::notice::` first, PR comment on top and tolerated. `tests/bats/test_pr_nudges.bats` runs it against a stubbed `gh` that 403s (kubestellar/hive#4440's regression test, ported).

**Contributor contract (item 14).** `just setup`, `test-fast`, `test-contract`, `test-random`, `test-cell <variant> <flavor>`, `test-e2e <iso>`, and `ci` = check + contract + test. Table in CONTRIBUTING.md.

**Graduation (item 8).** `graduation:` policy block (3 consecutive scheduled passes, ≤10% failures over 20 runs; `install` overrides to 2 because it is monthly). `scripts/graduation-check.py` measures every advisory criterion's gate workflow from `gh run list` and `graduation-check.yml` opens/refreshes one issue weekly when a criterion is ready. Advisory only — a human flips `enforcement`.

**Legacy cleanup.** Removed `prototype/iso-builder/` (a "moved" README), `iso_files/` (unreferenced since the tacklebox ISO path), `.github/workflows/archive/` (git history has them). `.github/copilot-instructions.md` still described the pre-July-2026 `base → dx → nvidia` flavor chain; it is now a pointer to AGENTS.md like `CLAUDE.md` and `.cursor/rules` (#2248). `docs/quality.md` pointed at `.claude/skills/…`, which is gitignored; the skill lives at `skills/`.

### Testing

Run on this branch at 7b60f805 (main merged in at 9c2c3828):

```
python3 -m pytest tests/ tests/pytest/ -v
================= 2 failed, 1055 passed in 1168.25s (0:19:28) ==================
FAILED tests/test_just_modules_resolve_repo_paths.py::...[utilities.just]   <- mine; fixed in the amended commit, 11 passed after
FAILED tests/test_package_parity_is_scheduled.py::test_audit_roster_derives_from_build_config
```

The parity failure is on `main` as-is: the test asserts `.github/build-config.yml` appears in `scripts/package-parity.sh`, and `git show origin/main:scripts/package-parity.sh | grep -c build-config.yml` is 0. Neither file is touched here.

New suites, run in isolation after the fix:

```
tests/test_ci_contract.py + tests/test_green_criteria.py       40 passed
tests/test_fork_safe_workflows.py                               26 passed
tests/test_graduation_check.py                                  10 passed
tests/regressions + tests/test_regression_convention.py          8 passed
tests/test_just_modules_resolve_repo_paths.py                   11 passed
bats tests/bats/test_pr_nudges.bats                             10/10 ok
scripts/check-ci-contract.py  -> CI contract: 10 criteria, every gate exists, is reachable, and meets its freshness SLA
TUNAOS_TEST_SEED=4242 pytest tests/test_ci_contract.py ...      40 passed, "order seed: TUNAOS_TEST_SEED=4242" printed
just --unstable --fmt --check -f Justfile / just/utilities.just  clean
yamllint -c .yamllint.yml on every new/changed workflow          clean (pre-existing line-length warnings elsewhere unchanged)
```

Full bats, before vs after this change: the same 10 files fail with the same 38 cases (1377 → 1399 ok, +10 new). Those failures are environmental on the box that ran them — it has the Python `yq` wrapper, not mikefarah's `yq` the scripts call — and are untouched by this PR.

### Related issues

Part of #2250; the remaining items are #2256–#2264.

---

hanthor:
Two red checks on 7b60f805, one now fixed here and one that is main's.

**build (live-iso-bootc, `just iso skipjack gnome`)**: root-caused and fixed in 5c9cb514. Since 886444e3 landed on main, `_start_bus` in `customize-live.sh` runs `dbus-daemon --fork --pidfile=...`, and dbus-daemon has no `--pidfile` option (only `--nopidfile`; the path is config syntax). Usage error, exit 1, on a discarded stderr, so the session bus never starts and the customize dies under `set -eu`. Every run of that workflow since 2026-08-25 ended the same way (32875192246, 33594824101); the last green run, 32829777966, is the commit before. The suite stayed green because the fake dbus-daemon in the fails-fast test honoured the option the real binary rejects. Details in #2265. The fix uses `--nopidfile --print-pid=3 3>"$_pidfile"`, keeps the reaper and the `/dev/null` redirection, makes the fake reject `--pidfile` like the real daemon, and adds `tests/regressions/test_issue_2265_*.py`, which runs the script's own helper against the real dbus-daemon. Verified locally with dbus 1.14.10; CI is re-running on the push.

**Unit Tests**: the single failure is `tests/test_package_parity_is_scheduled.py::test_audit_roster_derives_from_build_config` (1056 passed otherwise). It fails identically on main: `git show origin/main:scripts/package-parity.sh | grep -c build-config.yml` is 0, and this PR does not touch that script or test. Leaving it for whoever owns the parity roster; will re-check when main is green.

---

Issue #2083, opened 2026-08-26
Bump REMORA_VERSION to v0.3.0 — images ship a remora whose rebuilds can silently fail to stage


hanthor:
`build_scripts/install-remora.sh` pins `REMORA_VERSION="v0.2.0"`. [remora v0.3.0](https://github.com/tuna-os/remora/releases/tag/v0.3.0) shipped 2026-08-26 and fixes a defect that affects every image carrying the v0.2.0 binary.

> **Resolved by #2087.** The pin is now `v0.4.0` (not v0.3.0 — see below), it carries a `# renovate:` marker so Renovate keeps it current, and `scripts/check-download-checksums.py` guards the checksum half Renovate cannot maintain. That guard also caught a live breakage while being written: the `chezmoi_sha256` entries were stranded on v2.71.0 after a Renovate bump to v2.72.0, failing the Ubuntu/niri build. Both are fixed.
>
> v0.4.0 rather than v0.3.0 because v0.3.0 carried a second defect of its own: scheduled builds never regenerated the Containerfile after moving the base pin, so the base never updated on the timer path. v0.4.0 is the first release with both that and the bare-tag rebase fixed.

## Why this matters (not the reason tuna-os/remora#21 gives)

remora#21 frames the stale pin as *"every TunaOS image ships the vulnerable install path"*, referring to the install-snippet hardening in remora#19/#20. **That framing does not apply to this repo.** remora#19/#20 changed remora's **README** snippet — the copy-paste instructions for manual installs, which had no `set -e` and ran `sudo install` even after `sha256sum` reported FAILED. `install-remora.sh` is a separate implementation that already does `set -xeuo pipefail` plus `sha256sum --check --strict`. It was never vulnerable, and bumping the pin delivers nothing on that front.

The actual reason to bump is a functional defect in v0.2.0. Its generated quadlet rebases with:

```
ExecStartPost=/usr/bin/bootc switch --quiet --transport=containers-storage localhost/remora:latest
```

That hands bootc the **same image specification on every run**. A scheduled rebuild that produces new content under an unchanged tag risks never being staged — so a user's layered packages can quietly stop updating, with a green timer and no error. v0.3.0 resolves the freshly built digest and switches to `tag@sha256:...` instead.

Also in v0.3.0, relevant to anyone using remora on a non-Fedora variant:

[...]

PR #2087, opened 2026-08-26
fix: let Renovate bump the remora pin, and guard the checksums it cannot

hanthor:
Closes the gap behind #2083, and fixes a live breakage found while doing it.

## The chezmoi pins are stale on `main` right now

Renovate bumped `chezmoi: "v2.71.0"` → `"v2.72.0"` and left `chezmoi_sha256` untouched. `build_scripts/desktop/niri.sh` therefore downloads `chezmoi_2.72.0_linux_amd64.deb` and checks it against **v2.71.0's** hash, so `sha256sum -c -` fails and the Ubuntu/niri build dies.

Verified against upstream rather than inferred:

| | |
|---|---|
| pinned amd64 | `a8880a1b…` |
| in `chezmoi_2.72.0_checksums.txt` | **absent entirely** |
| what `a8880a1b…` actually is | `chezmoi_2.71.0_linux_amd64.deb` |

It fails closed, so nothing bad shipped — but the build breaks and nothing named the cause. Both hashes are corrected here.

## Why it happened, and why it will happen again without a guard

Both custom managers in `renovate.json` capture a **single version line**:

```
"#\\s*renovate:\\s*datasource=(?<datasource>[^\\s]+)\\s+depName=(?<depName>[^\\s]+)\\n\\s+[\\w-]+:\\s*\"(?<currentValue>[^\"]+)\""
```

A sibling `_sha256` map falls outside that match. Renovate maintains the version and *cannot* maintain the checksums. `postUpgradeTasks` would solve it, but this repo uses the Mend-hosted app where that isn't available.

This is exactly why `install-remora.sh` deliberately carried **no** `# renovate:` marker — the comment there is correct. But the cost was that nothing bumped it at all: it sat on v0.2.0 for six weeks, which is #2083.

## The fix: guard the half Renovate can't maintain

`scripts/check-download-checksums.py` fetches the `checksums.txt` published for the **pinned** version and asserts every pinned hash matches. That catches both:

- a version bump that left checksums behind (the Renovate case), and
- a release re-published under the same tag with different content, which nothing checked before.

With that guard, the `# renovate:` marker is safe to add — so remora now bumps automatically, and a bump with stale checksums **cannot merge itself**: `automerge: true` covers patch updates and `platformAutomerge` waits on checks. `--fix` rewrites them, so correcting a bump is one command instead of a hunt through release pages.

Follows the existing `check-*-pins` convention: a script that names the problem in one line, driven by a small workflow (PR-triggered to gate, plus nightly for the re-publish case).

## remora v0.2.0 → v0.4.0

Not the reason #2083's parent issue gives — that framing conflated remora's README snippet with this script, which was never vulnerable (see tuna-os/remora#32). The real reason:

- **v0.2.0** rebases to a bare tag, handing bootc the same image specification every run, so a rebuild with new content under an unchanged tag can fail to stage — layered packages quietly stop updating behind a green timer.
- **v0.3.0** fixed that but carried a second defect: scheduled builds never regenerated the Containerfile after moving the base pin, so the base never updated on the timer path.
- **v0.4.0** is the first release with both fixed.

## Testing

- Script run before the fix: **caught the chezmoi bug**, remora green.
- `--fix` applied, re-run clean — all four pins match.
- Renovate's `build_scripts` manager regex verified to match the new marker, capturing `datasource=github-releases`, `depName=tuna-os/remora`, `currentValue=v0.4.0`.
- `actionlint` clean on the new workflow; `bash -n` clean on the modified script.
- Both remora checksums cross-checked two ways: the release's `checksums.txt` and the release API's per-asset `digest`.

---
_Generated by [Claude Code](https://claude.ai/code/session_0157X4TYXBuFmh7yAjsbNZr5)_

### DIFF
--- a/.github/workflows/check-download-checksums.yml
+++ b/.github/workflows/check-download-checksums.yml
@@ -1,0 +1,43 @@
+name: Check download checksums
+
+# Several downloads are pinned twice: a version, and a per-architecture sha256
+# of the asset at that version. Renovate maintains the version and cannot
+# maintain the checksums -- both custom managers in renovate.json capture a
+# single version line, so a sibling `_sha256` map is outside the match.
+#
+# That gap has already bitten: Renovate bumped chezmoi v2.71.0 -> v2.72.0 and
+# left chezmoi_sha256 on v2.71.0's hashes, so the Ubuntu/niri build downloaded
+# a v2.72.0 .deb and checked it against a v2.71.0 hash. It fails closed, so
+# nothing bad shipped -- but the build died and nothing named the cause.
+#
+# Running on pull requests is the point: `automerge: true` covers patch
+# updates, and platformAutomerge waits on checks, so a Renovate bump that
+# leaves checksums behind cannot merge itself. The schedule catches the other
+# case -- a release re-published under the same tag with different content.
+on:
+  pull_request:
+    paths:
+      - 'image-versions.yaml'
+      - 'build_scripts/install-remora.sh'
+      - 'scripts/check-download-checksums.py'
+      - '.github/workflows/check-download-checksums.yml'
+  schedule:
+    - cron: '35 22 * * *'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: check-download-checksums-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  checksums:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: Verify every pinned checksum matches its pinned release
+        run: python3 scripts/check-download-checksums.py
--- a/build_scripts/install-remora.sh
+++ b/build_scripts/install-remora.sh
@@ -14,18 +14,30 @@
 # The binary is static Go and works on every base (dnf/zypper/pacman/apt),
 # so there is nothing distro-specific to guard here.
 #
-# The version is pinned for reproducible image builds. Deliberately NO
-# `# renovate:` marker: the custom manager for build_scripts/*.sh would bump
-# REMORA_VERSION without touching the checksums below, and the sha256 check
-# would then fail every build.
+# The version is pinned for reproducible image builds, and the binary is
+# checked against a pinned sha256 so a re-published release cannot change what
+# lands in the image without someone noticing.
+#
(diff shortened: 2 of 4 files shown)


hanthor:
**`Unit Tests` is red here, and it is not this PR's failure.** Standing down on it rather than widening this PR — details so nobody has to re-derive them.

## What is failing

`Unit Tests` → step 8, **`Run BATS tests`**. Step 9 (`pytest`) passes: 927 tests green. Every other check on this PR is green, including the new `checksums` job this PR adds.

## Why it is not this PR's

`main` fails the identical step, with the identical shape (BATS red, pytest green):

| run | head | Test workflow |
|---|---|---|
| 2027 | `58ae644` (current `main`) | **failure**, step 8 |
| 2024 | `aa8dc37` | failure |
| 2009 | `62376277` | failure |
| this PR | `343ec83` | failure, step 8 |

Reproduced locally too: checked out `origin/main` detached and ran `bats tests/bats/test_build_scripts_remaining.bats` — test **74**, *"40-services.sh declares the privsep directory sshd cannot start without"*, fails there exactly as it does on this branch. This PR touches `build_scripts/install-remora.sh`, `image-versions.yaml`, and adds a script plus a workflow; it does not touch `40-services.sh`, sshd, or tmpfiles.

No open PR appears to be fixing it, so there is nothing to port into this branch.

## A caveat on the root cause

Locally the trigger is that **`ssh-keygen` is absent**. `tunaos_declare_sshd_privsep_dir` guards on it:

```
if ! ssh-keygen -q -t ed25519 -N '' -f "${keydir}/probe" >/dev/null 2>&1; then
    echo "WARNING: could not generate a probe host key; skipping privsep check" >&2
    return 0
fi
```

It returns 0 having written nothing, and the test then `cat`s a `tunaos-sshd-privsep.conf` that was never created. The test stubs `sshd` but not `ssh-keygen`, so it depends on a host binary it does not control.

**I have not confirmed this is the CI-side cause** — `ubuntu-latest` normally ships `ssh-keygen`, so the runner may be failing for a different reason inside the same step. I did not pull the full BATS log to confirm which test fails there. What is confirmed is that the same step is red on `main`.

## Proposed patch (not applied here)

Make the test own its dependency instead of inheriting it — stub `ssh-keygen` into `$bin` next to the existing `sshd` stub so the probe is hermetic, or `skip` when it is unavailable. Either keeps the property under test (the tmpfiles rule sshd asked for) while removing the environment coupling. Happy to open that as its own PR.

## On re-running

The failure is deterministic and reproduces on `main`, so a re-run would only re-observe it. `main`'s own CI run at `58ae644` is the stronger confirmation, and it is already recorded above — so I am not spending the re-run.

---

PR #2095, opened 2026-08-26
fix(desktops): the hummingbird repo must resolve on aarch64 too

hanthor:
`install-desktop.sh` writes a manifest's `baseurl` **verbatim** into `/etc/yum.repos.d/<name>.repo`:

```sh
echo "baseurl=${_TD_RB}"
```

dnf expands `$basearch` when it reads that file. But `gnome.yaml` and `cosmic.yaml` both hardcoded

```
https://repo.tunaos.org/hummingbird/20251124-x86_64/
```

so on aarch64 the desktop lane pointed at the **x86_64** prefix, which serves no aarch64 RPMs.

## Why this is quiet rather than loud

Those lanes install with `--skip-unavailable`. So the failure mode is not a red build — it is an image that builds **green and silently contains no desktop**. That is the exact shape `build_scripts/checks/verify-desktop-experience.sh` already records for `hummingbird:gnome`:

> the image carried 410 packages — gnome-backgrounds and gnome-user-docs, **no gnome-shell, no gdm, no mutter, no gtk4** — and this check called it passed

`build_scripts/10-base-packages.sh:167` already writes the same repo the right way, so the manifests were the odd ones out, not the pattern.

## Verified before changing anything

`$basearch` is only correct if both prefixes exist. Both are served:

```
20251124-x86_64  -> HTTP 200   repodata/repomd.xml
20251124-aarch64 -> HTTP 200   repodata/repomd.xml
```

## Two pins deliberately left alone

The first version of the test asserted that **no** desktop manifest may hardcode an arch. It failed two more, and checking them showed both are deliberate and documented — I would have broken working lanes:

- **`cosmic.yaml` el10 `tuna-os-fprintd`** → `fprintd/10-stream-aarch64/`. Its own comment explains only aarch64 RPMs are published there, and that x86_64 correctly falls back to CentOS Stream's `fprintd-pam`. `$basearch` would point x86_64 at a prefix that does not exist — and the repo writer sets `skip_if_unavailable=False`, so it would **hard-fail** the build rather than degrade.
- **`xfce.yaml` el10 `tunaos-xfce`** → `xfce/10-stream-x86_64/`, an x86_64-only lane.

So the test is scoped to the `hummingbird:` sections, where both arch prefixes demonstrably exist. A rule those two must violate would either red-line them or invite someone to "fix" them into a broken state.

## Tests

Mutation-verified:

- restoring either hardcoded URL fails the test for that manifest
- a separate test fails if the extraction ever finds no hummingbird repo at all, so the check cannot pass by examining nothing

## Context

This is one half of getting `hummingbird:gnome` to actually reach gdm. The other half is upstream: the package factory has not published since 2026-08-21 because a single package (`dconf`) fails the build cell, which exits non-zero and skips the publish — so `gnome-shell`, `gdm` and `mutter` are not in the repo for any arch to install. That is tuna-os/tunaos-packages#545. This change makes sure that once they are published, aarch64 can see them.

---
_Generated by [Claude Code](https://claude.ai/code/session_015sZp9iEZAJS44joA2Z8mCK)_

### DIFF
--- a/manifests/desktops/cosmic.yaml
+++ b/manifests/desktops/cosmic.yaml
@@ -181,7 +181,7 @@ packages:
   hummingbird:
     repos:
       - name: tunaos-hummingbird
-        baseurl: https://repo.tunaos.org/hummingbird/20251124-x86_64/
+        baseurl: https://repo.tunaos.org/hummingbird/20251124-$basearch/
         priority: 5
     packages:
       # See the fedora list's dconf note — needed here for real (the rebuild
--- a/manifests/desktops/gnome.yaml
+++ b/manifests/desktops/gnome.yaml
@@ -125,7 +125,7 @@ packages:
   hummingbird:
     repos:
       - name: tunaos-hummingbird
-        baseurl: https://repo.tunaos.org/hummingbird/20251124-x86_64/
+        baseurl: https://repo.tunaos.org/hummingbird/20251124-$basearch/
         priority: 5
     packages: *fedora_gnome_packages
 
--- a/tests/bats/test_hummingbird_desktop_wiring.bats
+++ b/tests/bats/test_hummingbird_desktop_wiring.bats
@@ -47,11 +47,17 @@ yq_bin() { command -v yq; }
 	[ "$status" -eq 0 ]
 }
 
+# `$basearch`, not a literal arch. install-desktop.sh writes this baseurl
+# VERBATIM into /etc/yum.repos.d, and dnf expands $basearch when it reads the
+# file. Pinned to x86_64 (as this assertion used to require), the aarch64
+# desktop lane pointed at the x86_64 prefix, which serves no aarch64 RPMs --
+# and because these lanes install with --skip-unavailable, that produced a
+# GREEN build carrying no desktop rather than a red one.
 @test "gnome.yaml declares a hummingbird repo at the measured target path" {
 	[ -n "$(yq_bin)" ] || skip "yq not available"
 	run yq -r '.packages.hummingbird.repos[0].baseurl' "$GNOME"
 	[ "$status" -eq 0 ]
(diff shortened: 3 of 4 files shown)


hanthor:
**Unit Tests was red for two reasons, one mine and one not. Fixed the mine; recording the other rather than folding it in.**

### Mine — fixed in `f4e178c`

Two BATS assertions required the literal `hummingbird/20251124-x86_64`, which is exactly the string this branch replaces with `$basearch`:

```
not ok 4 gnome.yaml declares a hummingbird repo at the measured target path
not ok 7 cosmic's hummingbird list is fedora's minus exactly cosmic-settings-daemon
```

They are the guards for this line, so they had to move with it. Both now assert `hummingbird/20251124-$basearch`, in **single** quotes — it's dnf's variable, expanded when dnf reads the written `.repo` file, and double quotes would let bash expand it to nothing here first.

Mutation-verified rather than assumed: restoring the hardcoded arch in `gnome.yaml` fails BATS test 4 *and* the pytest arch-neutrality guard, each on its own. The assertions pin the new behaviour; they don't merely tolerate it.

### Not mine — reproduces on `main`

The run also reported:

```
not ok 10 customize-live.sh: gives headless Flatpak an explicit session bus
```

and the full local suite finds two more the CI job didn't surface:

```
not ok 34 every zypper desktop installs the display manager it enables
not ok 74 40-services.sh declares the privsep directory sshd cannot start without
```

I checked these out on `main` at `5cf9f35` and ran them there rather than inferring it — **all three fail on main too**. This branch changes two manifest baseurls and adds one test file; none of it reaches `customize-live.sh`, the zypper desktop lane, or `40-services.sh`.

Leaving them alone: folding an unrelated repair into a two-line fix would widen this PR, and `pytest` itself was already green (941 passed).

---

Issue #2052, opened 2026-08-25
wahoo:xfce — BLOCKED upstream: ELN composes zero xfce* packages


hanthor:
Child of [#2048](https://github.com/tuna-os/tunaOS/issues/2048). **Do not start.** Filed because the blocker is the finding; no `build-config.yml` row should exist until it clears.

## Measurement

`dnf repoquery` against pinned `registry.fedoraproject.org/eln-bootc@sha256:1909ffb6…`, 2026-08-25, x86_64.

```
dnf -q repoquery 'xfce*'   →  0 packages
dnf -q group list          →  no xfce-desktop group
```

Individually confirmed absent: `xfce4-session`, `xfdesktop`, `xfwm4`, `xfce4-panel`, `thunar`, `xfce4-terminal`, `mousepad`, `ristretto`.

Against `xfce.yaml`'s `fedora:` list: **3 of 7 present**, and the four misses include the file manager and the terminal. The `fedora:` section also installs via `groups: [xfce-desktop]`, a group ELN does not define — so the group install would be a no-op and the flavor would produce an image tagged `xfce` containing no XFCE at all. That is precisely the #858 failure this project has already paid for.

## Why it is blocked, precisely

This is the hardest of the four blockers: **ELN simply does not compose XFCE.** It is not a repo we have not enabled, not a COPR gap, not a `--skip-unavailable` away. Zero packages.

Contrast with [#2051](https://github.com/tuna-os/tunaOS/issues/2051) (Niri), where the software is packaged for Fedora and only our *source* is unavailable. Here there is nothing in the distribution to point a manifest at.

This is the situation [VARIANT-LIFECYCLE.md](../blob/main/VARIANT-LIFECYCLE.md) §1 gate 3 describes with the Zorin counter-example ([#944](https://github.com/tuna-os/tunaOS/issues/944)): a base that does not publish the component. The gate's answer there was to not admit the variant. Same answer here, scoped to the flavor.

## Context worth knowing before anyone attempts a workaround

`xfce.el10` already pulls its whole stack from a tunaOS-built repo — `https://repo.tunaos.org/xfce/10-stream-x86_64/` — plus COPR `yselkowitz/cosmic-epel`. That repo is **EL10 x86_64 only** (it is why `yellowfin:xfce` is pinned off aarch64 in `build-config.yml`). It will not satisfy an EL11 buildroot.

[...]

Issue #2051, opened 2026-08-25
wahoo:niri — BLOCKED on packaging: every tunaOS path to Niri is a COPR, and COPRs do not build for ELN


hanthor:
Child of [#2048](https://github.com/tuna-os/tunaOS/issues/2048). **Do not start.** Filed because the blocker is the finding; no `build-config.yml` row should exist until it clears.

## Measurement

`dnf repoquery` against pinned `registry.fedoraproject.org/eln-bootc@sha256:1909ffb6…`, 2026-08-25, x86_64.

No niri/sway/wayland-WM group exists. Of the whole compositor stack, ELN ships **one** package:

```
swaybg          ← the only hit for niri* / waybar* / sway*
greetd-0.10.3-8.eln158   ← the display manager IS present
```

Absent: `niri`, `waybar`, `fuzzel`, `swaylock`, `swayidle`, `brightnessctl`, `playerctl`, `SwayNotificationCenter`, `papirus-icon-theme`, `quickshell-git`, `dms`, `dms-cli`, `dms-greeter`.

Against `niri.yaml`'s `fedora:` list (including its COPR packages): **16 of 29 present** — and the 13 misses are the entire desktop, not trimmings.

## Why it is blocked, precisely

This is *not* the XFCE problem ([#2052](https://github.com/tuna-os/tunaOS/issues/2052)) — Niri exists as software and is packaged for Fedora. The blocker is that **every tunaOS path to it runs through COPR**:

| manifest section | COPRs it depends on |
|---|---|
| `niri.fedora` | `yalter/niri-git`, `avengemedia/danklinux`, `avengemedia/dms-git` |
| `niri.el10` | the above plus `yselkowitz/wlroots-epel`, `yselkowitz/cosmic-epel` |

Two independent walls:

1. **COPR does not build for ELN.** COPR chroots target Fedora releases and EPEL; there is no ELN chroot to enable.
2. **A new `copr:` block is forbidden.** [PACKAGE-SOURCING.md](../blob/main/PACKAGE-SOURCING.md) puts COPRs in the `✗` tier — *"no exception without an allowlist entry"* — and `scripts/check-package-sources.py` fails CI on a newly added one. Adding these five to a new `eln:` section would be rejected at the gate, correctly.

So Niri on ELN is gated on the packaging effort, not on a manifest edit.

## What would unblock it

Either of:

[...]

Issue #2048, opened 2026-08-25
Wahoo (Fedora ELN): desktop expansion beyond GNOME — KDE, COSMIC, Niri, XFCE


hanthor:
Tracker for adding desktops to `wahoo`, the Fedora ELN / EL11 preview lane added in [#2042](https://github.com/tuna-os/tunaOS/pull/2042). That PR deliberately ships **base + GNOME only**: the other four were unmeasured, and declaring a desktop flavor that contains no desktop is the #858 shape.

They are measured now. All figures below come from `dnf repoquery` against the pinned base
`registry.fedoraproject.org/eln-bootc@sha256:1909ffb6…` on **2026-08-25**, x86_64. Reproduce with:

```
podman run --rm registry.fedoraproject.org/eln-bootc:latest \
  dnf -q repoquery --queryformat '%{name}-%{evr} (%{reponame})' <names>
```

## The result splits cleanly in two

| desktop | ELN group | core components | manifest pkgs present | verdict |
|---|---|---|---:|---|
| **COSMIC** | `cosmic-desktop` (9) | `cosmic-session` `cosmic-comp` `cosmic-panel` `cosmic-files` `cosmic-greeter` `xdg-desktop-portal-cosmic` — all **1.6.0**, `eln-extras` | 23/24 | **Ready to attempt** |
| **KDE** | `kde-desktop` (86) | `plasma-desktop` `plasma-workspace` `kwin` `kscreen` `plasma-nm` **6.7.4**, `sddm` 0.21.0, `dolphin` `konsole` 26.08 | 16/23 | **Ready to attempt** |
| **Niri** | — | `niri` `waybar` `fuzzel` `swaylock` **all absent** (only `swaybg` exists); `greetd` 0.10.3 present | 16/29 | **Blocked — packaging** |
| **XFCE** | — | **zero `xfce*` packages in ELN**, no `thunar` | 3/7 | **Blocked — upstream** |

The two "blocked" rows are not the same kind of blocked, and that distinction drives the sequencing:

- **XFCE** is blocked *upstream*: ELN does not compose XFCE at all. `dnf repoquery 'xfce*'` returns **0 packages**. Nothing tunaOS does to a manifest changes that. This is the same shape as the Zorin counter-example in [VARIANT-LIFECYCLE.md](../blob/main/VARIANT-LIFECYCLE.md) §1 gate 3 — a base that does not publish the component.
[...]

PR #2103, opened 2026-08-27
feat(wahoo): add COSMIC and KDE — every desktop ELN actually composes

hanthor:
Follows [#2042](https://github.com/tuna-os/tunaOS/pull/2042) (base + gnome). The four remaining desktops were measured in [#2048](https://github.com/tuna-os/tunaOS/issues/2048); **two need no new packages at all**, so they land here. The other two stay filed rather than declared.

## Measured, then actually depsolved

`dnf repoquery` against the pinned `eln-bootc` digest, 2026-08-26:

| flavor | packages | source |
|---|---|---|
| `cosmic` | **23/24** | COSMIC **1.6.0**, `eln-extras` |
| `kde` | **16/23** + the `kde-desktop` group (86 entries) | Plasma **6.7.4**, `eln-extras` |

Then — because "the names resolve" is weaker evidence than it looks — I ran a real `dnf install` of each new section against live ELN repos:

```
===== cosmic: depsolve test =====
  packages OK (23 names)
===== kde: depsolve test =====
  group OK: kde-desktop
  packages OK (16 names)
```

Both transactions resolved clean. That is a stronger check than the GNOME section shipped with.

## The COSMIC result is worth more than the flavor

`cosmic.el10` sources COSMIC from COPR `yselkowitz/cosmic-epel` (**twice**) and `fprintd` from `https://repo.tunaos.org/fprintd/10-stream-aarch64/` — both tier-3 migration inventory under [PACKAGE-SOURCING.md](../blob/main/PACKAGE-SOURCING.md).

On ELN **neither is needed**. The compose carries `cosmic-session`/`comp`/`panel`/`files`/`greeter`/`settings` and `xdg-desktop-portal-cosmic` natively, and `fprintd-1.94.5-6.eln158` resolves from `eln-appstream` — so even the aarch64 fprintd gap that pinned cosmic off arm64 on EL10 ([#732](https://github.com/tuna-os/tunaOS/issues/732)) does not reproduce here.

If EL11 keeps this, **that COPR dependency retires itself at the EL10 → EL11 transition** rather than needing a Tideforge migration. That is exactly the class of early signal the preview lane exists to produce, and it showed up before the lane has even booted.

Neither new section declares a `copr:` block. ELN needs none, and a new one would fail `scripts/check-package-sources.py` — correctly.

## Misses go under `optional:`, not into the bin

- **cosmic**: `cosmic-wallpapers`
- **kde**: `evtest`, `input-remapper`, `libratbag-ratbagd`, `openrgb-udev-rules`, `pamu2fcfg`, `solaar-udev`, `yubikey-manager`

All seven KDE ones are peripheral input / security-key extras — none is needed for a session to start. Listing them strictly would fail a build over a fingerprint helper; dropping them silently would be the [#1555](https://github.com/tuna-os/tunaOS/issues/1555) shape, where a `--skip-unavailable` shipped ten nightlies with no tailscale. `optional:` is neither: they install themselves the week ELN grows them, with no edit here.

## Niri and XFCE are deliberately absent

ELN ships **one** package of the Niri stack (`swaybg`) and **zero** `xfce*` packages. Adding either row would be [#858](https://github.com/tuna-os/tunaOS/issues/858) — an image tagged for a desktop that contains no desktop. They need packages built first: [#2051](https://github.com/tuna-os/tunaOS/issues/2051), [#2052](https://github.com/tuna-os/tunaOS/issues/2052), and issues now filed in `tunaos-packages`.

## LUKS denominator 51 → 53

Recorded in the test with its reasoning. This is the same distinction the `53 → 51` hummingbird drop directly above it drew: **a cell is only honest when a package set exists to pay it.** Hummingbird's kde/niri came out because they had 0 packages; wahoo's cosmic/kde go in because they have measured, depsolving ones.

The test now also asserts `matrix["wahoo"] == {"gnome", "cosmic", "kde"}`, so the absence of niri and xfce is pinned rather than incidental — re-declaring either puts its cell back and trips the assertion.

## One thing left for whoever reads a Gate failure first

Noted in `kde.yaml`: the `kde-desktop` group makes `plasma-login-manager` mandatory (Plasma 6.7 ships its own) while this manifest also installs `sddm` and declares `display_manager: sddm`. Both packages resolve. If a session ever comes up on the wrong greeter, that is the thing to look at — and it would be an ELN/Plasma-6.7 signal worth reporting upstream, not a tunaOS packaging bug.

## Testing

`pytest tests/` — **897 passed, 2 skipped**.

Not claimed: neither new flavor has been built in CI or booted. `Gate` is skipped on a dispatch, same as the existing wahoo flavors. A `build-wahoo.yml` dispatch after merge is what turns these into build evidence.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HvHdxf4BacZVvvMG35qkVG)_

### DIFF
--- a/.github/build-config.yml
+++ b/.github/build-config.yml
@@ -636,10 +636,18 @@ variants:
     # publishes no ELN branch). See docs/GREEN-MASTER-PLAN.md's wahoo section
     # and the TUNAOS_CODEC_GAP marker in verify-desktop-experience.sh.
     #
-    # MVP scope: base + GNOME only. ELN AppStream carries GNOME 51~beta;
-    # KDE/COSMIC/Niri/XFCE are not measured here and must not be declared
-    # before they are (#858: an image tagged for a desktop that contains no
-    # desktop). No ISO and no qcow2 — a preview lane publishes images to
+    # Scope: every desktop ELN actually composes, and no others. GNOME 51~beta
+    # and COSMIC 1.6.0 come from eln-appstream/eln-extras; KDE comes from the
+    # `kde-desktop` group (86 entries, Plasma 6.7.4). Each was repodata-measured
+    # before its row was added — #858 is an image tagged for a desktop that
+    # contains no desktop, and the guard against it is measuring first.
+    #
+    # Niri and XFCE are deliberately NOT here and must not be added by analogy:
+    # ELN ships one package of the Niri stack (`swaybg`) and ZERO `xfce*`
+    # packages at all. Both need packages built before a row is honest —
+    # tracked in #2051 / #2052 and in tunaos-packages.
+    #
+    # No ISO and no qcow2 on any flavor — a preview lane publishes images to
     # `bootc switch` onto, not installer media.
     flavors:
       - id: base
@@ -648,6 +656,23 @@ variants:
         build_iso: false
         build_qcow2: false
       - id: gnome
+        stage: 2
+        build_image: true
+        build_iso: false
+        build_qcow2: false
+      # COSMIC 1.6.0 from eln-extras — 23 of 24 manifest packages present
+      # (only cosmic-wallpapers absent). Notably needs NO copr and no
+      # repo.tunaos.org fprintd repo, both of which the el10 section carries:
+      # ELN has fprintd-1.94.5-6.eln158 in eln-appstream (#2049).
+      - id: cosmic
+        stage: 2
+        build_image: true
+        build_iso: false
+        build_qcow2: false
+      # Plasma 6.7.4 via the eln-extras `kde-desktop` group; 16 of 23 explicit
+      # packages present, the seven misses are peripheral input/security-key
+      # extras carried as `optional:` (#2050).
+      - id: kde
         stage: 2
         build_image: true
         build_iso: false
--- a/docs/MATRIX-STATUS.md
+++ b/docs/MATRIX-STATUS.md
@@ -83,7 +83,7 @@ Scored against `.github/green-criteria.yml`: a cell is green only when every **b
 | **marlin** | ✅ | ✅ | ❌ | ❌ | ✅ |
 | **sailfin** | ⬜ | ⬜ | ⬜ | ⬜ | ⬜ |
 | **skipjack** | ❌ | ⬜ | ❌ | ❌ | ✅ |
-| **wahoo** | ⬜ | — | — | — | — |
+| **wahoo** | ⬜ | ⬜ | ⬜ | — | — |
 | **yellowfin** | ✅ | ✅ | ❌ | ❌ | ✅ |
 
 Cells outside the desktop columns (base, hwe, nvidia and friends) are in the count above but not the table; only `builds` applies to them today.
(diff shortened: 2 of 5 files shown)


hanthor:
**`Unit Tests` is red on this head, and it is not this PR's failure.** Standing down on it, with the evidence and a proposed patch below rather than widening this PR.

## What is failing

pytest passed — **941 tests**. The job fails in the **BATS** step, on one assertion:

```
not ok 10 customize-live.sh: gives headless Flatpak an explicit session bus
# (in test file tests/bats/test_customize_live.bats, line 88)
#   `grep -q 'dbus-daemon --session' "${SCRIPT}"' failed
```

## Why it is not this PR's

This PR changes five files:

```
.github/build-config.yml   docs/MATRIX-STATUS.md
manifests/desktops/cosmic.yaml   manifests/desktops/kde.yaml
tests/test_matrix_status.py
```

Neither `tests/bats/test_customize_live.bats` nor `live-iso/common/src/customize-live.sh` is among them, and `git diff origin/main...HEAD` on those two paths is **empty** — they are byte-identical to main.

Confirmed by running rather than inferring: in a worktree checked out at pristine `origin/main`,

```
$ bats tests/bats/test_customize_live.bats
not ok 10 customize-live.sh: gives headless Flatpak an explicit session bus
```

Same failure, no part of this branch present. It is red on the base branch too, so every PR's `Unit Tests` job is currently red for this reason.

## Root cause: the test is stale, not the script

The assertion greps for the contiguous literal `dbus-daemon --session`. That spelling was deliberately removed when the bare binary stopped being safe to assume — the script's own comment records why:

> Modern desktop images ship dbus-broker as the bus implementation and no longer pull in the classic `dbus-daemon` binary … every cosmic flavor plus sailfin/grouper/guppy/marlin died here with "dbus-daemon: command not found" (exit 127)

The call was refactored into a helper, so the behaviour the test *names* is fully intact:

```bash
_start_bus() {
    local _pidfile="$1"; shift
    _live_bus_pidfiles+=("$_pidfile")
    dbus-daemon "$@" --fork --pidfile="$_pidfile" >/dev/null 2>&1
}
...
_start_bus /tmp/live-session-bus.pid --session --address="unix:path=${SESSION_BUS_SOCK}"
export DBUS_SESSION_BUS_ADDRESS="unix:path=${SESSION_BUS_SOCK}"
```

Headless Flatpak still gets an explicit session bus. Only the string the test pattern-matches on moved. This is an assertion pinned to a spelling rather than to behaviour — the same class the repo has been burned by before.

## Proposed patch (verified passing, deliberately NOT applied here)

```diff
   grep -q 'DBUS_SESSION_BUS_ADDRESS' "${SCRIPT}"
-  grep -q 'dbus-daemon --session' "${SCRIPT}"
+  # The session bus is started through the _start_bus helper, which execs
+  # `dbus-daemon "$@"`, so the literal `dbus-daemon --session` no longer
+  # appears contiguously — it was refactored into the helper when dbus-broker
+  # images made the bare binary unsafe to assume. Assert the two halves that
+  # actually carry the behaviour instead of the old spelling.
+  grep -q 'dbus-daemon "$@"' "${SCRIPT}"
+  grep -qE '_start_bus [^|]*--session' "${SCRIPT}"
```

Applied locally it turns the failure into `ok 10`. It is reverted here rather than committed: it belongs in its own small PR, not folded into a variant change where it would muddy review.

No re-run spent on this one — a `grep` against an unchanged file is deterministic and a second run would fail identically, so a 20-minute job would buy nothing.

Everything else on this head is green: `regenerate`, `Package source policy`, Lint ×5, `Lint Summary`, `fix`, CodeQL ×2.

---

PR #2238, opened 2026-08-31
fix(desktop): decouple Debian XFCE from sid GNOME

### LINKED ISSUE #1833: [flounder-sid] gnome and xfce builds wedged by sid's GNOME 50→51 transition: libgjs0 Breaks gnome-shell (< 51~beta~)
## What

flounder-sid `gnome` and `xfce` amd64 builds both fail in `install-desktop.sh` on the 2026-08-17 nightly ([31987439738](https://github.com/tuna-os/tunaOS/actions/runs/31987439738)), all three in-job attempts, with the same apt conflict:

```
libgjs0:amd64 Breaks gnome-shell (< 51~beta~)
gnome-shell:amd64 is available in version 50.3-1
E: Unable to satisfy dependencies. Reached two conflicting assignments
```

- **gnome** reaches it via `gnome-shell-extension-manager → gjs → libgjs0`, which forces the new gjs and the old gnome-shell into one unsatisfiable transaction.
- **xfce** reaches the identical wall via `gdm3 → gnome-shell → libgjs0` (xfce on sid pulls gdm3 as its DM).

This closes the "flounder-sid gnome/xfce — undiagnosed" box from docs/GREEN-MASTER-PLAN.md Phase 0: one classified upstream skew, two cells.

## Classification

This is Debian sid's **GNOME 50 → 51 transition** mid-flight — gjs has moved ahead of gnome-shell in the archive. It is the sid equivalent of the rawhide liboapv skew (#1810): upstream will heal it when gnome-shell 51 lands; nothing in this repo is wrong.

## Options

1. **Wait it out** — sid cells are expected to break during transitions; the composite scoreboard now renders them honestly as failing rather than silently absent.
2. If the outage drags: for **xfce**, the dependency on the GNOME stack via gdm3 is incidental — xfce-on-sid could use the same greetd/lightdm path other variants use, decoupling it from GNOME transitions entirely (likely worth doing regardless).
3. For **gnome**, `gnome-shell-extension-manager` could move to the best-effort tier so a transition can cost the extension manager rather than the whole desktop — same philosophy as `install_rawhide_tolerant`, and the wishlist gate keeps the omission loud.

hanthor:
Closes #1833

Debian sid's GNOME 50→51 transition can leave `libgjs0` ahead of `gnome-shell`. Keep the GNOME extension manager out of the Debian manifest's required transaction while the archive converges, and add a Debian-specific XFCE manifest that uses LightDM instead of pulling `gdm3` and the GNOME stack. This lets XFCE remain buildable independently of the transition while preserving the GNOME baseline.

Added a Bats contract covering the Debian XFCE override and the optional GNOME extension-manager package. Validation: shell syntax, manifest assertions, and `git diff --check`.

— hive: backend=pi model=openai-codex/gpt-5.6-luna
---
🐝 **Hive Agent**: `contributor` | **SHA:** `78798e13`

### DIFF
--- a/manifests/desktops/gnome-debian.yaml
+++ b/manifests/desktops/gnome-debian.yaml
@@ -13,7 +13,8 @@ packages:
     - gnome-core
     - gnome-disk-utility
     - gnome-system-monitor
-    - gnome-shell-extension-manager
+    # Optional during sid's GNOME 50→51 transition: libgjs0 can advance
+    # before gnome-shell, making this extension manager unsatisfiable.
     - nautilus
     - xdg-desktop-portal-gnome
     - xdg-desktop-portal-gtk
--- a/manifests/desktops/xfce-debian.yaml
+++ b/manifests/desktops/xfce-debian.yaml
@@ -1,0 +1,18 @@
+# XFCE Desktop Manifest — Debian override.
+#
+# Debian sid's gdm3 pulls the GNOME stack. During the GNOME 50→51
+# transition, sid's libgjs0 explicitly Breaks the available gnome-shell,
+# which makes an otherwise independent XFCE build impossible to resolve.
+# XFCE does not require GNOME or gdm3, so use LightDM on Debian bases.
+
+display_manager: lightdm
+
+packages:
+  apt:
+    - xfce4
+    - xfce4-goodies
+    - lightdm
+    - lightdm-gtk-greeter
+    - xdg-desktop-portal-gtk
+    - thunar
+    - mousepad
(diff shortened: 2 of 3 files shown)


hanthor:
The one red check here (`Unit Tests`) is the pre-existing `tests/test_package_parity_is_scheduled.py::test_audit_roster_derives_from_build_config` failure — the hardcoded roster in `scripts/package-parity.sh` is missing `bonito-rawhide`, `flounder-sid` and `gurnard`. It fails on `main` today and is unrelated to this change; #2268 is the open fix for it. Everything else is green, so merging.

---

PR #2341, opened 2026-09-05
cosmic, niri: install the software Vulkan driver these manifests never had

hanthor:
## What this is

Adds the software Vulkan driver (lavapipe) to the cosmic and niri manifests across every distro section.

**Read the correction below before reviewing this as a fix for the red cosmic cells — it is not one.**

## Correction: this does not fix the `boots` gate

I opened this claiming it addressed the seven failing `boots` cells. That was wrong, and the evidence was from the wrong harness.

The actual failing gate — `albacore:cosmic`, job [101038352430](https://github.com/tuna-os/tunaOS/actions/runs/33848074014) in nightly run 33848074014 — never ran a compositor. It never ran QEMU:

```
Failed to launch runner: failed to create fleet:
[{"ErrorCode":"VcpuLimitExceeded",
  "ErrorMessage":"You have requested more vCPU capacity than your current vCPU limit
   of 0 allows for the instance bucket that the specified instance type belongs to.",
  "InstanceRequirements":{"AllowedInstanceTypes":["g4dn*"] ...
```

`verify_boot` routes cosmic/niri/xfwl4 to `runs-on=<id>/runner=gpu` for a real DRM render node. The account's vCPU limit for the **g4dn bucket is 0**, so the fleet request fails, no runner ever attaches, and the job dies before its first step. That's why the boot-gate artifact is empty (`No files were found ... verify-out/serial.log`).

That is #2123. **No package fixes a quota of zero.**

## Where the original evidence actually came from

The zink/lavapipe symptoms are real — `MESA: error: ZINK: failed to choose pdev`, and run 32681262659 finding `renderD128` present with `[drm] features: -virgl`. But they come from the **installer smoke** path, which runs on hosted runners with no GPU. That is the `install` criterion (advisory), not `boots` (blocking). Conflating the two axes was the error.

## What the change is still worth

Software Vulkan is genuinely missing from these manifests on every distro, and it's what the hosted-runner smoke path needs to get GL through zink at all — as well as what anyone running cosmic or niri in a plain VM needs. So the packages stay, with an honest justification in the manifest comments.

**Untested.** This does not claim to fix the smoke path either; it has not been measured there. It is a defensible packaging gap being closed, nothing more.

## Package names (verified against live repositories)

| section | package |
|---|---|
| apt, fedora, hummingbird, eln, el10 | `mesa-vulkan-drivers` |
| zypper | `libvulkan_lvp` (openSUSE splits the ICDs; no `Mesa-vulkan-drivers` there) |
[...]

PR #2343, opened 2026-09-05
tests(cosmic): record mesa-vulkan-drivers as a measured apt name

hanthor:
## Fixes a red `main` that I caused

#2341 added `mesa-vulkan-drivers` to `cosmic.yaml`'s apt list. `test_cosmic_apt_package_names.bats` pins that list against the set measured on 2026-08-07, and the new name wasn't in it:

```
not ok 5 every apt package is one the PPA publishes, or greetd
FAIL: cosmic.yaml's apt list names 'mesa-vulkan-drivers', which neither
      ppa:hepp3n/cosmic-epoch nor the Ubuntu archive was measured to publish.
      apt fails the WHOLE install on an unknown name, not just that package.
```

**My miss, and the gap was in how I checked.** I ran five bats files by hand before #2341 and this wasn't one of them — including `test_declared_flavors_installable.bats`, whose *"COSMIC on apt … names packages that exist there"* passed and gave me false confidence that the apt side was covered. pytest was never the problem: 1338 passed on this same commit, in CI and locally. The red came from the BATS step alone.

## Why update the list rather than relax the check

The test is right to exist and right to be strict — apt fails the **whole** install on an unknown name, so a bad entry here breaks `grouper:cosmic` entirely rather than skipping one package. Its own comment says what to do:

> Update the list when the PPA gains a package — **after checking that it has.**

So this records a measurement instead of weakening an assertion.

## Checked, not assumed

`packages.ubuntu.com` returns HTTP 200 for `mesa-vulkan-drivers` on:

| suite | result |
|---|---|
| **resolute** (grouper's base) | 200 |
| questing | 200 |
| noble | 200 |

`grouper` is the only apt variant that declares `cosmic`, and its `base_image` is `ubuntu:resolute` — so the name resolves for the one image that actually installs it.

It's annotated like `greetd` immediately above it, which is in the list for exactly the same reason: present in Ubuntu's own archive rather than the PPA.

## Verification

`test_cosmic_apt_package_names.bats` 6/6, plus `test_declared_flavors_installable.bats` and `test_desktop_script_apt_branch.bats` — 12/12 together.

Kept deliberately minimal: one entry and its justification, nothing else, so it can land quickly and unblock `main`.

---

### DIFF
--- a/tests/bats/test_cosmic_apt_package_names.bats
+++ b/tests/bats/test_cosmic_apt_package_names.bats
@@ -88,6 +88,13 @@ setup() { command -v yq >/dev/null || skip "yq not available"; }
 		# Not from the PPA. Ubuntu's own archive ships greetd, which is
 		# why niri.sh installs it straight from there too.
 		greetd
+		# Also not from the PPA: Mesa's software Vulkan driver
+		# (lavapipe), added with the cosmic/niri Vulkan change. Measured
+		# 2026-09-05 against packages.ubuntu.com for grouper's own base
+		# and its neighbours -- resolute, questing and noble all return
+		# 200 for mesa-vulkan-drivers. grouper is the only apt variant
+		# that declares cosmic, and it is ubuntu:resolute.
+		mesa-vulkan-drivers
 	)
 	local p known fail=0
 	while read -r p; do


hanthor:
Adding the measurement that shows this fixes the regression and introduces nothing.

I ran the **full** bats suite twice — once on unmodified `main` in a separate git worktree, once on this branch — and diffed the failure sets:

```
main       bats exit: 1   →  39 failures
this branch bats exit: 1   →  38 failures

on main but NOT on this branch:
  "every apt package is one the PPA publishes, or greetd"     ← the fix

on this branch but NOT on main:
  (none)
```

Exactly one difference, in the right direction, and **zero new failures**.

## About the other 38

They are pre-existing on `main` and environmental to my container, not defects this PR touches. CI's bats step on `main` reported only the one real failure — the one fixed here. Several of the local 38 are self-evidently container artefacts, e.g.:

```
not ok  root check: detects non-root and errors
not ok  root_check: errors when not root
```

both of which assert non-root behaviour while my sandbox runs as root. I'm flagging them as observed-but-not-investigated rather than claiming they're all benign; the A/B is what establishes this PR's own effect, independent of why the rest fail here.

## On how this got missed in the first place

Worth recording, since the fix is small but the process failure wasn't. Before #2341 I ran five bats files by hand and treated that as coverage. One of them — `test_declared_flavors_installable.bats`, whose test reads *"COSMIC on apt is manifest-driven, and names packages that exist there"* — passed, which made the apt side feel checked. It is a different assertion from this file's, and the gap was invisible precisely because the name sounded right.

The suite is 1516 tests across 125 files; a hand-picked five is not a sample of it. Full-suite runs before pushing from here.

---

PR #2392, opened 2026-09-07
fix(live): present GNOME installer above startup overview

### LINKED ISSUE #1941: [installer-smoke] Walkthrough stalls on welcome: keys stop reaching the installer, focus ends up in GNOME Shell chrome
## Where the ISO axis now stands

Smoke run [32450214451](https://github.com/tuna-os/tunaOS/actions/runs/32450214451) is the first run where the readiness gate works, after #1937 fixed it to read the flatpak path the stamp actually lands on. **gnome, cosmic and kde all pass the assert in ~1 second** (it had never passed once in runs 63–68) and now fail one stage later, in the walkthrough.

This issue is about that next stage. It is a *new* frontier, not a regression.

| flavor | ISO build | readiness assert | walkthrough |
|---|---|---|---|
| gnome | ✅ | ✅ 1s | ❌ reaches `welcome` only |
| cosmic | ✅ | ✅ 1s | ❌ |
| kde | ✅ | ✅ 1s | ❌ |
| niri | ✅ | ❌ `org.tunaos.InstallerNiri` not running | — |
| xfce | ✅ | ❌ | — |

## The gnome walkthrough verdict

```
ok - gnome: captured at least 2 frames
ok - gnome: screen is not blank
  # 9/9 frames above stddev 0.02
ok - gnome: screen changes between steps
  # 1/8 transitions changed >500px; primary action activated with 'spc'
  # 2 distinct visual state(s) across 9 frames
ok - gnome: reached 'welcome' screen (Welcome)
not ok - gnome: reached 'disk' / 'encryption' / 'summary' / 'install' / 'done'
# Results: 4 passed, 5 failed — screens reached: welcome
```

## What the frames show

Frame sizes tell the story before the pixels do: `00/01/02` are ~56.6 kB, `03`–`08` are ~61.4 kB and essentially identical. **One transition, then six static frames.**

Comparing frame 00 with frame 03, the only thing that changes is **GNOME Shell chrome, not the installer**:

- frame 00 — the installer's own small `×` sits inside its header; the dock is idle
- frame 03 — a **large circular `×` overlay** appears above and right of it, and the dock shows a `bootc Installer` tooltip with the item highlighted

The installer window itself is byte-for-byte in the same place, at the same size, showing the same welcome content. **The "Install Yellowfin" row never shows a focus ring in any of the nine frames.**

So the harness's key events stop reaching the installer's content area, and something in the shell has taken focus or hover. The single counted "transition" is that chrome appearing — not the installer advancing.

## What I am not claiming

I can't tell from stills whether this is the Activities overview opening, focus tabbing into the dash, or a hover artefact. The window is *not* scaled or tiled, which argues against the overview, but that's inference rather than evidence. Guessing between those would produce a fix for the wrong one — the same trap that cost six runs on the readiness stamp, where an empty directory looked like proof of one thing and meant another.

Worth noting `installer-walkthrough.py` already carries scar tissue for two earlier versions of this:

> run 31183217981 opened the Credits modal and never left the welcome screen

> tab order is install -> bluetooth -> recovery -> poweroff -> credits, so sweeping outward walks toward Credits and away from Install

…so the tab-sweep starting point has already been tuned once. This is a third variant of "focus is not where the harness assumes".

## The diagnostic that would settle it

Capture **which widget actually has focus**, rather than inferring it from pixels. Options, cheapest first:

1. Log focus changes in the installer itself — a `notify::focus-widget` handler on the window writing to the debug log that #1934 already collects. That names the widget directly.
2. Query AT-SPI over SSH in the capture step.
3. Have the harness send `Escape` before each retry and record whether that restores its ability to advance — which would distinguish "a modal/overview is up" from "focus is merely lost".

Option 1 is the smallest and reuses plumbing that already exists and is already read.

## Also worth considering

Whatever the mechanism, it may be a real UX finding rather than only a test-harness one: a keyboard user opening this installer appears not to land on the primary action. If the installer set initial focus on the "Install Yellowfin" row, both the harness and a real keyboard user would be better off.

There is also a latent hazard: the large `×` is a close control. A harness that keeps escalating activation keys while focus sits near it could close the installer rather than advance it.

Evidence: artifact `installer-smoke-yellowfin-gnome` on run 32450214451 (`walkthrough-gnome-00.png` … `-08.png`).

Related: #1937 (the gate fix that got us here), #1934 (the debug-log capture option 1 would reuse).

hanthor:
## Summary

- add a live-session GNOME Shell extension that closes Activities after the startup animation
- enable the extension before launching the installer without replacing the image's existing extension set
- keep installer startup working if the convenience extension cannot be enabled
- cover the extension behavior and launcher ordering in the live-installer Bats checks

Fixes #1941.

## Testing

- `npx --yes bats tests/bats/test_live_installer_launch.bats`
- `bash -n live-iso/common/src/desktop-gnome.sh`
- `git diff --check`

### DIFF
--- a/live-iso/common/src/desktop-gnome.sh
+++ b/live-iso/common/src/desktop-gnome.sh
@@ -42,6 +42,52 @@ EOF
 # matrix (added in #1039), and that is the workflow whose whole job is to
 # assert the installer process is running.
 #
+# GNOME deliberately starts a normal user session in Activities. An XDG
+# autostart application does not dismiss that overview, so the installer maps
+# behind Shell chrome and receives neither focus nor keyboard input. Install a
+# live-session-only extension which closes the overview once Shell's startup
+# animation completes. The launcher enables it in liveuser's volatile dconf
+# before starting the installer; using `gnome-extensions enable` appends to the
+# user's extension set instead of replacing the image's enabled extensions.
+_gnome_live_extension="tunaos-live-installer@tunaos.org"
+_gnome_live_extension_dir="/usr/share/gnome-shell/extensions/${_gnome_live_extension}"
+mkdir -p "${_gnome_live_extension_dir}"
+tee "${_gnome_live_extension_dir}/metadata.json" <<'METADATAEOF'
+{
+  "uuid": "tunaos-live-installer@tunaos.org",
+  "name": "TunaOS Live Installer",
+  "description": "Present the installer instead of the startup overview in the live session",
+  "shell-version": ["45", "46", "47", "48", "49", "50", "51"]
+}
+METADATAEOF
+tee "${_gnome_live_extension_dir}/extension.js" <<'EXTENSIONEOF'
+import * as Main from 'resource:///org/gnome/shell/ui/main.js';
+
+export default class TunaOSLiveInstallerExtension {
+    enable() {
+        // Hide now when the extension is enabled after startup, and again at
+        // startup-complete when it is enabled while Shell is still animating.
+        Main.layoutManager.connectObject(
+            'startup-complete', () => Main.overview.hide(), this);
+        Main.overview.hide();
+    }
+
+    disable() {
+        Main.layoutManager.disconnectObject(this);
+    }
+}
+EXTENSIONEOF
+
+install -d /usr/libexec
+tee /usr/libexec/tunaos-live-installer-gnome <<'LAUNCHEREOF'
+#!/usr/bin/env bash
+# Failure to load the convenience extension must not prevent installation.
+gnome-extensions enable tunaos-live-installer@tunaos.org ||
+    echo "tunaos-live-installer: could not suppress GNOME's startup overview" >&2
+exec flatpak run org.bootcinstaller.Installer
+LAUNCHEREOF
+chmod 0755 /usr/libexec/tunaos-live-installer-gnome
+
 # Deliberately no OnlyShowIn=, for the reason spelled out in desktop-cosmic.sh:
 # systemd-xdg-autostart-generator gates on systemd-xdg-autostart-condition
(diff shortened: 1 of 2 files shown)


hanthor:
Reviewed and merging — the mechanism is right and the reasoning ("an XDG autostart application does not dismiss that overview, so the installer maps behind Shell chrome and receives neither focus nor keyboard input") matches what I measured independently.

**One thing this does not cover, with evidence, so it isn't assumed closed.**

I ran the published `bonito` (gnome) ISO through the new `--published` harness mode (#2411) and OCR'd the frames. The session isn't sitting on the Activities overview — it's sitting on **`gnome-initial-setup`**:

```
frame 01:  Welcome!  /  Unspecified
frame 08:  Previous | Privacy | Next
           Location Services
           Allows apps to determine your geographical location
frame 16:  (still on Privacy)
```

Those are verbatim `gnome-initial-setup` pages. And it explains the input behaviour precisely: my `sendkey` presses advanced *initial-setup* from Welcome to Privacy, while the installer stayed behind — consistent with your "receives neither focus nor keyboard input", but with a different window on top.

`gnome-initial-setup` is present in the live image (`/usr/libexec/gnome-initial-setup`, `.desktop`, sysusers.conf) and **nothing in `live-iso/` or `build_scripts/` disables it** — `grep -rn initial-setup` over both is empty on current main, and this PR doesn't touch it either.

So after this merges, the overview is handled and initial-setup is still in front of the installer on gnome live media. Worth a follow-up; I haven't filed one so as not to duplicate if you'd rather fold it in here.

One caveat on my own evidence: live customisation is applied at boot rather than baked into `combined.rootfs.sfs`, so squashfs inspection can't confirm the runtime ordering — the OCR of the booted session is what I'm going on.

PR #2399, opened 2026-09-08
fix(live): detect virgl-less render nodes

### LINKED ISSUE #1945: [live] The software-GL generator's render-node test is false on virtio-gpu without virgl, so xfce/niri get no fallback
## The premise that no longer holds

`system_files/usr/lib/systemd/user-environment-generators/60-tunaos-software-gl` guards on the presence of a render node:

```bash
if compgen -G '/dev/dri/renderD*' >/dev/null 2>&1; then
	exit 0
fi
echo "LIBGL_ALWAYS_SOFTWARE=1"
echo "WLR_RENDERER=pixman"
echo "KWIN_COMPOSE=Q"
```

and states its reasoning inline:

> A render node is the honest test for 3D: virtio-gpu without virgl exposes `card*` but **no** `renderD*`.

**That is false on the current smoke runner.** From run [32465179023](https://github.com/tuna-os/tunaOS/actions/runs/32465179023)'s xfce leg, same boot:

```
[drm] pci: virtio-vga detected at 0000:00:02.0
[drm] features: -virgl +edid -resource_blob -host_visible
[drm] number of cap sets: 0
[drm] Initialized virtio_gpu 0.1.0 for 0000:00:02.0 on minor 1
```

```
$ ls -l /dev/dri/
crw-rw----+ 1 root video  226,   1 card1
crw-rw-rw-. 1 root render 226, 128 renderD128      <-- present
```

virgl absent, cap sets zero, **and `renderD128` exists anyway**. So the generator exits early, emits nothing, and the session gets no software-GL fallback while the hardware GL it inferred from the device node does not work.

## The consequence

```
##[error]xfce: no compositor process running (stuck at greeter or TTY?)
```

with `xfwl4` present and installed:

```
=== does the session command's compositor exist at all? ===
startxfce4       /usr/bin/startxfce4
xfwl4            /usr/bin/xfwl4
```

So this is **not** the missing-compositor hypothesis that was current — xfwl4 is installed and on PATH. It is a wlroots compositor being asked to use a render node that cannot render, with the fallback that exists for exactly this case switched off by a guard that misreads the situation.

This also retires a related assumption. `iso-e2e.sh` prints

```
==> GPU: -vga virtio headless (no render node/virgl) — niri/xfwl4 will not render here
```

which is half right: no virgl, but there **is** a render node. The two halves are not equivalent and the code treats them as if they were.

## Why the greeter works and the session does not

`build_scripts/desktop/greetd-gtkgreet.sh` exports the same variables **inline** in the session script rather than relying on the generator:

```bash
if ! compgen -G '/dev/dri/renderD*' >/dev/null 2>&1; then
	export WLR_RENDERER=pixman
	export LIBGL_ALWAYS_SOFTWARE=1
fi
```

It carries the same wrong guard, so on this hardware it is equally inert — but the greeter is `cage` + `gtkgreet`, a much lighter GL consumer than a full wlroots compositor, which is likely why the login screen renders and the session does not. That difference was previously read as evidence the generator was working.

## Options

1. **Widen the guard to "no render node OR a virtio-gpu with no virgl."** Strictly more conservative than today — it can only add fallbacks on machines where GL does not work, and real hardware never reports `-virgl`, so nothing changes there. The awkward part is detecting virgl from a user-environment generator: the clean signal is the kernel line `[drm] features: ... -virgl`, and generators may not be able to read the kernel log.
2. **Detect once at boot in a system unit** that writes a flag (e.g. `/run/tunaos-no-3d`), and have the generator and `greetd-gtkgreet.sh` both read it. More moving parts, but the detection runs as root where the kernel log is readable, and both consumers stop carrying their own copy of the test.
3. **Probe for working GL rather than inferring it** — the honest version of what the comment claims to do. Needs a small EGL probe in the image; `eglinfo` and `vulkaninfo` are both absent today (the capture step reports `command not found` for each).

(2) looks best to me: it fixes both call sites at once, puts the detection where it has the privileges to be accurate, and replaces two copies of a wrong heuristic with one measurement. But I have already inferred rather than measured once today (tuna-os/tunaos-packages#467), so I would rather this be chosen deliberately than by me guessing again.

## Scope

Affects any flavor whose compositor needs working GL on a virtio-gpu without virgl — xfce (`xfwl4`) and niri directly. cosmic-comp falls back on its own, which is why cosmic renders here; that asymmetry was measured back in #1087 and is unchanged.

Related: #1941 (the gnome overview stall, a different failure at a later stage), #1087 (which introduced the generator and the premise), run 32465179023 (evidence).

hanthor:
## Summary
- detect virtio-gpu's kernel-reported `-virgl` capability as root at boot instead of treating every render node as working 3D
- share the resulting runtime flag with both the user environment generator and the greetd/cage wrapper while retaining the no-render-node fallback
- correct the plain-QEMU diagnostic and add behavioral coverage for no-node, working-virgl, and render-node-without-virgl cases

Closes #1945

## Testing
- `git diff --check`
- `bash -n build_scripts/40-services.sh build_scripts/desktop/greetd-gtkgreet.sh scripts/iso-e2e.sh system_files/usr/libexec/tunaos/detect-software-gl system_files/usr/lib/systemd/user-environment-generators/60-tunaos-software-gl`
- manually exercised detector fixtures for absent render node, `+virgl`, `-virgl`, and unavailable kernel log
- manually exercised extracted greetd wrapper with no render node, a render node, and a render node plus detector flag

### DIFF
--- a/build_scripts/40-services.sh
+++ b/build_scripts/40-services.sh
@@ -706,6 +706,11 @@ safe_disable mcelog.service
 safe_enable tailscaled.service
 safe_enable uupd.timer
 
+# Resolve the software-rendering policy as root before greeters and user
+# environment generators start. The preset preserves this enablement when
+# systemd reapplies distro policy on first boot.
+systemctl enable tunaos-software-gl-detect.service
+
 # Base boot contract (green criterion 3, GREEN-MASTER-PLAN W3). Every image
 # gets the unit — on desktop images it is the multi-user complement to
 # tunaos-desktop-contract.service, on base images it is the only boot proof
--- a/build_scripts/desktop/greetd-gtkgreet.sh
+++ b/build_scripts/desktop/greetd-gtkgreet.sh
@@ -57,18 +57,16 @@ if command -v gtkgreet &>/dev/null && command -v cage &>/dev/null; then
 	# it. -s keeps VT switching available (without it a greeter crash locks
 	# you out of the machine entirely).
 	# cage is wlroots-based and needs a renderer. A VM without virgl has a DRM
-	# card (virtio-gpu) but NO render node, so GL initialisation fails, cage
-	# exits, greetd restarts it forever and never reaches active — a black
+	# card (virtio-gpu) and may still have a render node that cannot provide
+	# accelerated GL, so cage exits, greetd restarts it forever — a black
 	# screen with no login. That is not a CI artefact: it is GNOME Boxes,
 	# VirtualBox and every plain `-vga virtio` guest, which is exactly how
-	# most people will first try TunaOS. scripts/iso-e2e.sh:281 says the same
-	# thing about this runner ("no render node/virgl — niri/xfwl4 will not
-	# render here").
+	# most people will first try TunaOS. scripts/iso-e2e.sh identifies this as
+	# the virgl-unavailable path where guest software rendering is required.
 	#
-	# So pick the renderer at boot rather than baking one in: software
-	# (pixman) only when there is no render node, hardware GL everywhere else.
-	# A login screen is cheap to render, so the fallback costs nothing where
-	# it is used, and machines with a GPU are unaffected.
+	# So pick the renderer at boot rather than baking one in. The root detector
+	# records virtio-gpu's `-virgl` capability in /run; the no-render-node check
+	# remains a safe fallback. Hardware with working GL is unaffected.
 	install -Dm0755 /dev/stdin /usr/libexec/tunaos/greetd-session <<'SESSION_EOF'
 #!/usr/bin/env bash
 # Launch gtkgreet under cage on hardware that may have neither 3D nor a GPU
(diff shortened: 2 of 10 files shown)


hanthor:
Reviewed — **not merging yet.** Unit Tests is red, and it's three failures in this PR's own subject area rather than unrelated flake:

```
not ok 94 gpu: aarch64 virt machine gets virtio-gpu-pci, never -vga virtio
not ok 95 gpu: x86 pc machine keeps -vga virtio
not ok 96 gpu: virgl on aarch64 probes for and uses virtio-gpu-gl-pci
```

Worth checking these aren't the cascade from the `rawhide_rpmdb_probe` `rm -rf $PWD` bug (#2405, merged since this branch was pushed) — that one produced ~60 spurious failures via `getcwd: cannot access parent directories` and made unrelated suites look broken. These three don't have that shape, but a rebase onto current main would rule it out cheaply and is worth doing regardless.

The approach itself looks right to me, and I have adjacent evidence for it. Testing published ISOs through the new `--published` mode (#2411) on a headless host, `iso-e2e.sh` prints:

```
==> GPU: -vga virtio headless (no render node/virgl) — niri/xfwl4 will not render here
```

and on that host GNOME renders fine under llvmpipe while `gurnard-pantheon` came back `stddev=0` on every frame. That's precisely the "render node present but no virgl" middle case this PR adds coverage for — treating every render node as working 3D is exactly the wrong assumption, and the current message conflates "no node" with "no virgl".

(The pantheon black screen turned out to have a different root cause — it was detected as gnome and handed GDM autologin into a LightDM-only image, fixed in #2416 — so it isn't evidence *for* the virgl case. Mentioning it so it isn't miscounted as such.)

hanthor:
Rebased onto current main and fixed the three red `gpu:` tests — pushed to this branch.

**They weren't failing on GPU selection.** The test's own extractor, `gpu_block_file()`, terminated on the string `no render node/virgl` — and this PR rewords that message to `virgl unavailable`, correctly, since the old text conflated "no render node" with "node present but no virgl", which is the entire point of the change.

With the terminator no longer matching, the extractor pulled **3,396 lines instead of ~72**, ran on into the UEFI-firmware check, and failed with:

```
ERROR: UEFI firmware not found for  — install the architecture's OVMF/AAVMF package
```

So the PR was penalised for improving the message its test parsed. The extractor now terminates on `TUNAOS_CHECKPOINT_NO_GPU=` — code rather than prose. All three pass with your behaviour change intact:

```
ok 1 gpu: aarch64 virt machine gets virtio-gpu-pci, never -vga virtio
ok 2 gpu: x86 pc machine keeps -vga virtio
ok 3 gpu: virgl on aarch64 probes for and uses virtio-gpu-gl-pci
```

Also worth knowing: `test_missing_contract_collects_diagnostics_before_evidence_capture` is red on main right now for an unrelated slice-fragility bug (#2427 fixes it), so if you see that one it isn't this branch.

Approach still looks right to me. The message conflation is real — I hit it from the other side testing published ISOs, where the harness prints "no render node/virgl" on a host that renders GNOME fine under llvmpipe.

Issue #2326, opened 2026-09-03
[yellowfin] gnome boot Gate on main: gdm.service fails to start and dbus-broker.service crash-loops, contract marker never emitted


hanthor:
## What

In the `flavor=all` dispatch of Build Yellowfin on main `3cb8794b` (run 33728906010, 2026-09-03), every one of the 21 flavors passed Manifest (the #2314 fix at scale), and base, base-hwe, base-nvidia, gnome-asahi, kde, kde-hwe, kde-nvidia, xfce, xfce-hwe and xfce-nvidia promoted. The four GNOME cells that go through the boot Gate did not:

| cell | Gate | why |
|---|---|---|
| gnome | failure | `Boot and verify`: see below |
| gnome-hwe | failure | same |
| gnome-nvidia | failure | same |
| gnome-nvidia-hwe | failure | same |

`gnome / Gate` (job 100592259618) serial log, after the 900 s wait for the desktop contract marker:

```
==> Waiting up to 900s for the desktop contract marker...
ERROR: desktop contract marker was not emitted
[FAILED] Failed to start gdm.service - GNOME Display Manager.
[DEPEND] Dependency failed for tunaos-desktop-… verify TunaOS gnome desktop experience.
[FAILED] Failed to start dbus-broker.service - D-Bus System Message Bus.   (x6)
[  100.940087] sh[1321]: TUNAOS_LIVE_READY uptime=100.93
```

The guest reaches `TUNAOS_LIVE_READY` and networking, so the image boots; what does not come up is the system bus, and gdm (and the desktop contract unit that depends on it) fail behind it. kde (sddm) and xfce on the same base and the same run pass their Gates, so this is specific to the gnome payload or something it installs (a `/etc/dbus-1/system.d` policy, an SELinux label, a `dbus-broker` override) rather than to the EL10 base.

## What it is not

- Not the cosmic/niri reds in the same run: those six Gates failed at `Set up runner` on the runs-on (AWS) pool before any step ran, and no `verify-out/` was produced. Their failed jobs were re-run once (12:58Z) on that basis.
- Not the Attest SBOM reds: those are the Rekor 502 class (#1994 / #2282), non-blocking by design.
- Not #1820 (gnome-nvidia-hwe dconf database never compiled): that gate reached the contract; this one never gets a system bus.

## Next

[...]

PR #2427, opened 2026-09-08
fix(test): anchor the #2326 slice on the dispatch, not the first case header

hanthor:
## main is red

`test_missing_contract_collects_diagnostics_before_evidence_capture` fails on a clean checkout of `main`. I hit it as a "failure" on two of my own PRs before checking whether it was mine — it isn't.

## Cause

It sliced `iso-e2e.sh` as:

```python
SCRIPT.split('case "$MODE" in', 1)[1].split("\nready)", 1)[0]
```

That file contains **three** `case "$MODE" in` statements. Splitting on the first starts the region ~1600 lines above the dispatch, so the `.index()` calls below matched helper definitions instead of the `disk)` arm — and the test failed whenever anything shifted around them.

Both #2369 (`collect_disk_boot_diagnostics()`) and #2411 (the `published)` arm) did shift things, and **neither touched the ordering this test is about**.

## Fix

Anchor on the `disk)` arm of the **last** case statement — the dispatch, which is what the test means.

Verified it still fails for its real regression: moving `collect_disk_boot_diagnostics` after the paint capture trips `assert collect < paint`.

## Why this matters beyond the one test

A test whose subject is an ordering should fail when the ordering changes **and at no other time**. This one had drifted into failing on unrelated edits, which is the same shape as a check that always fails — it stops carrying information, and the next person reads the red as noise. That's the dead-gate pattern recorded in AGENTS.md, arriving from the other direction.

https://claude.ai/code/session_019rprz14ZYS9uCdd51ayuKD

### DIFF
--- a/tests/regressions/test_issue_2326_disk_gate_collects_the_failed_systemd_journal.py
+++ b/tests/regressions/test_issue_2326_disk_gate_collects_the_failed_systemd_journal.py
@@ -27,7 +27,14 @@ def test_disk_qemu_attaches_the_prepared_vsock_credentials():
 
 
 def test_missing_contract_collects_diagnostics_before_evidence_capture():
-    disk_mode = SCRIPT.split("case \"$MODE\" in", 1)[1].split("\nready)", 1)[0]
+    # Slice the DISPATCH arm, not "everything after the first case header".
+    # iso-e2e.sh contains three `case "$MODE" in` statements; splitting on the
+    # first one grabbed a region starting ~1600 lines above the dispatch, so
+    # the .index() calls below matched helper definitions instead of the arm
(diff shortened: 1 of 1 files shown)


hanthor:
Folded into #2419 and closing this.

These two were deadlocked: **#2427 failed the `fix` check** (because `just fix` produces churn on main, which #2419 fixes) and **#2419 failed Unit Tests** (because of the slice bug in `test_issue_2326`, which #2427 fixes). Each was red for exactly the problem the other solves, so neither could go green on its own.

#2419 now carries both commits and is green locally: 13 passing across the formatting contract, the waived-exemption test and the #2326 regression test, with `just fix` producing zero churn.

The diagnosis and verification from this PR are preserved in the cherry-picked commit.

PR #2451, opened 2026-09-10
fix(ci): exclude guppy:gnome from the build matrix until Gentoo ships GNOME 50

hanthor:
## Summary
- `Build Guppy` has failed on the `gnome` flavor every scheduled run since 2026-09-03 (last green: 2026-08-30), burning 1-4h of CI per day: `verify-desktop-experience.sh` enforces a GNOME >=50 floor (maintainer directive 2026-09-03), and `manifests/desktops/gnome.yaml` already documents that Gentoo's `::gentoo` tree tops out at 49.9 — this was known and expected the day the floor was set, but guppy's `gnome` flavor was never removed from the build matrix to match.
- Fixes: removes the `gnome` flavor entry for guppy in `.github/build-config.yml`, following the exact pattern already used for guppy's absent `niri`/`cosmic` flavors (tunaOS#923) — `build-variant.yml` builds every flavor `build-config.yml` declares with no `build_image` filter, so declaring-or-not is what actually gates the attempt.
- Updates `tests/test_matrix_status.py`'s LUKS-cell-count regression test (52 → 51) with the same dated-history convention the file already uses for prior denominator changes.
- Filed tunaOS#2450 to track re-adding guppy:gnome once Gentoo ships a GNOME >=50 source.

## Test plan
- [x] `python3 -m pytest tests/ -q -k "matrix or build_config or guppy"` — 114 passed
- [x] `python3 -m pytest tests/test_the_gnome_floor_is_enforced.py` — 9 passed (floor logic itself untouched)
- [x] `yamllint` clean on `.github/build-config.yml`
- [x] Dispatched `Build Guppy` against this branch: [run 34484699133](https://github.com/tuna-os/tunaOS/actions/runs/34484699133) — `generate_matrix` succeeds and the matrix no longer contains a `gnome` cell (confirmed via job list at dispatch time). Full run takes 1-4h; will update this PR with the final conclusion once it completes, but the structural fix (no more guaranteed-red gnome attempt) is already confirmed.

https://claude.ai/code/session_01YPy3UKQtRXxGzZZG3KXVEk


hanthor:
Picked this up to drive it green. Two things worth flagging.

### The `Unit Tests` failure was not this PR's

```
AssertionError: dms_qml is pinned to v1.6.1 but its checksum is aef0e90ad39bf85f676743da7bbe4c32758059d49342d82741d93ec414b88888;
the asset … hashes to 904ad2709a2075f10dbb089b42a9f4eb39d8f06cb2634fd4c95fa739ef9c4a2d
```

That is the placeholder DMS QML checksum, corrected on `main` in #2464 — this branch simply predated the fix. Merged `main` in (a merge commit, so anyone's checkout of this branch stays valid) rather than rebasing.

### But the PR was genuinely incomplete, in two asserted places

Dropping `guppy:gnome` takes guppy from 4 flavors to 3 and the matrix from 140 cells to 139. Two records of those numbers were left behind, and both are pinned by tests — so this could not have gone green as it stood:

```
test_nightly_schedule_registry.py::test_the_registry_flavor_counts_are_real
  registry says guppy has 4 flavors; build-config.yml declares 3

test_the_documented_cell_count_is_the_configured_one.py
  the guide says 140 cells; build-config.yml declares 139
  the README box in the guide's diagram still says ('140','140'), not 139
```

Fixed both: the nightly registry table in `build-variant.yml` (4 → 3), and `docs/DEVELOPER-GUIDE.md`'s heading, prose and diagram box (140 → 139). The registry comment already says *"If you add a variant or move a cron, keep this table true"* — removing a flavor is the same promise.

Local verification: full Python suite **0 failures**; STE **3048 findings, equal to `main`**.

### One thing for you to decide

**#2460 does the same thing.** It is the more thorough version — it carries a regression test (`test_issue_2450_guppy_excludes_gnome_below_version_floor.py`), the README row and the developer guide — and it fails on the *identical* two checks for the same stale-checksum reason. The two also touch the same files, so whichever lands second will conflict.

I drove this one rather than #2460 for a practical reason only: #2460 is from the `hanthor/tunaos-999` fork, which this session cannot push to. If you would rather land #2460, merge `main` into it and it should go green the same way — and this PR should then close as superseded. Your call; I have not touched #2460.

---

PR #2459, opened 2026-09-10
feat(ci): honour freshness SLA in composite green scoring

### LINKED ISSUE #2259: [hive] Freshness SLA honoured by the composite scorer (stale_is_not_green gets a number)
Follow-up from the #2250 review. `green-criteria.yml` has said `rule.stale_is_not_green: true` since 2026-08-17, but nothing defined *stale*. The first #2250 PR gave every criterion a `freshness_sla_days` and `tests/test_ci_contract.py` checks the gate's schedule fires at least that often. The scorer does not yet use it.

## What

`scripts/gen-matrix-status.py` `composite_section()` already records a date per (cell, axis) in `docs/matrix-provenance.json`. Make the composite verdict treat an axis whose evidence is older than `freshness_sla_days` as ⬜ (not green), and render the age in MATRIX-STATUS.

## Acceptance

- [ ] A cell whose only `desktop` verdict is 5 days old (SLA 2) is not composite-green.
- [ ] `tests/test_composite_green_scoring.py` gains the stale case.
- [ ] README line reports `stale: N` next to `never reached`.

hanthor:
## Summary

- Honour `freshness_sla_days` in `scripts/gen-matrix-status.py` `composite_section()`: treat evidence older than the criterion's SLA as untested (⬜) in composite green scoring.
- In rendered composite table, only link evidence URLs for passing verdicts that are not stale.
- Add test coverage in `tests/test_composite_green_scoring.py` asserting that a 5-day-old desktop verdict (SLA 2) is not composite-green, while fresh verdicts remain green.
- Report `(stale: N)` next to `never reached` in `.github/scripts/update-build-status.sh` and `README.md`.

Fixes #2259

— hive: backend=agy model=gemini-3.7-flash-high effort=low
---
🐝 **Hive Agent**: `contributor` | **SHA:** `9bee79e2`

### DIFF
--- a/.github/scripts/update-build-status.sh
+++ b/.github/scripts/update-build-status.sh
@@ -17,6 +17,7 @@ trap 'rm -f "$tmp_table" "$tmp_readme"' EXIT
 total_green=0
 total_cells=0
 total_unreached=0
+total_stale=0
 composite_green=0
 
 # The boots criterion's scope, read from the criteria file so this script and
@@ -130,6 +131,21 @@ while IFS=$'\t' read -r variant emoji; do
 	composite_green=$((composite_green + cgreen))
 	total_unreached=$((total_unreached + ${#unreached[@]}))
 
+	builds_sla=$(yq -r '.criteria[] | select(.id == "builds") | .freshness_sla_days // 2' .github/green-criteria.yml 2>/dev/null || echo 2)
+	builds_sla="${builds_sla:-2}"
+	today=$(date -u +%F)
+	run_age=0
+	if [[ -n "$run_date" && "$run_date" =~ ^[0-9]{4}-[0-9]{2}-[0-9]{2}$ ]]; then
+		run_epoch=$(date -u -d "$run_date" +%s 2>/dev/null || echo 0)
+		today_epoch=$(date -u -d "$today" +%s 2>/dev/null || echo 0)
+		if [[ "$run_epoch" -gt 0 && "$today_epoch" -ge "$run_epoch" ]]; then
+			run_age=$(( (today_epoch - run_epoch) / 86400 ))
+		fi
+	fi
+	if [[ "$run_age" -gt "$builds_sla" ]]; then
+		total_stale=$((total_stale + green))
+	fi
+
 	failing_text=$(join_or_dash "${failing[@]+"${failing[@]}"}")
 	unreached_text=$(join_or_dash "${unreached[@]+"${unreached[@]}"}")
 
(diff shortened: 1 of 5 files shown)


hanthor:
Shepherding pass — two things, one blocking and one substantive.

**The conflict is README-only, and the resolution is to drop the README hunk.**

```
$ git merge-tree --write-tree --name-only origin/main pr2459
README.md
CONFLICT (content): Merge conflict in README.md
```

That's the whole conflict surface — the other four files merge clean. The README hunk here is the *output* of the script this PR changes:

```
$ git diff $(git merge-base origin/main pr2459) pr2459 -- README.md
-…**86 never reached**; no job asserted the latter…
+…**86 never reached** (stale: 0); no job asserted the latter…
```

and `update-build-status.sh` emits that exact sentence from `${total_stale}`. `update-build-status.yml` rewrites the block on main, so committing a hand-rolled copy of it only collides with the next regeneration. Taking `main`'s README verbatim (`git checkout origin/main -- README.md` during the merge) loses nothing — the first run after merge re-emits the line with a real count.

**`total_stale` is counted against the wrong denominator.** The counter is incremented with `green`:

```bash
if [[ "$run_age" -gt "$builds_sla" ]]; then
    total_stale=$((total_stale + green))
fi
```

so it is the number of cells that **built green on a run older than the SLA** — a subset of *built*, not of *never reached*. But it renders inside the never-reached clause:

> The remainder has **4 failures** and **86 never reached** (stale: 0)

A reader parses that as "86 never reached, 0 of them stale", which is not what was measured and is not even a coherent quantity (a cell nothing asserted has no run date to age out). With a non-zero value the two numbers actively contradict each other. Either move it next to the `Built N/M` figure it belongs to, or spell the noun out — `(N built green on results older than the {SLA}-day SLA)`.

Two smaller notes while in there: `builds_sla` is re-read from `green-criteria.yml` with `yq` on every iteration of the per-variant loop though it is a constant, and `today` is recomputed with it — both belong above the loop, and hoisting them also removes the (narrow) chance of the date rolling over mid-run and giving two variants different ages.

Note I can't push these: this branch lives on `hanthor/tunaos-999`, which this session can't attach (`add_repo` refuses cross-owner adds), so the merge and the fix are yours to make.

---

hanthor:
Reviewed in a sweep of PRs open more than two days. This one has been open **8 days** with CI red.

Failing required check(s): `Unit Tests, regenerate`.

Requesting changes rather than closing — the change itself may well be fine, and nothing here is stale enough to discard. But it cannot merge in this state, and a red PR that nobody is looking at is indistinguishable from an abandoned one after a while.

If the failure is not caused by this PR — a broken base branch or shared CI problem — say so here and it should be treated as infrastructure rather than left against this change.

hanthor:
Ported onto current main in #2672 (this branch was 300+ commits behind; the only conflict was the generated README, which the next status refresh regenerates). Closing in favour of that PR.

---

Issue #2359, opened 2026-09-06
albacore:niri fails the branding gate on every arch: COPR dms-greeter 1.6 ships no QML, tunaos-packages dms for el10 is unpublished


hanthor:
## What happened

Run [34008659578](https://github.com/tuna-os/tunaOS/actions/runs/34008659578) (`main` @ `1d3782f0`, `workflow_dispatch` flavor `niri`) is the first albacore:niri build under the #2349 branding gate. All three image legs (`linux-amd64`, `linux-amd64-v2`, `linux-arm64`) fail in `install-desktop.sh niri` with:

```
  ok: greeter command dms-greeter present
  FAIL: greetd launches dms-greeter but /usr/share/quickshell/dms-greeter/DMSGreeter.qml is missing
  FAIL: no wallpaper daemon (swaybg/swww/wpaperd) and no dms — background will be blank
TUNAOS_BRANDING_NIRI_FAIL variant=albacore failures=2
```

The `iso:niri` jobs then fail at the provenance gate because no image was published, which is the correct consequence.

## Why

The dnf transactions did exactly what `manifests/desktops/niri.yaml` (el10 section) asks:

```
dnf -y --enablerepo=copr:copr.fedorainfracloud.org:avengemedia:danklinux install quickshell-git dms-greeter
  dms-greeter      x86_64  1:1.6.0-1.el10
  quickshell-git   x86_64  0.3.2^856.git2d3b3e9-1.el10.6
dnf -y --enablerepo=copr:copr.fedorainfracloud.org:avengemedia:dms-git install dms dms-cli
  dms               x86_64  2:0.0.git.4694.f9eb2381-1.el10   23 k
  dms-cli           x86_64  2:0.0.git.4694.f9eb2381-1.el10   20 M
```

None of those RPMs carries a `.qml` file. From the COPRs' `filelists.xml` for `epel-10-x86_64`:

- `dms-greeter` 1.6.0-1.el10: `/usr/bin/dms-greeter`, `/usr/lib/sysusers.d/dms-greeter.conf`, `/usr/lib/tmpfiles.d/dms-greeter.conf`, shell completions, docs, `/var/cache/dms-greeter`, `/var/lib/greeter`. No `/usr/share/quickshell/dms-greeter/`.
- `dms-cli` 0.0.git.4694: `/usr/bin/dms` plus completions.
- `dms` 0.0.git.4694: `dms.service`, three `.desktop` files, an icon, docs, licence.

[...]

PR #2464, opened 2026-09-11
fix(build): unbreak six variants' base builds — dead digests, the niri DMS payload, hummingbird arm64, and the boot gate that cannot launch

hanthor:
## Summary

Every variant's build workflow was red on `main`. I read the **Promote** job of each variant's newest conclusive nightly (Promote is what `gen-matrix-status.py` scores the `builds` axis from) and sorted the causes. Five are fixed here, each verified by a dispatched build on this branch; three are reported rather than worked around.

> Two findings below were made *by* the verification builds, after the first version of this PR: the rawhide fix was initially wrong, and the niri fix uncovered a second bug underneath it. Both are described where they happened rather than folded away.

### 1. Base image digests garbage-collected upstream

The pull fails before a single build step runs, so the variant's whole tree goes with it — base, every desktop layered on it, and every ISO built from those (bonito's nightly showed 13 `ISO …` jobs failing with `Artifact not found`, all downstream of this one line):

```
Error: creating build container: unable to copy from source docker://<ref>:
  reading manifest sha256:<pin> in <repo>: manifest unknown
```

| variant | image | pinned | fix |
|---|---|---|---|
| bonito | `quay.io/fedora/fedora-bootc:44` | `9c8fae9f…` → 404 | repinned `2c38c5f7…` |
| sailfin | `registry.opensuse.org/opensuse/tumbleweed:latest` | `1691585b…` → 404 | repinned `349094be…` |
| bonito-rawhide | `quay.io/fedora/fedora-bootc:rawhide` | `6c04d260…` → 404 | **unpinned — see below** |

Both repins were confirmed to be multi-arch indexes carrying amd64 and arm64 before they went in, so no variant quietly loses an architecture. Every other pin in `build-config.yml` was re-checked in the same pass and resolves.

**bonito-rawhide needed a different fix, and the build that tested the repin is what showed it.** The freshly-resolved digest was already gone 41 minutes later:

```
12:55  :rawhide -> sha256:0d942e1b…   pinned
13:36  :rawhide -> sha256:2caf5d4d…   and 0d942e1b… answers 404
13:36  :44      -> sha256:2c38c5f7…   still 200, hours later
```

Quay garbage-collects the untagged Rawhide manifest immediately, so repinning is a fix with a sub-hour half-life — and the repo's own safety net cannot close that gap either: `check-base-image-pins.yml` runs at 22:20 against builds that start at 01:00. `check-base-image-pins.sh` already names this variant in its own history (#1788) for exactly this reason. The tag is the honest reference for a variant whose purpose is to track a target that moves several times a day; a digest that cannot be fetched buys no reproducibility, it only guarantees a red base nightly. Nothing requires the digest form — the checker treats an unpinned base as a declared state and reports it per run (`SKIP (not digest-pinned)`), and `test_generate_workflows.py` already uses tag-only base images. It is deliberately the only unpinned base in the file, with the conditions for re-pinning written next to it.

This also explains something that had been hiding: bonito-rawhide's *desktops* were green while its base was red, because the desktops layer on previously published tags and never pull the base image.

### 2. niri's DMS payload had never reached an image

`install-dms-qml-fallback.sh` was called only from `build_scripts/desktop/niri.sh` — the **legacy** installer. Every variant that ships niri takes the manifest path (`manifests/desktops/niri.yaml` → `install-desktop.sh`), which never sources it. So the fallback ran on nothing, and the AvengeMedia EL10 `dms-greeter` 1.6 RPM ships no QML, failing two assertions in `verify-branding-niri.sh` that read the same directory:

```
FAIL: greetd launches dms-greeter but /usr/share/quickshell/dms-greeter/DMSGreeter.qml is missing
FAIL: no wallpaper daemon (swaybg/swww/wpaperd) and no dms — background will be blank
```

Identical on albacore, skipjack and yellowfin the same night. Now a `post_install` hook — the interface `install-desktop.sh` actually reads. It is *sourced*, so it runs the installer in its own shell rather than exiting the caller mid-install, and it is conditioned on the broken state (`dms-greeter` present, QML absent) rather than a distro list, so it no-ops once an RPM owns the path (#2359) and never touches variants whose greeter is `cosmic-greeter` or `gtkgreet` — hardcoding a DMS path is what reddened `marlin:niri` before.

**And the first build that ran it found the bug underneath: the pinned checksum had never matched the asset.**

```
sha256sum: WARNING: 1 computed checksum did NOT match
```

It could not have been noticed, because nothing had ever downloaded the tarball to compare. The value's own shape says as much — it ended in five 8s. Corrected from the checksum upstream publishes *next to* the asset rather than from a local hash of whatever arrived, and added to `scripts/check-download-checksums.py`, which covered remora and chezmoi but not this. A checksum no build verifies and no checker covers is not a pin, it is a comment. `Pin` grows an `arches` field for it, since this payload is one set of bytes for every architecture (its greeter is a shell script).

The existing bats case pinned the call site *in `niri.sh`* and stayed green through all of this. A second case now pins the path that builds.

### 3. hummingbird has no buildable arm64

The base stage's `xfsprogs` guard is mandatory (`bootc install to-disk` cannot format the root without it) and nothing on aarch64 satisfies it. Measured against the live indexes, not inferred (`repo.tunaos.org/hummingbird/20251124-$basearch/repodata`):

| arch | packages | `xfsprogs` | `malcontent` |
|---|---:|---:|---:|
| x86_64 | 13442 | 1 | 2 |
| aarch64 | 7177 | **0** | **0** |

The aarch64 wave is about half-published, which is also why `flatpak` is *broken* rather than absent there (nothing provides `libmalcontent-0.so.0`). The direction of travel agrees: `projectbluefin/utah-packages` is a single-manifest x86_64 OCI image with no manifest list, so a wider move onto it puts arm64 further out of reach. Dropped rather than left failing nightly, with both restore conditions written down and pinned by the honesty test.

### 4. The boot gate is no longer scheduled where it cannot launch

Maintainer decision. `verify_boot` routes cosmic/niri/xfwl4 to the RunsOn `gpu` group; the account's g4dn vCPU limit is 0, so EC2 rejects every fleet request and the job dies before its first step. Scheduling it measures nothing — no runner, no serial log, no compositor — and costs the cell its image on top, because `Promote` needs `verify_boot` to be success **or** skipped and a runner that never attaches is neither. That is why `builds` read not-green for cosmic and niri on ten variants whose images built, pushed and passed every other gate.

This does not make those cells green and does not claim they boot: `green-criteria.yml` already scored them out of scope for `boots`, unchanged here. It stops a runner shortage from being recorded as a build failure.

The guard is pinned to the routing by a new case in `test_the_boots_scope_matches_the_gpu_routing.py`, which already pinned the routing to the criteria file. Three lists, all equal, failing in **both** directions — a compositor routed to the GPU group but still scheduled loses its image to an unrunnable gate; one skipped but not routed silently stops being boot-gated at all. Verified by deleting the `xfwl4` negation in a scratch copy: the new case fails and names the difference. So when capacity returns, the negations come out in the same change that removes the routing — a partial revert is a test failure, not a silent hole.

### Not fixed here — these need a decision, not a patch

- **`flounder:gnome` (48) and `guppy:gnome` (49)** fail the GNOME 50 floor from the 2026-09-03 directive. Neither base has a GNOME 50 to install; the gate works as designed.
- **`yellowfin:gnome` and `skipjack:gnome`** fail the boot Gate itself (`SSH check failed: Connection timed out during banner exchange`, `desktop contract marker was not emitted`) — a real boot regression, distinct from the runner shortage above.
- **Sigstore rekor 502s** fail `Attest SBOM` across most variants. Not this repo's to fix, and already `continue-on-error: true`, so `Promote` is unaffected and `builds` does not see it.

## Testing

Every fix is verified by a dispatched build on this branch, judged on `Promote` — the job the `builds` axis is scored from.

| build | proves | result |
|---|---|---|
| [bonito:base](https://github.com/tuna-os/tunaOS/actions/runs/34602168260) | fedora-bootc repin | ✅ `Promote` success, both arches |
| [sailfin:base](https://github.com/tuna-os/tunaOS/actions/runs/34602171527) | tumbleweed repin | ✅ `Promote` success |
| [hummingbird:base](https://github.com/tuna-os/tunaOS/actions/runs/34602181055) | arm64 drop | ✅ `Promote` success, no arm64 leg scheduled |
| [marlin:cosmic](https://github.com/tuna-os/tunaOS/actions/runs/34603336165) | the gate change | ✅ whole run success — `Gate` skipped, `Promote` success |
| [yellowfin:niri](https://github.com/tuna-os/tunaOS/actions/runs/34604264857) | DMS payload + checksum, and the gate | ✅ all three arches, `Desktop contract` success, `Promote` success |
| [bonito-rawhide:base](https://github.com/tuna-os/tunaOS/actions/runs/34605966963) | the rawhide tag | arm64 ✅, amd64 in flight |

`marlin:cosmic` is the clean demonstration of #4: its last nightly had the four GPU `Gate` jobs as the *only* failing jobs, and the run is now green end to end.

Local, before each push:

- Full Python suite, every file: **0 failures**, including the two tests changed here.
- `bats test_niri_dms_payload.bats` 9/9, `shellcheck --severity=error` clean, both YAMLs parse.
- `scripts/check-base-image-pins.sh`: 13 pins resolve, 0 unresolvable, 22 declared platforms, 0 unsatisfiable.
- `scripts/check-download-checksums.py`: all pins match, and re-running it with the old DMS hash restored reports the mismatch and exits 1.
- STE: **3048 findings across 102 files — exactly `main`'s count**, so the ratchet is where it was.
- Full bats suite: 11 files fail identically with and without this change (verified against a stashed tree). Environment failures in this container — running as root, and a `yq` that is the jq wrapper rather than the Go binary.

## Related issues

Refs #2123 (g4dn capacity — revert #4 with it), #2359 (an EL10 RPM owning `DMSGreeter.qml`), #1397 (hummingbird flatpak), #1788 (the base-pin GC history).

### DIFF
--- a/.github/build-config.yml
+++ b/.github/build-config.yml
@@ -1120,7 +1120,40 @@ variants:
     emoji: "🐉"
     upstream_track: rolling
     description: "Based on Fedora Rawhide (rolling)"
-    base_image: "quay.io/fedora/fedora-bootc:rawhide@sha256:0d942e1b637ad5480cbdbdabb790060f894223a55ebedbc1729f97da8b5b1ab1"
+    # NOT digest-pinned, and this is the only base image here that is not.
+    #
+    # A digest pin on this tag is dead before the nightly that would use it.
+    # Measured 2026-09-11: the digest resolved at 12:55 was already 404 by
+    # 13:36 -- 41 minutes -- and `:rawhide` had moved on. It is not that the
+    # digest goes stale; quay garbage-collects the untagged manifest at once,
+    # so the pin does not resolve AT ALL:
+    #
+    #   Error: creating build container: ... reading manifest sha256:0d942e1b…
+    #   in quay.io/fedora/fedora-bootc: manifest unknown
+    #
+    # Compare fedora-bootc:44 in this same file, whose digest was still 200 in
+    # the same measurement. Stable tags are retained; Rawhide's are not.
+    #
+    # That makes repinning a fix with a one-hour half-life, and the repo's own
+    # safety net cannot close the gap either: check-base-image-pins.yml runs at
+    # 22:20 and the variant builds at 01:00, so the digest it verified is
+    # already gone by the time anything builds. scripts/check-base-image-pins.sh
+    # names this exact variant in its own history (#1788, 2026-08-16), because
+    # it keeps happening here and nowhere else for the same reason.
+    #
+    # So the tag is the honest reference. The point of bonito-rawhide is to
+    # track Rawhide, a target that moves several times a day; a digest that
+    # cannot be fetched buys no reproducibility, it only guarantees a red base
+    # every night and takes every downstream flavor with it. The pin checker
+    # already treats this as a declared state rather than an error, and says so
+    # per run:
+    #
+    #   SKIP  (not digest-pinned)  quay.io/fedora/fedora-bootc:rawhide
+    #
+    # Re-pin by digest only if quay starts retaining Rawhide manifests, or if
+    # something resolves the tag at build time and threads one digest through
+    # every stage of that build.
+    base_image: "quay.io/fedora/fedora-bootc:rawhide"
     # Friendly alias: same digest also published as ghcr.io/tuna-os/<alias>:<flavor>
     aliases: [fedora]
     akmods: "ublue-os"
--- a/.github/build-config.yml
+++ b/.github/build-config.yml
@@ -765,7 +765,7 @@ variants:
   - id: bonito
     emoji: "🎣"
     description: "Based on Fedora 44"
-    base_image: "quay.io/fedora/fedora-bootc:44@sha256:9c8fae9f47e7287d27542a6fe30ec408151655b142a61beea787c94654f1911a"
+    base_image: "quay.io/fedora/fedora-bootc:44@sha256:2c38c5f71a43993e07577629ef8e1695420ed38143482340215abdcea85a31fd"
     # Friendly alias: same digest also published as ghcr.io/tuna-os/<alias>:<flavor>
     aliases: [fedora]
     akmods: "ublue-os"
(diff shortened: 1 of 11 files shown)


hanthor:
Status, because this PR shows a lot of red check runs that are no longer live.

### The red checks on `8a5db79` are fixed, not outstanding

Five of them — `niri / linux-amd64`, `linux-amd64-v2`, `linux-arm64`, then `niri / Manifest` and both `iso:niri` legs cascading off them — are all one cause, and it was worth finding. The DMS QML hook worked: the build reached the fallback, downloaded the payload, and failed closed on its checksum.

```
dms-qml-payload: dms-greeter is installed but ships no QML; installing the pinned payload
sha256sum: WARNING: 1 computed checksum did NOT match
/tmp/tmp.5NcpUhiidJ/dms-qml.tar.gz: FAILED
```

The pinned hash had never matched the asset, and could not have been noticed: `install-dms-qml-fallback.sh`'s only caller was the legacy `niri.sh`, which the manifest build path never sources, so no build had ever downloaded the tarball to compare against. Fixed in c81d363, taking the value from the checksum upstream publishes next to the asset rather than from a local `sha256sum` of whatever arrived, and adding the pin to `scripts/check-download-checksums.py` so the nightly catches it independently of a build. Re-dispatched as [34604264857](https://github.com/tuna-os/tunaOS/actions/runs/34604264857).

### `Attest SBOM` is a Sigstore outage, and is not blocking

`🦈-sailfin / base / Attest SBOM` and its siblings fail on rekor, not on anything in this diff:

```
Error: ... Post "https://rekor.sigstore.dev/api/v1/log/entries": status 502: 502 Bad Gateway
::error::SIGSTORE_OUTAGE: attest still failing after 10m of Sigstore unavailability; giving up.
```

It has been failing across every variant's nightly since before this branch existed, there is no fix for it in this repository, and the job is already `continue-on-error: true` — so the run concludes **success** and `Promote` is unaffected. Sailfin's own run is the demonstration: `base / Promote` **success**, `Attest SBOM` failure, run conclusion success. `rerun-infra-failures.yml` already greps for that `SIGSTORE_OUTAGE` marker and re-runs these. Recording it here rather than re-running it, since the one re-run is better spent if something real appears.

### Where the verification builds stand

| build | result |
|---|---|
| `hummingbird:base` [34602181055](https://github.com/tuna-os/tunaOS/actions/runs/34602181055) | ✅ `Promote` success, and no `linux-arm64` job scheduled at all |
| `sailfin:base` [34602171527](https://github.com/tuna-os/tunaOS/actions/runs/34602171527) | ✅ `Promote` success — the openSUSE digest repin |
| `bonito:base` [34602168260](https://github.com/tuna-os/tunaOS/actions/runs/34602168260) | running — the fedora-bootc repin |
| `marlin:cosmic` [34603336165](https://github.com/tuna-os/tunaOS/actions/runs/34603336165) | running — `Gate` should read skipped and `Promote` should now run |
| `yellowfin:niri` [34604264857](https://github.com/tuna-os/tunaOS/actions/runs/34604264857) | running — re-dispatch after the checksum fix |

Two of the three fixes are already demonstrated end to end by a green `Promote`, which is the job `gen-matrix-status.py` scores the `builds` axis from.

---

PR #2473, opened 2026-09-12
fix(wahoo): record fastfetch as an ELN absence, not a strict install

hanthor:
## Summary

ELN retired `fastfetch` between two wahoo runs ten hours apart, and the early-warning lane caught it exactly as designed:

```
run 34657968072 (2026-09-11 23:24) — Installing fastfetch-0:2.68.1-1.eln159
run 34684438275 (2026-09-12 08:53) — No match for argument: fastfetch
```

Same `eln159` build tags on both. The strict transaction failed on both arches, base never published, and **all four wahoo cells went with it** — including `wahoo:cosmic`, which #2471 had just unblocked.

I confirmed this is a retirement and not a broken compose, by querying the live repos (`dl.fedoraproject.org/pub/eln/1`, AppStream+BaseOS+CRB+Extras, 12996 packages):

```
PRESENT podman, systemd, flatpak, upower    ← sanity check on the probe itself
absent  fastfetch
PRESENT glow, gum, distrobox                ← its neighbours in this very transaction
```

`fastfetch`'s list-neighbours still resolving is what rules out a compose outage. (Two earlier probe attempts of mine returned "absent" for *everything* including `kernel`; the sanity line is what caught them, which is why it's in there.)

So this moves the name into the block that already records `systemd-oomd`, `just` and `tailscale`, with the measurement attached.

It deliberately does **not** add `--skip-unavailable`. The same comment block says that is how #1555 shipped ten nightlies silently missing tailscale, and an early-warning lane that skips what it cannot find isn't one.

The Fedora and EL10 lists still install `fastfetch` — separate transactions against separate bases, and it resolves on both.

### The test

The absence block and the transaction are one statement, and nothing kept them in step. Both directions fail quietly:

- **listed absent, still installed** → the build dies every time on a package the file already documents as unavailable, and the next reader trusts the prose over the code.
- **removed, not documented** → a package leaves the lane with no record of why, which is the "silently skipped" outcome the strict list exists to prevent.

This PR is itself a two-place edit, which is the gap. Now, when ELN ships `fastfetch` again, moving the name back and deleting its entry become the same change — either half alone fails the test.

## Testing

```
$ pytest tests/test_the_eln_absences_match_the_eln_transaction.py -q
5 passed

# negative control — fastfetch put back in the transaction:
FAILED …::test_a_documented_absence_is_not_strictly_installed[fastfetch]

$ shellcheck --severity=error --exclude=SC1091 build_scripts/10-base-packages.sh
(clean)

$ bash -n build_scripts/10-base-packages.sh
$ bats tests/bats/test_10_base_packages.bats     # 26 tests, 0 failures
```

Parsed values, to show the selectors aren't matching nothing: documented absences `['systemd-oomd', 'just', 'tailscale', 'fastfetch']`, strict ELN transaction 37 packages.

## Related issues

Unblocks the verification of #2471 — `wahoo:cosmic`'s promotion is still untested, because the desktop flavors never got scheduled once base failed.

---

### DIFF
--- a/build_scripts/10-base-packages.sh
+++ b/build_scripts/10-base-packages.sh
@@ -296,8 +296,8 @@ elif [[ ${IS_ELN:-false} == true ]]; then
 	# Base set, strict. Every name here was verified present in
 	# eln-{baseos,appstream,crb,extras} on 2026-08-25; a miss is a real ELN
 	# regression worth failing the build on, which is the whole point of an
-	# early-warning lane. The four names the EL/Fedora lists carry that ELN
-	# does NOT ship are deliberately absent rather than silently skipped:
+	# early-warning lane. The names the EL/Fedora lists carry that ELN does
+	# NOT ship are deliberately absent rather than silently skipped:
 	#
 	#   systemd-oomd  — `dnf repoquery --whatprovides systemd-oomd` returns
 	#                   nothing (EL drops the subpackage); systemd-oomd-defaults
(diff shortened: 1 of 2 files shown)


hanthor:
Answering the docs nudge: user-visible in one small way, and documented in the absence block.

A wahoo image built after this no longer ships the `fastfetch` command. That is worth saying out loud rather than hiding behind "CI change" — but it isn't a tunaOS decision. ELN removed the package; the only choice in this PR is whether wahoo fails to build or builds without it, and the strict-list comment already settles that (the alternative, `--skip-unavailable`, is how #1555 shipped ten nightlies silently missing tailscale).

The record lives where the other three ELN absences live — the `does NOT ship` block in `10-base-packages.sh`, next to `systemd-oomd`, `just` and `tailscale` — with the two run IDs and the live-repo query that established it. That block is what a contributor reads when they wonder why a package the Fedora list carries isn't in the ELN one, and it's now what tells them to move `fastfetch` back the moment ELN ships it again.

One nearby doc mentions the name and is deliberately untouched: `docs/ci-troubleshooting.md:31` says to add `flatpak` "sorted alphabetically under `fastfetch`", but that refers to the **EL10** block, which still installs it. Only the ELN transaction changes here.

---

PR #2475, opened 2026-09-12
fix(wahoo): name gnome-keyring on ELN instead of inheriting it

hanthor:
## Summary

The desktop contract now fails `wahoo:gnome` at build time:

```
GNOME 51 meets the floor of 50
missing required path: none of [/usr/bin/gnome-keyring-daemon] exist
```

ELN used to pull `gnome-keyring` in transitively and stopped. The same image built clean at 23:24 on 2026-09-11 ([run 34657968072](https://github.com/tuna-os/tunaOS/actions/runs/34657968072), which got as far as its boot Gate) and failed all three attempts at 09:5x on 2026-09-12 ([run 34686896701](https://github.com/tuna-os/tunaOS/actions/runs/34686896701)) — GNOME went 50 → 51 between them, in the same compose churn that retired `fastfetch` the same morning (#2473).

So the gate is right: the manifest was relying on a `Requires` that no longer holds. This names the package instead.

Both names resolve on ELN, measured against the live repos (`dl.fedoraproject.org/pub/eln/1`, AppStream+BaseOS+CRB+Extras) rather than assumed — with a sanity check on `gdm`, `gnome-shell`, `nautilus` and `polkit` first, since two earlier probes of mine returned "absent" for everything.

`-pam` alongside it for the reason the other sections already give: without the PAM module the keyring isn't unlocked by the login password, so saved credentials and the SSH/GPG agents prompt separately on first use. `apt` lists `gnome-keyring` + `libpam-gnome-keyring`, `zypper` lists `gnome-keyring` + `gnome-keyring-pam`; this makes ELN say the same thing.

**`fedora`, `el10` and `hummingbird` are deliberately untouched.** They still get the package transitively and their contract checks pass. Changing a list on a measurement taken against a different base is how a guess ships — if the same `Requires` drops on those bases, their own contract check will catch it and the fix belongs in that change, with that measurement.

**No new test.** `verify-desktop-experience.sh`'s `require_any_glob '/usr/bin/gnome-keyring-daemon'` *is* the test, and it's what caught this. A unit test asserting the manifest names a package would only restate the manifest.

## Testing

```
$ pytest tests/ -k "manifest or desktop or wishlist or package"
207 passed, 5 skipped

$ pytest tests/ -k gnome
46 passed, 2 skipped
```

Confirmed the other package sections are byte-identical after the edit: `fedora` 52, `el10` 27, `hummingbird` 52 packages, none naming a keyring package; `apt` and `zypper` unchanged with theirs.

## Related issues

Third failure in one morning's ELN churn, after #2471 (ELN package gaps the wishlist gate found) and #2473 (`fastfetch` retired from ELN).

Context on what this does and doesn't unblock: `wahoo:base`, `:kde` and `:cosmic` all promote as of run 34686896701 — `:cosmic` is #2471 landing, verified. `:gnome` needs this to build at all, and then still has to clear its boot Gate, where `dbus-broker` fails for a reason #2472 should now record on the serial console.

---

### DIFF
--- a/manifests/desktops/gnome.yaml
+++ b/manifests/desktops/gnome.yaml
@@ -171,6 +171,28 @@ packages:
   # GNOME version on this base: 51~beta (gnome-shell 51~beta-3.eln158,
   # gdm 51~beta-1.eln158, mutter 51~beta-1.eln158), on BOTH x86_64 and
   # aarch64 — so no per-arch flavor pin is needed, unlike hummingbird.
+  # gnome-keyring is named here rather than left to whatever gdm or
+  # gnome-shell happen to Require. ELN used to pull it in transitively and
+  # stopped: wahoo:gnome built clean on 2026-09-11 23:24 (run 34657968072) and
+  # failed on 2026-09-12 09:5x (run 34686896701) with
+  #
+  #     GNOME 51 meets the floor of 50
+  #     missing required path: none of [/usr/bin/gnome-keyring-daemon] exist
+  #
+  # from verify-desktop-experience.sh, in the same compose churn that retired
+  # fastfetch the same morning. Both names resolve on ELN, measured against the
+  # live repos (dl.fedoraproject.org/pub/eln/1, AppStream+BaseOS+CRB+Extras).
+  #
+  # -pam alongside it for the reason the apt and zypper sections give: without
+  # the PAM module the keyring is not unlocked by the login password, so saved
+  # credentials and the SSH/GPG agents prompt separately on first use. apt lists
+  # gnome-keyring + libpam-gnome-keyring and zypper lists gnome-keyring +
+  # gnome-keyring-pam; this makes ELN say the same thing.
+  #
+  # The other dnf sections (fedora, el10, hummingbird) still get it
+  # transitively and are deliberately untouched -- their contract checks pass,
+  # and changing a list on a measurement taken against a different base is how
+  # a guess ships.
   eln:
     packages:
       - ModemManager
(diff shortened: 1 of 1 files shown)


hanthor:
Answering the desktop-change nudge — I checked all three, and one of them turned up a pre-existing inaccuracy that isn't mine to fix here.

**The desktop contract** (`verify-desktop-experience.sh`) already describes this correctly, and is in fact the thing that caught it: `require_any_glob '/usr/bin/gnome-keyring-daemon'` at lines 337 and 403. No change needed — the contract was right and the manifest was wrong.

**`docs/IMAGE-TAGS.md`** says nothing about keyring packages or wahoo's package set. Nothing to update.

**The README variant table** — this one is worth flagging. Line 45 reads:

> `| 🎏 **Wahoo** | Fedora ELN — EL11 preview (experimental, no codecs) | ghcr.io/tuna-os/wahoo | Base, GNOME | x86_64, arm64 |`

but `.github/build-config.yml` declares four flavors for wahoo, all `build_image: true`:

```
base    build_image=True
gnome   build_image=True
cosmic  build_image=True
kde     build_image=True
```

So the hand-written row is missing COSMIC and KDE. That is not a consequence of this PR — the row was already stale, and it won't self-correct, because the auto-generated block further down (line 88) is a different table that `update-build-status.yml` rewrites while line 45 is maintained by hand.

I'm leaving it out of this PR deliberately: it's unrelated to the manifest change, and a package fix and a docs correction in one diff is harder to review and harder to revert. Worth its own one-line change, and the durable version is probably a test asserting the hand-written flavor column matches `build-config.yml` — this is the second hand-maintained count in that file to drift (`140` → `139` in #2451 was the first).

On what this PR does change for a user: `wahoo:gnome` gets `gnome-keyring` and `gnome-keyring-pam` named explicitly. That restores what the image used to receive transitively rather than adding anything new, so nothing a user could observe changes relative to the last image that built.

---

Issue #1009, opened 2026-08-05
LUKS E2E: niri branding failure — missing greetd config and DMS greeter QML (bonito)


hanthor:
## Problem

Niri cells fail the branding-niri contract:

\`\`\`
FAIL: greetd references /etc/greetd/niri/config.kdl with -C, but that file DOES NOT EXIST
FAIL: missing /usr/share/quickshell/dms-greeter/DMSGreeter.qml
TUNAOS_BRANDING_NIRI_FAIL variant=bonito failures=2
\`\`\`

## Root Cause

The niri branding requires:
1. \`/etc/greetd/niri/config.kdl\` — the greetd niri session config
2. \`/usr/share/quickshell/dms-greeter/DMSGreeter.qml\` — the DMS (DankMaterialShell) greeter

These come from the zirconium payload (\`install-zirconium.sh\`), but bonito's niri stage either doesn't call it or the tarball doesn't include these files for Fedora.

Related: #637 (bonito/bonito-rawhide niri: DMS shell silently absent)

## Affected Cells

- bonito: niri
- Likely skipjack:niri, yellowfin:niri, albacore:niri once VERSION_CODENAME is fixed

## Branch

\`fix/marlin-luks-green\`

Issue #637, opened 2026-07-10
bonito/bonito-rawhide niri: DMS shell silently absent (avengemedia COPRs have no Fedora 44 build)


hanthor:
## Problem
On **bonito** (Fedora 44) and **bonito-rawhide**, the niri image builds green but ships **without the DankMaterialShell (DMS) desktop shell** — `dms`, `dms-cli`, `dms-greeter`, `quickshell-git` are installed best-effort (COPR block is `|| true`), and on Fedora 44 the avengemedia COPRs (`avengemedia/dms-git`, `avengemedia/danklinux`) have **no fedora-44 build** yet, so all four are silently skipped. Verified 2026-07-10 on quay.io/fedora/fedora-bootc:44: `dms`, `dms-cli`, `dms-greeter`, `quickshell-git` all `No match` even with both COPRs enabled.

This means bonito/bonito-rawhide niri is a non-functional niri (no shell) even though the build is "green", and likely underlies the niri daily-verification failures (#628).

## What "green + functional" needs
- Get the avengemedia COPRs to build for `fedora-44` / `fedora-rawhide`, **or**
- pin bonito niri's DMS packages to a Fedora version the COPR does build for, **or**
- make the DMS copr-block install non-best-effort (hard-fail) so the gap is loud instead of silent.

EL10 (yellowfin/albacore/skipjack) is unaffected — the COPRs build for epel-10 and DMS installs correctly there (fixed in the niri EL10 PR).

Related: #628, #629.

hanthor:
## Update 2026-08-08: the upstream blocker this issue names has just cleared — avengemedia's COPRs now build for fedora-44

Re-verified both claims in this issue directly against the COPR API/repodata (not just `dnf install` inference):

**1. The tunaOS-side bug (wrong package↔repo attribution, single failing `dnf` transaction) was already fixed** — `manifests/desktops/niri.yaml`'s `fedora:` copr block now installs `quickshell-git`, `dms`, `dms-cli`, `dms-greeter` in one transaction with both `avengemedia/danklinux` and `avengemedia/dms-git` enabled via `--enablerepo`, matching the working el10 pattern (landed in #1009, referenced in the file's own comment). No further manifest change needed for this part.

**2. The actual upstream gap — no fedora-44 build — is now closed**, as of TODAY:

```
$ curl -s https://copr.fedorainfracloud.org/api_3/project?ownername=avengemedia&projectname=dms-git | jq .chroot_repos
[...]

PR #2477, opened 2026-09-12
fix(ci): run the tests on the files the tests read

hanthor:
## Summary

`test.yml` is path-filtered, so a pull request matching nothing in the filter runs **no Unit Tests** — and shows every applicable check green, which reads exactly like a PR that was tested.

Measured this morning, not hypothesised: **#2475 changed one file, `manifests/desktops/gnome.yaml`, and merged on 13 green checks with `Unit Tests` absent** — while ten test files read `manifests/`. One of them, `tests/bats/test_niri_dms_payload.bats`, exists because a niri image once shipped, booted and published with its entire shell missing (#1009, #637). A manifest edit that broke that guard would have gone green here and failed on a nightly instead.

I ran the suite locally on #2475, so nothing shipped broken — but nothing in CI would have stopped it.

Reference counts across `tests/**/*.py` and `tests/**/*.bats`, which is how the additions were chosen rather than by guessing:

```
build_scripts 262   scripts 139   .github 101   live-iso 35
Containerfile* ~36  manifests 23  .github/build-config.yml 20
just 14  image-versions.yaml 11  system_files 11
registry-map.yaml 4  files 2
```

### The `.github` entry is the one worth reading twice

Six specific entries were listed — `scripts/**`, `changelogs.py`, `workflows/**`, `green-criteria.yml` — while **`.github/build-config.yml` was not**. That file *is* the matrix: what variants exist, what flavors each declares, which arches they build. Twenty test files read it, and a change to it alone ran none of them.

A partially covered directory is the exact shape of this bug, so those entries collapse into `.github/**`, which covers strictly more.

### What stays out, and why

`docs/**` is deliberately absent. The two docs files the suite reads more than once are `MATRIX-STATUS.md`, which the `regenerate` check owns end to end, and `LUKS-TPM.md`, which is prose under `ste`. Neither gains coverage, and `docs/**` would run the whole suite on every typo fix. `docs/CI_SPEC.md` keeps its own entry — the contract test reads it as a spec, not as prose.

### One existing test changed, and not to weaken it

`test_ci_contract.py::test_the_contract_test_runs_on_workflow_changes` required the **literal strings** `".github/workflows/**"` and `".github/green-criteria.yml"`, so widening the filter to `".github/**"` — which covers both, *plus* `build-config.yml` — failed it. A literal check cannot tell "broader" from "gone", so it reads a safer filter as a broken one.

It now asserts that a change to those paths **would match** the filter, which is strictly stronger than string equality, and ships with a guard proving the matcher can still say no.

## Testing

```
$ pytest tests/ -q      (minus four slow files, run separately — see note)
1440 passed, 16 skipped in 110.45s
```

Negative controls, so this isn't just "the check stopped firing":

```
manifests/** removed          → FAILED …test_a_path_the_suite_reads_also_triggers_the_suite[manifests]
.github/** → partial entries  → FAILED …test_a_path_the_suite_reads_also_triggers_the_suite[.github]

_filter_matches('.github/build-config.yml', ['.github/workflows/**'])  → False
_filter_matches('.github/build-config.yml', ['.github/**'])            → True
```

The four excluded files (`test_the_welcome_keywords_match_what_kde_says.py`, `test_installer_walkthrough.py`, `test_sign_step_retries_cosign.py`, `test_the_progress_screen_keywords_were_measured.py`) are merely slow, not failing — they take 8–10 minutes each locally and are unaffected by this diff.

## Related issues

Same class of defect as #2472: a guard that exists, passes, and does not run where it is needed.

Note this PR increases CI spend — `.github/**` and the build-input paths mean more PRs run the suite. That is the intended trade, but it is a real cost and your call if you'd rather narrow it.

---

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,18 +2,54 @@ name: Test
 on:
   pull_request:
     paths:
+      # Every path the suite asserts on, so that a change to one of them
+      # re-runs the tests that read it. tests/
+      # test_the_test_filter_covers_what_the_tests_read.py holds this list to
+      # that rule, because the gap is invisible from a pull request: a PR that
+      # matched nothing here showed every applicable check green and ran no
+      # Unit Tests at all, which reads exactly like a PR that was tested.
+      #
+      # That is measured, not theoretical. tunaOS#2475 changed one file,
+      # manifests/desktops/gnome.yaml, and merged on 13 green checks with Unit
+      # Tests absent -- while ten test files read manifests/. One of them,
+      # tests/bats/test_niri_dms_payload.bats, exists because a niri image once
+      # shipped, booted and published with its entire shell missing
+      # (tunaOS#1009, #637); a manifest edit that broke that guard would have
+      # gone green here and failed on a nightly instead.
+      #
+      # Reference counts across tests/**/*.py and tests/**/*.bats:
+      #   build_scripts 262   scripts 139   .github 101   live-iso 35
+      #   Containerfile* ~36  manifests 23  .github/build-config.yml 20
+      #   just 14  image-versions.yaml 11  system_files 11
+      #   registry-map.yaml 4  files 2
       - 'scripts/**'
       - 'tests/**'
       - 'build_scripts/**'
-      - '.github/scripts/**'
-      - '.github/changelogs.py'
-      # The CI contract (tests/test_ci_contract.py) proves every green gate
-      # exists and runs; it has to re-run when the gates themselves move.
-      - '.github/workflows/**'
-      - '.github/green-criteria.yml'
-      - 'docs/CI_SPEC.md'
+      # One entry rather than six. .github/scripts/**, changelogs.py,
+      # workflows/**, green-criteria.yml, runs-on.yml and actions/** were all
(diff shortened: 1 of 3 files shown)


hanthor:
Answering the workflow nudge: no green gate moved, so no `gates` block changes.

This PR changes *when* `test.yml` triggers, not what any gate asserts or which workflow asserts it. `.github/green-criteria.yml` has no entry naming `test.yml`, and `tests/test_ci_contract.py` passes (36 tests) — including the case this PR rewrote, which is the only reason the contract test is touched at all.

That rewrite is the one thing here worth a reviewer's attention, so to restate it in the open: `test_the_contract_test_runs_on_workflow_changes` asserted the literal strings `".github/workflows/**"` and `".github/green-criteria.yml"` were present in the filter. Widening the filter to `".github/**"` covers both of those *and* `.github/build-config.yml`, which twenty test files read and no entry matched — but the literal check failed anyway, because string equality cannot tell "broader" from "gone". It now asserts a change to those paths **would match** the filter, with a guard test proving the matcher still returns False for a filter that does not cover them.

Neither `docs/PIPELINE.md` nor `docs/build-pipeline.md` mentions `test.yml` or its path filter, so there is nothing to bring in step there.

---

PR #2481, opened 2026-09-12
fix(base): unpin Tumbleweed, repin Fedora 44 — both digests are 404

hanthor:
## Summary

Two dead base pins, **both mine from #2464**, both found by running the repo's own checker against the live registries:

```
FAIL  404  quay.io/fedora/fedora-bootc:44@sha256:2c38c5f7…
FAIL  404  registry.opensuse.org/opensuse/tumbleweed@sha256:349094be…
```

**sailfin has already paid.** Run [34720071388](https://github.com/tuna-os/tunaOS/actions/runs/34720071388) died on `reading manifest sha256:349094be… : manifest unknown`, taking all 7 cells — 6 of which had just gone green. That is the single biggest reason the composite count went **75 → 74** rather than up.

**bonito is next.** It built 14/16 at 14:09 today and its pin is 404 as of now.

### The guard exists and did not help

This is the part worth recording. `check-base-image-pins.yml` runs at 22:20, deliberately ahead of the 01:00 nightlies, and it caught #1788 (24 cells, three variants). Here:

```
2026-09-11 ~17:00  #2464 pins sha256:349094be… (verified alive at the time)
2026-09-12 00:22   pin check run 34661427002: success
2026-09-12 21:28   build-sailfin: manifest unknown
```

Under 28 hours from written to dead, and dead in the window *between* the check and the build it runs ahead of. **A daily check cannot protect a pin with a sub-day lifetime.**

### So the two bases are treated differently, on purpose

**Tumbleweed follows `bonito-rawhide` to a bare tag.** It is a rolling snapshot whose replaced manifests are collected. The checker reports it as a declared state rather than an error:

```
SKIP  (not digest-pinned)  registry.opensuse.org/opensuse/tumbleweed:latest
```

**Fedora 44 keeps a digest**, because it is a release tag rather than a rolling snapshot — repinned to what `:44` resolves to now, verified against the registry rather than inferred:

```
GET …/manifests/sha256:2c38c5f7…  -> 404
GET …/manifests/44                -> sha256:4a460ed0…
sha256:4a460ed0… is an OCI index over amd64 arm64 ppc64le s390x
```

If quay collects that one inside a day too, that's the second strike and bonito should follow the other two to a bare tag. I've written that rule into the file so the next person inherits the decision rather than the argument.

## Testing

```
$ scripts/check-base-image-pins.sh
(no FAIL lines; two declared SKIPs, arch checks ok)

$ pytest tests/ -k "base_image or build_config or arch_honesty or matrix"
117 passed, 2 skipped
```

## Related issues

Fixes the regression behind sailfin 6/7 → 0/7 in today's status refresh (#2479), and pre-empts the same failure on bonito's next nightly.

Same failure mode as #1788 and as the rawhide pin I replaced with a floating tag in #2464 — the third time this class of bug has cost this repo a variant, which is the argument for the bare tag rather than a fourth repin.

---


hanthor:
Answering the matrix nudge: no variant or flavor is added or removed, so none of the three follow-ups apply.

Checked rather than asserted — comparing the parsed config against `main`:

```
variants identical: True
flavors identical:  True
```

The whole diff is two `base_image` values and the comments recording why each is what it is. The matrix shape, the `scope` blocks in `.github/green-criteria.yml`, the README variant table and `docs/HARDWARE.md` all describe the same set of cells before and after.

`docs/MATRIX-STATUS.md` is untouched, as it should be — it is generated, and the `regenerate` check is running on this PR to prove the committed block still matches what the generator produces from this config.

What *does* change is whether those cells can build at all: sailfin's base is currently unfetchable, and bonito's is one nightly away from the same.

---

PR #2484, opened 2026-09-13
fix(gate): fail the boot contract when the system message bus is dead

hanthor:
## Summary

`boots` had a hole big enough to promote a machine with **no D-Bus at all**.

Measured on skipjack, [run 34705564876](https://github.com/tuna-os/tunaOS/actions/runs/34705564876), where *both* desktops hit the identical failure:

```
dbus-broker: Access denied in /etc/selinux/targeted/contexts/dbus_contexts +1
          -> dbus.socket: Failed with result 'service-start-limit-hit'

gnome   gdm hard-depends on dbus.socket
        -> Dependency failed for gdm.service     -> Gate RED
kde     emitted its desktop contract marker anyway
                                                 -> Gate GREEN, promoted
```

One broken image, two verdicts, and the green one shipped. Nothing disagreed, because the desktop contract measures the *desktop*, and the base contract deliberately tolerates `degraded` — where a dead bus hides on any image whose display manager doesn't hard-depend on it.

So the bus is now named explicitly, exactly as `bootc status` already is, and for the same reason the file gives: it isn't a per-variant unit-set opinion, it's whether the machine is operable. logind, polkit, NetworkManager, upower and every display manager reach the session through it.

The unit is **resolved, not hardcoded** — `dbus.service` is an alias, and which implementation it points at is precisely the per-variant detail this check must not care about. Hardcoding `dbus-broker` would pass a `dbus-daemon` image with a dead bus, so there's a test for that.

This is a real defect being reported, not a bar nothing can clear: **albacore, built by the same pipeline on a non-rolling EL10 base, starts the bus cleanly** ([run 34726078460](https://github.com/tuna-os/tunaOS/actions/runs/34726078460) — `dbus_contexts` denials 0, dbus-broker failures 0, `Started dbus-broker` 5).

## ⚠️ Expect cells to go red

Any cell whose image boots without a system bus now fails `boots` instead of promoting. `skipjack:kde` is the one already measured; there are likely others on yellowfin and skipjack.

**That is the point** — the count should describe the images rather than flatter them — but it is a deliberate reduction, not an accident, and you should see it coming rather than discover it in tomorrow's refresh.

## Testing

```
$ bats tests/bats/test_base_contract_asserts_the_system_bus.bats
1..5   (all ok)

# negative control — bus check deleted from the script:
4 of 5 fail

$ pytest tests/test_boot_gate_blocks_green.py -q
8 passed

$ shellcheck --severity=error --exclude=SC1091 build_scripts/checks/verify-base-contract.sh
(clean)
```

`test_build_scripts_remaining.bats` has two failures (#34 zypper DM, #74 sshd privsep). I checked them against a clean `main` with this diff stashed and they reproduce identically — pre-existing, not this change.

## Related issues

The underlying `dbus_contexts` denial on the rolling EL10 bases (Kitten 10, Stream 10) is a separate, unfixed defect — this PR makes it *visible* rather than fixing it. Full diagnosis in my next comment / issue.

---


hanthor:
Answering the docs nudge with a change rather than a claim — this one *is* contributor-visible, so it needed documenting.

`.github/green-criteria.yml` is where a contributor reads what a green `boots` cell actually asserts, and its `proves` line described the old contract:

> base to an operable multi-user bootc system (`TUNAOS_BASE_CONTRACT_OK`)

It now names the bus, and says why it had to be named:

> base to an operable multi-user bootc system **with a live system message bus** (`TUNAOS_BASE_CONTRACT_OK`). The bus is named because it was the gap: `skipjack:kde` promoted an image whose `dbus-broker` never started, since its display manager did not hard-depend on the socket the way `gdm` does (run 34705564876).

The gap was never that nobody thought of D-Bus. It was that a dead bus is invisible on any image whose DM doesn't hard-depend on the socket — so one broken machine produced a red gnome cell and a green kde cell, and the green one shipped.

The new failure reason a contributor may hit is `TUNAOS_BASE_CONTRACT_FAIL reason=system-bus-inactive unit=… state=…`, documented in the script header alongside the existing `bootc-status-failed`.

Pushed as 82aa5a1. `pytest tests/test_green_criteria.py tests/test_ci_contract.py tests/test_boot_gate_blocks_green.py` — 52 passed; STE 3048, unchanged.

---

PR #2488, opened 2026-09-13
fix(grouper): a sourced post_install hook must return, not exit

hanthor:
## Summary

`install-desktop.sh` **sources** its `post_install` hooks, so `gnome-extensions.sh` ending its apt branch in `exit 0` did not end the hook — it ended the whole desktop install, with status 0.

`gnome.yaml` names three hooks. Only two ran ([run 34714039087](https://github.com/tuna-os/tunaOS/actions/runs/34714039087)):

```
Running post-install: tuna-flatpak-remote.sh
Running post-install: gnome-extensions.sh
                    ← flatpak-preinstall.sh never ran
```

so the image failed its own contract on:

```
missing required path: /usr/share/flatpak/preinstall.d/*.preinstall
```

### The dangerous part is that only one cell went red

**Every apt-based gnome image had the same truncated install.** It is invisible wherever the Containerfile stage also runs `configure-desktop-runtime.sh`, which lays the flatpak baseline down a second time — that script's own comment says it does so "inside the same block that runs the contract", precisely so a non-manifest stage cannot fail on it.

`grouper:gnome-zfs` is simply the one stage without that compensator. So it isn't a gnome-zfs bug; it's the one cell that reported a defect all of them had.

### The fix already existed twelve lines below

```bash
# Sourced by install-desktop.sh's post_install; return so we don't exit it.
return 0 2>/dev/null || exit 0
```

The non-dnf branch got this right. The apt branch never got the same treatment.

### The test generalises a guard the repo already had

`tests/bats/test_niri_dms_payload.bats` pins this exact property for *one* hook, after the same bug shipped a niri image with no shell (#1009, #637):

```bash
# Sourced, not executed: an `exit` here would abandon the rest of the
# desktop install (branding checks included) and still report success.
! _code "$hook" | grep -qE '^[[:space:]]*exit[[:space:]]'
```

But the rule isn't about that hook — it's about how hooks are *invoked* — so it belongs to every script a manifest names. The new test reads the hook list out of the manifests rather than hardcoding it, so a hook added tomorrow is covered without touching the test.

I swept all 8 hooks across every desktop manifest. `gnome-extensions.sh` was the only violator.

## Testing

```
$ pytest tests/test_a_sourced_hook_never_exits_the_desktop_install.py -q
9 passed

# negative control — bare exit restored:
FAILED …::test_a_post_install_hook_does_not_exit[gnome-extensions.sh]

$ pytest tests/ -k "desktop or manifest or hook or install"
214 passed, 16 skipped

$ bats tests/bats/test_niri_dms_payload.bats      # 9 ok — the older guard still holds
$ shellcheck --severity=error --exclude=SC1091 build_scripts/desktop/gnome-extensions.sh
(clean)
```

## Related issues

Same failure mode as #1009 / #637 (niri image shipped with no shell), which is why the guard for it already existed on one hook.

Expected effect: `grouper:gnome-zfs` should pass its Desktop contract, taking grouper 6/7 → 7/7. Its `Gate` also failed in that run and I have not diagnosed that separately — if it turns out independent of the missing flatpak baseline, it needs its own fix.

---


hanthor:
Answering both nudges, and the first one turned up something bigger than this PR.

**Desktop contract** — already describes it, and is the thing that caught this: `require_glob '/usr/share/flatpak/preinstall.d/*.preinstall'` plus the per-app grep for `io.github.kolunmi.Bazaar` and `org.mozilla.firefox`. No change needed; the contract was right and the install was truncated.

**`docs/IMAGE-TAGS.md`** — no mention of flatpak or preinstall declarations. Nothing to update.

**Docs** — contributor-visible, documented in the diff: the `return 0 2>/dev/null || exit 0` line carries the explanation, and the new test's docstring records the measurement. What changes for a *user* is that apt-based gnome images now actually receive the curated preinstall list they were always meant to have, so this is a fix toward the documented behaviour rather than a change away from it.

---

**The README variant table is stale in both directions, on most rows.** Comparing its flavor column against `build-config.yml` (`build_image: true`), ignoring `base` and the `-hwe`/`-nvidia`/`-cachyos` hardware suffixes since that column lists desktops:

| variant | README says | actually built | wrong how |
|---|---|---|---|
| guppy | GNOME, KDE | kde, xfce | advertises **GNOME**, removed in #2451; omits XFCE |
| grouper | GNOME, KDE, Niri, XFCE | gnome, gnome-zfs, kde, cosmic, xfce | advertises **Niri**, which it does not build; omits COSMIC |
| flounder | KDE, COSMIC, Niri, XFCE | kde, xfce | advertises **COSMIC** and **Niri**, neither built |
| flounder-sid | GNOME, KDE, COSMIC, Niri, XFCE | gnome, kde, xfce | advertises **COSMIC** and **Niri**, neither built |
| wahoo | Base, GNOME | base, gnome, cosmic, kde | omits COSMIC and KDE |
| sailfin | GNOME, KDE, Niri, XFCE | gnome, kde, niri, xfce, cosmic | omits COSMIC |
| bonito | GNOME, KDE, COSMIC, Niri | …, xfce | omits XFCE |
| skipjack | GNOME, KDE, COSMIC, Niri | …, xfce | omits XFCE |

The guppy row is partly mine: #2451 dropped `guppy:gnome` for the GNOME 50 floor and I updated the generated counts without touching this hand-written row.

None of it belongs in this PR — it is unrelated to a sourced hook, and mixing a docs sweep into a one-line behaviour fix makes both harder to review and to revert. It is worth its own change, with a test pinning that column to `build-config.yml`, since nothing checks it today and it has now drifted on eight rows.

---

PR #2490, opened 2026-09-13
fix(ci): stop building the flavors build-config turns off

hanthor:
## Summary

`build_image: false` is how build-config says **"do not build this"**, and `build-variant.yml`'s matrix generator never read it.

Every other consumer does — `luks-e2e`, `bootc-lifecycle`, `catalog-facts`, `daily-verify`, `live-overlay`, `live-initramfs` and `desktop-contract-sweep` all `select(.build_image == true)`. So the disabled flavors were still built, and they fail, because they're disabled for reasons that make them unbuildable.

Measured across the newest run of each variant:

| variant | flavor | jobs | result |
|---|---|---|---|
| albacore | gnome-asahi | 2 | all failure ([34726078460](https://github.com/tuna-os/tunaOS/actions/runs/34726078460)) |
| yellowfin | gnome-asahi | 2 | all failure ([34705563513](https://github.com/tuna-os/tunaOS/actions/runs/34705563513)) |
| skipjack | gnome-asahi | 2 | all failure ([34705564876](https://github.com/tuna-os/tunaOS/actions/runs/34705564876)) |
| flounder | gnome | 3 | all failure ([34657636354](https://github.com/tuna-os/tunaOS/actions/runs/34657636354)) |
| flounder | gnome-nvidia | 2 | all failure (same run) |

albacore's names the reason outright:

```
no image found in image index for architecture "arm64"
```

`gnome-asahi` is the arm64 leg of a `gnome` that build-config pins to amd64 — stated in a comment sitting **directly above** the `build_image: false` that was being ignored.

### Why it matters beyond the wasted runners

Eleven guaranteed-red jobs a night, and they set the run conclusion. **flounder reported `failure` in that run with all five of its published flavors promoted.** Anyone checking whether flounder is healthy saw a red run caused entirely by two images nobody ships.

### No cell count moves

`gen-matrix-status.py` already scores `build_image` flavors only, so these were never counted. 144 config entries filter to **exactly the 139** the docs have always claimed — which the new test pins, so the matrix and the published count can't drift apart silently.

### The parent test is what makes the filter safe

A stage-3 flavor layers on its prefix (`gnome-nvidia` on `gnome`), so filtering could orphan a child whose parent is off. No flavor is in that state today — flounder turns `gnome` and `gnome-nvidia` off together — and the test now fails the moment a config change creates one, which is the point to revisit the filter rather than discover it in a nightly.

## Testing

```
$ pytest tests/test_the_build_matrix_honours_build_image.py -q
4 passed

# negative controls:
filter removed     → FAILED …::test_the_matrix_filters_on_build_image
parent turned off  → FAILED …::test_no_disabled_flavor_is_the_parent_of_an_enabled_one
                     FAILED …::test_the_filtered_matrix_is_the_cell_count_we_publish

$ pytest tests/ -k "ci_contract or build_variant or matrix or job_budget or nightly"
175 passed, 2 skipped
```

## Related issues

Same family as #2451 (guppy:gnome burning 1–4h of CI a day on a guaranteed-red cell) — except here the flavors were already turned off and the matrix was ignoring the switch.

---


hanthor:
Answering the workflow nudge: no green gate moved, and the pipeline docs did need a change.

**No `gates` block changes.** This does not move, rename or re-scope any gate — it stops scheduling build jobs for flavors `build-config.yml` already marks `build_image: false`. Those flavors were never counted as cells (`gen-matrix-status.py` has always scored `build_image` only), so no criterion's scope changes. `tests/test_ci_contract.py` passes.

**`docs/build-pipeline.md` did describe this, incompletely.** It documents `generate_matrix` twice — "reads `build-config.yml` via `yq` + `jq`, emits one matrix per stage" — and neither mention said which flavors it keeps, which is precisely the behaviour that was missing from the code. Both now name the filter, and the first records what it is worth:

> Only flavors with `build_image: true` enter the matrix. This filter turns the 144 declared flavors into the 139 published cells.

`docs/PIPELINE.md` describes the stages rather than the selection, so it stays as is.

One note on that wording, since it cost a round: my first draft was a single 28-word sentence and tripped `STE 3.1` (limit 20 for instructions), taking the ratchet 3048 → 3049. Split into two sentences, back to 3048. The ratchet only goes down, so I fixed the prose rather than the number.

Pushed as 6c43ece.

---

PR #2494, opened 2026-09-13
fix(bonito): the t2 kernel swap constrained the wrong half

hanthor:
## Summary

`bonito:gnome-t2` is bonito's only unpromoted cell, and it fails identically every run — three attempts, [run 34613732493](https://github.com/tuna-os/tunaOS/actions/runs/34613732493) and again in [34750383558](https://github.com/tuna-os/tunaOS/actions/runs/34750383558):

```
Failed to resolve the transaction:
No match for argument 'kernel' in repositories 'copr:…:sharpenedblade:t2linux'
```

`dnf swap --from-repo=X A B` constrains **both** specs to `X`. The remove spec is the **installed** kernel, which does not come from the COPR — so it matched nothing and the transaction died before the install side was ever considered.

### The install side was never the problem, and that's checkable

The build log resolves `VERSION_ID=44`, so the enabled chroot is `fedora-44`. Querying that chroot directly rather than assuming:

```
GET …/sharpenedblade/t2linux/fedora-44-x86_64/repodata/…
kernel 7.1.9-200.t2.fc44      (kernel-core present too, 54 packages)
```

The package is right there. Since a repo-constrained *install* of `kernel` would have found it, the argument that failed to match has to be the remove.

### The fix

Remove unconstrained, install constrained, **assert the result**. The shape is `asahi.sh`'s, which swaps a kernel for the same reason and already pairs `remove --noautoremove … || true` with `install --allowerasing` plus a hard verification.

The assert is the point. `--from-repo` was there to stop a Fedora kernel silently winning — but a flag whose scope surprised us isn't that guarantee. `rpm -q kernel` matching `.t2.` is.

`rpm -qa 'kernel*'` runs first, so if an assumption here is wrong the log names the packages actually installed rather than repeating "No match for argument".

## ⚠️ Not yet verified against a real build

The diagnosis is solid — the failing argument cannot be the install side, because that package is demonstrably in that chroot — but **only a bonito build proves the replacement resolves**. I'll dispatch one when this merges. If it fails, the `rpm -qa` line makes the next iteration certain rather than another guess.

I'm flagging that rather than burying it: this is the one change in this session's run that ships on deduction instead of on a green build.

## Testing

```
$ bats tests/bats/test_t2_kernel_swap.bats
1..6   (all ok)

# negative control — old one-liner restored: 5 of 6 fail

$ pytest tests/ -k "t2 or overlay or kernel"     15 passed, 2 skipped
$ bats tests/bats/test_nvidia_overlay.bats       43 ok
$ shellcheck --severity=error --exclude=SC1091 build_scripts/overlay/t2.sh
(clean)
```

The tests pin the property, not the spelling: the remove must not be repo-constrained, the install must be, and the result must be checked.

## Related issues

Diagnosed early in this session and deferred as out of scope at the time; bonito is now 13/14 with this as the only cell left.

---


hanthor:
Answering the docs nudge: user-visible in principle, but there is currently nothing to document *against*, and no doc claims otherwise.

`bonito:gnome-t2` has never promoted — the swap fails all three attempts, every run — so no T2 image exists for a user to have expectations about. This change doesn't alter documented behaviour; it makes the flavor build for the first time.

I checked what the repo says about T2 rather than assuming:

```
$ grep -rln -i 't2linux|gnome-t2' docs/
docs/matrix-provenance.json      ← generated
```

Nothing in `docs/HARDWARE.md`, `docs/IMAGE-TAGS.md` or the README variant table mentions T2 support at all. The only human-readable mention anywhere is the auto-generated build-status row listing `gnome-t2` among bonito's unreached cells. So there is no prose to bring in step, and writing some now would be documenting a flavor that has not yet booted anywhere.

The behaviour that *is* worth recording lives in the diff: `t2.sh`'s header now explains why the swap is split (`--from-repo` constrains both specs of a `dnf swap`, and the remove spec is the installed kernel), with the run IDs and the live-repo query that establish it.

**If the dispatched bonito build promotes the cell, a `docs/HARDWARE.md` entry is the right follow-up** — at that point there is a real, booting T2 image with a known kernel to describe. Documenting it before then would be a claim ahead of the evidence, which is the failure mode the README variant table just got fixed for in #2489.

---

Issue #2518, opened 2026-09-14
*-nvidia-hwe cannot mount /sysroot on any EL10 variant: Fedora 43 kernel under an EL10 dracut


hanthor:
`albacore:gnome-nvidia-hwe` is the one cell keeping albacore off 18/18. `yellowfin` and `skipjack` declare the same flavor and fail identically. All three build, sign and pass their desktop contract, then boot into an emergency shell:

```
mount[597]: mount: /sysroot: unknown filesystem type 'xfs'.
Failed to mount sysroot.mount - /sysroot.
Dependency failed for ostree-prepare-root.service - OSTree Prepare OS/.
```

albacore [run 34745445506](https://github.com/tuna-os/tunaOS/actions/runs/34745445506) at 10.6s, yellowfin [run 34738897386](https://github.com/tuna-os/tunaOS/actions/runs/34738897386) at 10.3s. The root really is XFS.

## Cause

The fc43-kernel-on-EL10 split that #1561 records in `20-nvidia.sh`, one module further on.

| cell | akmods bundle | kernel installed | dracut | boots |
|---|---|---|---|---|
| `gnome-nvidia` | `centos-10` | `6.12.0-266.el10` | `dracut-107-8.el10_2` | ✅ |
| `gnome-nvidia-hwe` | `coreos-stable-43` | `7.1.8-100.fc43` | `dracut-107-8.el10_2` | ❌ |

`scripts/build-image-inner.sh:102` selects the coreos-stable akmods bundle whenever `ENABLE_HWE=1`, so `10-kernel-swap.sh` installs *that bundle's* kernel. On an EL10 variant that is Fedora 43's kernel under EL10's dracut. The non-HWE nvidia cells take the EL10 kernel from the `centos-10` bundle and boot fine, which is why only the `-hwe` combination is red.

Under `--no-hostonly` dracut has no fstab to read, so it infers the filesystem set. It infers correctly for the kernel it shipped beside and not for the Fedora one, **and it says nothing**: the dracut run in the build log reports no error and never mentions xfs at all.

**There is no bundle to switch to.** `ghcr.io/ublue-os/akmods-nvidia-open` publishes `centos-10`, `centos-10-aarch64`, `centos-10-x86_64` and `centos-10-6.12.0-233.el10` for that family. Nothing HWE. I read the tag list rather than assuming.

## What I tried, and why it is reverted

[...]

Issue #2514, opened 2026-09-14
The base boot contract waits on a settle its own job blocks — the bound can never be right, and it asserts nothing on any cell today


hanthor:
`tunaos-base-contract.service` waits for a settle that its own job is blocking. The wait can never succeed, at any bound. I tuned that bound twice today (#2501, then #2506) and both times the machine settled a few milliseconds *after* the bound expired — because the bound expiring is what let it settle.

## The signature

Same cell, `hummingbird:gnome`, two different bounds:

| bound | wait abandoned | `Startup finished` | gap |
|---|---|---|---|
| 120s (#2501) | `133.320660` | `133.334844` — userspace **2min 7.074s** | **14 ms** |
| 240s (#2506) | `252.338862` | `252.351227` — userspace **4min 7.191s** | **12 ms** |

Userspace settle time tracked the bound. It was never a property of the image. I read `2min 7s` off the first run and raised the bound to 240s to clear it; the second run then reported `4min 7s`.

## Why

[Run 34791370259](https://github.com/tuna-os/tunaOS/actions/runs/34791370259), in order, to microsecond resolution:

```
[ 252.338862] verify-base-contract[868]: TUNAOS_BASE_CONTRACT_NOTE settle-wait-timed-out state=starting
[ 252.339336] verify-base-contract[868]: TUNAOS_BASE_CONTRACT_FAIL reason=system-state=starting
[ 252.340347] systemd[1]: tunaos-base-contract.service: Main process exited, code=exited, status=1/FAILURE
[ 252.341088] systemd[1]: Failed to start tunaos-base-contract.service
[ 252.351227] systemd[1]: Startup finished in 1.227s (kernel) + 3.932s (initrd) + 4min 7.191s (userspace)
```

Startup finished **10 ms after the unit exited**, not after anything else happened. The unit is the last job in the initial transaction:

```
[Unit]
After=multi-user.target
...
[Install]
WantedBy=multi-user.target
```

`systemctl is-system-running --wait` returns when the manager leaves `starting`, and the manager leaves `starting` when the initial transaction completes. This unit's job is in that transaction. So the check waits for a state that cannot be reached until the check returns. `state=starting` is not a slow boot. It is the contract reporting its own job.

## The other half: it asserts nothing on desktop cells either

`bonito:gnome-t2` is green. Its boot gate log ([run 34768771243](https://github.com/tuna-os/tunaOS/actions/runs/34768771243)):

```
[...]

Issue #2492, opened 2026-09-13
sailfin:gnome-asahi builds and signs, then fails the Asahi manifest gate on eight missing pieces


hanthor:
## What happens

`sailfin:gnome-asahi` is the last red cell on sailfin after #2481 brought the variant back from 0/7 to 6/7. It is **not** a build failure — the image builds, the manifest assembles and it signs:

```
🦈-sailfin / gnome-asahi / linux-arm64                    success
🦈-sailfin / gnome-asahi / Manifest                       success
🦈-sailfin / gnome-asahi / Sign                           success
🦈-sailfin / gnome-asahi / Asahi manifest gate / verify   failure   ← here
🦈-sailfin / gnome-asahi / Promote                        skipped
```

The gate's findings, from [run 34750381815](https://github.com/tuna-os/tunaOS/actions/runs/34750381815):

```
FAIL macsmc.ko not in modules.dep
FAIL apple DTB dir missing/sparse (0)
FAIL t8103-j313.dtb missing
FAIL update-m1n1 missing (tried: /usr/bin/update-m1n1 /usr/sbin/update-m1n1)
FAIL update-m1n1 kernel hook missing (tried: /usr/lib/kernel/install.d/15-update-m1n1.install
                                             /etc/kernel/postinst.d/update-m1n1
                                             /etc/kernel/postinst.d/zz-update-m1n1)
FAIL asahi-fwupdate missing (tried: /usr/bin/asahi-fwupdate /usr/sbin/asahi-fwupdate)
FAIL /usr/bin/tiny-dfr missing
FAIL /usr/lib/systemd/system/tiny-dfr.service missing
```

## Why this is not simply "the overlay didn't run"

`build_scripts/overlay/asahi.sh` has a real openSUSE branch (OBS `home:mrkcee`), and its **required** transaction succeeded — the script exits 1 after three attempts otherwise, and it didn't:

```bash
zypper … install kernel-asahi m1n1 u-boot-asahi asahi-scripts && break
```

So `asahi-scripts` installed, yet `update-m1n1` and `asahi-fwupdate` are not at any path the gate looks for. Two readings, and I cannot separate them from the logs:

1. **openSUSE lays these out differently.** Every path the gate probes is the Fedora/EL layout (`/usr/lib/kernel/install.d/15-update-m1n1.install`, `/etc/kernel/postinst.d/…`). If OBS ships them elsewhere, the gate is wrong for this family, not the image.
[...]

Issue #2485, opened 2026-09-13
dbus-broker cannot read dbus_contexts on the rolling EL10 bases — 8 gnome cells red, and kde promotes without a system bus


hanthor:
## What happens

On **yellowfin** (AlmaLinux Kitten 10) and **skipjack** (CentOS Stream 10), `dbus-broker` fails to start at boot. The serial log names the cause directly, which it could not do before #2472 forwarded the journal to the console:

```
[ 8.638] dbus-broker-launch[948]: Looking up NSS user entry for 'dbus'... (ok)
[ 8.662] dbus-broker-launch[948]: Access denied in /etc/selinux/targeted/contexts/dbus_contexts +1:
                                  /etc/selinux/targeted/contexts/dbus_contexts
[ 8.664] dbus-broker-launch[948]: ERROR run @ ../src/launch/main.c +152: Return code 1
[ 8.664] dbus-broker-launch[948]: Exiting due to fatal error: -131
[ 8.672] systemd[1]: dbus-broker.service: Main process exited, code=exited, status=1/FAILURE
```

Five restarts, then:

```
[ 8.934] systemd[1]: dbus.socket: Failed with result 'service-start-limit-hit'
[10.50] systemd[1]: Dependency failed for gdm.service - GNOME Display Manager
```

and the cascade: `systemd-logind`, `polkit`, `upower`, `NetworkManager`, `tuned`.

Two other userspace SELinux consumers fail the same way, which is what identifies this as a `/etc/selinux/targeted/contexts/` access problem rather than something specific to D-Bus:

```
systemd-logind[1024]: Failed to initialize SELinux labeling handle: Permission denied
sshd-session[1512]: pam_selinux(sshd:session): Unable to get valid context for root
sshd-session[1512]: pam_systemd(sshd:session): Failed to connect to system bus: Connection refused
```

That last pair is why the boot gate could never diagnose itself: its diagnostics go over SSH, and `sshd` cannot open a PAM session without the bus.

The kernel loads policy fine — `systemd[1]: Successfully loaded SELinux policy in 173.514ms` — so this is userspace being denied the context files, not a missing or unloaded policy.

## Scope

| variant | base | `dbus_contexts` denied | outcome |
|---|---|---|---|
| albacore | AlmaLinux 10 | **0** | gnome + kde pass honestly |
| yellowfin | AlmaLinux **Kitten** 10 | yes | 4 gnome cells red |
[...]

PR #2548, opened 2026-09-17
fix(ci): unbreak the STE ratchet, withdraw *-nvidia-hwe on EL10, make #2485 measurable, and brand GNOME

hanthor:
## Summary

Four things, found by catching up on a matrix that had run unattended for three days.

### 1. The STE ratchet has been red on `main` since 2026-09-14, and it is mine

It fails on `main` and on **every** pull request that touches a `.md` file — most recently an unrelated one at 01:58 today.

```
2026-09-14T18:54 push         main                        failure
2026-09-15T11:08 pull_request automation/matrix-status    failure
2026-09-16T10:57 pull_request automation/matrix-status    failure
2026-09-17T01:58 pull_request strategy/q3-close-q4-...    failure
```

STE lints `README.md`. Nobody writes it by hand. #2512 changed the paragraph inside `update-build-status.sh`, and the STE run on that PR read **3048** — correctly, because the committed README still held the old sentence. Automation regenerated it two days later, the new wording carried a filler word and a passive clause, and the total reached **3052** against a budget of **3050**.

So a pull request moved linted prose and its own diff showed none of it. The gate had nothing to catch and I had nothing to see.

The second source needs no author at all: `gen-matrix-status.py` prints a caveat paragraph only while the newest sweep holds a cell with no clean verdict, so **a sweep result alone moves the repo total**. That is the same paragraph that broke the drift gate in failure-catalogue §22, from the other side.

Fixed at both ends — the generator *and* the committed file in one commit, which is the whole point — and the caveat rewritten so no future sweep can move the number. Back to **3048**, two under budget. **The budget is untouched:** the prose was wrong, the number was right.

`tests/test_generated_prose_is_visible_in_the_diff.py` holds the generator and the committed README in step, and fails against main's current pair.

### 2. `*-nvidia-hwe` withdrawn on the EL10 family — **138 → 135**, albacore reaches 18/18

All three EL10 variants declare it, all three build it, sign it, pass its desktop contract, then boot to an emergency shell:

```
mount: /sysroot: unknown filesystem type 'xfs'
Dependency failed for ostree-prepare-root.service - OSTree Prepare OS/.
```

`-hwe` selects the coreos-stable akmods bundle, so the kernel swap puts **Fedora 43's kernel under an EL10 dracut**, and that dracut leaves xfs out of the initramfs silently. The non-HWE nvidia cells take the EL10 kernel and boot, which is why only this combination is red. There is no EL10 HWE bundle to switch to — `ghcr.io/ublue-os/akmods-nvidia-open` stops at `centos-10`. Naming the filesystems was tried and reverted in #2517: `--filesystems` is strict, EL10 ships no btrfs, and it broke every EL10 `*-nvidia` cell to repair one.

Declared rather than deleted, on the #2492 precedent: the block keeps the evidence and the route back. #2518 tracks getting it back.

### 3. #2485 can finally be measured

Open since 2026-09-13 on a question nobody could answer: the **runtime** label of `/etc/selinux/targeted/contexts/dbus_contexts`.

The image cannot explain it — read straight out of GHCR, that file is byte-identical to albacore's, `0644 root:root`, and no image in the family carries a `security.selinux` xattr at all. And the machine cannot be asked: the gate collects diagnostics over SSH, and sshd cannot open a PAM session without the bus, so every broken boot's artifact says `guest SSH unavailable`. **The one image that could answer is the one that cannot be asked.**

Serial is immune. The probe prints enforcing mode, the runtime context of that file and its directory, what `matchpathcon` expects, whether root can read it, and the `libselinux` version — which is the hypothesis under test, since libselinux/libsepol/libsemanage/policycoreutils are **3.10 on albacore and 3.11 on both failing variants** while `selinux-policy` is pinned identical on all three.

It sits **above** the settle wait deliberately: everything below it is unreachable on exactly these images (#2514), and evidence that only prints on healthy machines is not evidence.

**I did not pin `libselinux-3.10`, though it was approved.** Doing so would be a second guess stacked on an unmeasured mechanism — the shape that cost a red build in #2517. The probe makes the next skipjack build decide it.

### 4. GNOME had no branding on any variant, and every check passed

`marlin:gnome` shows Arch's stock wallpaper. Read out of the published image, all three assets are present:

```
/usr/share/pixmaps/tunaos.svg                        4742 bytes
/usr/share/backgrounds/tunaos/tunaos-default.png   309864 bytes
/usr/share/plymouth/themes/default.plymouth        symlink to shark
```

and nothing tells GNOME to use any of them. A repo-wide grep for `picture-uri` across `system_files/`, `build_scripts/` and `experiences/` returns **nothing**. `verify-branding.sh` asks whether the wallpaper *file* exists, it does, so the cell was green the whole time. This is not a marlin bug — no variant applies GNOME branding.

`docs/BRANDING.md` is the plan: three layers, and which desktop owns which.

| layer | where | state |
|---|---|---|
| 1. Identity — os-release, image-info.json | `90-image-info.sh` | every variant |
| 2. Assets — logo, wallpaper, plymouth | `system_files/` | every variant |
| 3. **Application** — what the session reads | per desktop | GNOME was missing |

**The rule: assert the applied state, not the file.** A check that asks whether the wallpaper exists passes on an image where nobody set it.

- `gnome-set-branding.sh` writes the dconf keyfiles the session and the gdm greeter read, compiles them, and **asserts the compile** — `install-desktop.sh` runs `dconf update` above the branding block, so a keyfile written there would sit uncompiled until first boot, which is a state `verify-branding.sh` already fails.
- `verify-branding-gnome.sh` asserts that a keyfile sets `picture-uri`, that the URI resolves to a file in the image, that `picture-uri-dark` is set too, that the greeter has its own logo in `gdm.d`, and that both databases compiled. It deliberately does **not** look for our keyfile by name: any mechanism that leaves the session branded passes.
- Both are wired into **both** install paths. `kde-set-look-and-feel.sh` lived in one of them for a while and half the matrix shipped unbranded; a test now asserts both call sites by name.

Measuring layer 1 turned up a second one: **`HOME_URL` and `DOCUMENTATION_URL` pointed at `projectbluefin.io` on every single variant**, inherited verbatim from the upstream script. The URL loop in `verify-branding.sh` read only `SUPPORT_URL` and `BUG_REPORT_URL` — the two that were already ours — so nothing noticed. Both URLs now name this repo and the contract checks all four.

COSMIC and XFCE still apply nothing; `docs/BRANDING.md` says so and gives the four-step recipe for adding a desktop.

## Testing

- 173 contract/selinux/base tests pass; all 9 `test_base_contract_asserts_the_system_bus.bats` cases still pass, including those that execute the script.
- New tests: 8 for the probe, 5 for the withdrawal, 4 for the generator coupling, **17 for GNOME branding** — each with a verified negative control.
- The probe is tested by **running** it, with its tools present and absent. That caught a real bug no reading would: under `pipefail`, `2>&1 | head -1 || echo` put a shell error *inside* a marker line and its fallback on a second line, so one field became two lines and neither was a measurement. Now five single-line fields, exit 0, either way.
- The branding tests are likewise run, not read: `gnome-set-branding.sh` writes into a temp tree and `verify-branding-gnome.sh` judges it. The headline negative control reproduces the original bug verbatim — assets present, no keyfile, contract red. Six mutations were applied to the sources (drop either call site, revert `HOME_URL`, narrow the URL loop, drop `picture-uri-dark`, write the greeter logo to the wrong database) and **all six went red**.
- `shellcheck` clean, `yamllint` clean, STE **3048**.
- Full local suite running; CI is authoritative.

## Related issues

Refs #2485, #2514, #2518, #2492, #2512.


hanthor:
### On the nudge: `docs/MATRIX-STATUS.md` **was** hand-edited, on purpose

The reminder is right that the file is generated, so here is why this commit edits it anyway — and the check that confirms it was correct rather than convenient.

`--check-structure` drops every line matching `VOLATILE_LINE` from both sides before comparing. This PR changes the wording of the sweep caveat in `gen-matrix-status.py` and moves the mask to match the new opening. That leaves exactly two consistent states:

| committed file | generated side | masked? | drift gate |
|---|---|---|---|
| left stale (old sentence) | new sentence | committed **no**, generated **yes** | one side has a line the other lacks → **fails** |
| updated (this PR) | new sentence | both **yes** | both dropped → **passes**, including when the sweep is clean and the paragraph is absent |

So the edit is what keeps the generator and the committed copy consistent, not a shortcut around regeneration. **`Matrix Status` passed on `119e4f2`**, which is the check that actually exercises this.

It also carries the STE headroom that is the point of the first commit. `README.md` alone brings the total to 3050 — exactly the budget, zero margin, one generated-prose wobble from red again. With the caveat rewritten too it is **3048**.

The next scheduled refresh overwrites my edit with byte-identical content, because the generator now emits precisely that sentence. `tests/test_generated_prose_is_visible_in_the_diff.py` fails if the two ever drift apart again.

The alternative — mask both the old and new wording and touch nothing — would also have worked, at the cost of dead alternation in the regex and no STE margin. I took the margin.


---

PR #2556, opened 2026-09-17
fix(bootc): stop 99-cleanup leaving /home a dangling symlink — the flounder ISO failure

hanthor:
## Summary

`flounder:kde` and `flounder:kde-nvidia` have failed to produce an ISO on **every run since at least 2026-09-05**, in tacklebox's embedded live baseline:

```
>>> [customize] (1/2) baseline.sh
useradd: cannot create directory /home
Error: live customize for flounder-kde: ... exit status 12
##[error]No files were found with the provided path: *.iso
```

The cause is **two scripts in this repo disagreeing**, and neither looks wrong read on its own.

`ostree-layout.sh` makes `/home` a symlink to `var/home`, then creates `/var/home`, and states the reason directly above the `mkdir`:

> The /var targets must EXIST at image-build time, not only in tmpfiles.d. Package maintainer scripts in later layers probe the aliases …

`99-cleanup.sh` then runs

```sh
find /var -mindepth 1 -maxdepth 1 ! -path '/var/cache' -delete
```

and takes `/var/home` with it, restoring only the dpkg and apt paths below it.

## The mechanism predicts exactly who breaks

On every base where `99-cleanup` runs **after** `ostree-layout`, the published image ships `/home` pointing at nothing. That is precisely three Containerfiles:

| Containerfile | layout | cleanup | |
|---|---:|---:|---|
| `Containerfile.arch` | 156 | 245 | marlin |
| `Containerfile.debian` | 247 | 290 | **flounder** |
| `Containerfile.gentoo` | 539 | 659 | guppy (builds no ISOs) |

The bases with no `ostree-layout` step — `el10`, `opensuse`, `ubuntu` — take their layout from the parent image and do not show it. flounder is Debian.

## Why nothing already caught it

tacklebox is the consumer that finds it, because it runs `useradd --create-home` against the **published** image while building the ISO.

tunaOS' own `customize-live.sh` **already carries the fix** — `mkdir -p "$(readlink -f /home)"`, under a comment describing this precise failure — but it runs as script **2/2**, after the baseline that needs it. There is no earlier hook: tacklebox documents `--script` as running after the embedded baseline (`cmd/tacklebox/customize_live.go:53`), and its `baseline.sh` on current `main` still has the unguarded `useradd`. So a pin bump does not help either.

## The change

Restore the `/var` alias targets in `99-cleanup.sh`, in the same place and for the same stated reason as the dpkg paths immediately below. **The wipe itself is deliberate and stays**, and the `tmpfiles.d` entries `ostree-layout.sh` writes still own the runtime side after a factory reset. This owns build time.

## Testing

`tests/bats/test_var_alias_targets_survive_cleanup.bats`, 4 cases. Three are shape assertions across **both** scripts, because the bug is that they disagree. The fourth **executes** the wipe against a temp tree, asserts `/home` dangles (the bug, reproduced), applies the restore, and asserts a `useradd`-shaped `mkdir` through the symlink then succeeds — a grep cannot tell a resolving symlink from a broken one.

Negative controls run: deleting the restore is caught, **and so is moving it above the wipe**, which leaves both greppable lines present and the bug intact.

## What this does NOT claim

`marlin:gnome` arm64 fails the same baseline with a **different** error — `useradd: UID 1000 is not unique`, exit 4. marlin is `Containerfile.arch`, so it shares this defect, but that message is not this one and I have not shown the two are the same bug. It needs its own look.

---


hanthor:
> **Docs.** This PR changes build or pipeline code and no documentation.

Fair, and this one earns an entry rather than a "not user-visible" — the symptom pointed at another repository for two weeks while the fix sat in this one.

Added as **failure-catalogue §24** in `docs/ci-troubleshooting.md` (`1daf0ca`): the mechanism, who it hits (only the bases where `99-cleanup` runs after `ostree-layout` — arch, debian, gentoo, which names the failing cells exactly), and the part worth remembering — `customize-live.sh` already had this fix, under a comment describing the precise failure, as script 2/2 behind the script 1/2 that needed it.

Not user-visible in the installed system: `/var/home` is a directory `systemd-tmpfiles` already recreates at boot. What changes is the **published image**, and the audience for that is contributors, which is what the catalogue is for.

The first draft cost 6 STE findings against 14 of headroom. Rewriting three passive sentences brought it to 1 — **1636 → 1637**, budget untouched at 1650.

---

PR #2557, opened 2026-09-17
fix(arch): free UID 1000 on the Arch Linux ARM base — marlin's arm64 ISO

hanthor:
## Summary

`marlin:gnome` builds an ISO on amd64 and dies on arm64, in tacklebox's embedded live baseline:

```
>>> [customize] (1/2) baseline.sh
useradd: UID 1000 is not unique
Error: live customize for marlin-gnome: ... exit status 4
```

**This repo already fixed exactly this failure once**, for Ubuntu, in the block directly above the change: `docker.io/library/ubuntu` ships a cloud account at UID 1000, tacklebox's baseline asks for 1000 unconditionally, and the ISO build dies. The fix removed the account. The guard was `IS_UBUNTU`.

## The same account, a different base

marlin's arm64 legs build **FROM `ghcr.io/tuna-os/archlinuxarm`**, because `docker.io/archlinux` is x86_64-only. Read out of that published image's layer rather than assumed:

```
alarm:x:1000:1000::/home/alarm:/bin/bash
```

`build-archlinuxarm-base.yml` keeps it deliberately — *"alarm/root users stay (standard ALARM accounts)"* — which is correct for a base image and wrong for an ISO built downstream of it.

The x86_64 Arch base has no such account, and **that asymmetry is the whole reason marlin fails on one architecture only**. It is the piece #2556 explicitly did not claim to explain.

The removal is as narrow as the Ubuntu one: an account still named `alarm` **and** still at exactly 1000. A base that renumbers it, drops it, or an operator who repurposed the name is left alone rather than silently altered.

## The diagnostic moves out of the Ubuntu branch

The *"UID 1000 is still taken"* warning was written as a diagnostic for precisely this class of bug — its own comment says the real message arrives *"an hour downstream, in another repo's code, with no mention of this image."*

And then it sat inside `if IS_UBUNTU`, so the identical failure on Arch ARM never printed it, and had to be traced back from an ISO job by hand. It now runs for **every** base, so the next one that ships an account at 1000 says so in the build that causes it.

## Testing

`tests/bats/test_uid_1000_is_free_for_the_live_user.bats`, 5 cases:

- both bases' removals are present
- each is guarded on the account being **exactly** 1000
- the warning is un-indented (an indented copy is trapped in a branch again)
- the fifth **executes** the guard against a stubbed `id`, at 1000 and at 1001 — a grep cannot tell a correct comparison from an inverted one

Negative controls run: disabling the `alarm` removal is caught, **and so is moving the warning back inside a per-base branch**. `test_build_scripts.bats` still passes (46 cases), which matters because it runs `shellcheck` with only `SC1091` excluded, so info-level findings fail the gate.

## Relationship to #2556

Independent, and they stack. #2556 fixes `useradd: cannot create directory /home` (flounder, Debian — a dangling `/home` symlink). This fixes `useradd: UID 1000 is not unique` (marlin, Arch ARM — an occupied UID). Same script in tacklebox, two different reasons it cannot run.

---


hanthor:
> **Docs.** This PR changes build or pipeline code and no documentation.

Added as **failure-catalogue §25** in `docs/ci-troubleshooting.md` (`f6bffd4`).

It records the mechanism (Arch Linux ARM ships `alarm` at UID 1000 — read out of the published layer, not assumed), why only one architecture failed (the x86_64 Arch base has no such account), and the part worth remembering: **this repo had already solved the same problem for Ubuntu**, and the guard `IS_UBUNTU` was a claim about who was affected that nobody revisited when a second base arrived. The diagnostic written for exactly this failure sat inside that same branch, so Arch ARM hit it in silence.

Pairs with §24 in #2556: same tacklebox script, two different reasons it cannot run.

**One thing for whoever merges second:** #2556 appends §24 to the end of this same file. Whichever lands first, the other wants its section renumbered rather than merged blindly — the conflict is textual, the content is independent. I'll handle it on whichever is second.

There *is* one user-visible edge worth stating plainly: on an arm64 marlin image the `alarm` account no longer exists. Nothing in tunaOS references it, and the first real user an installed system creates now lands at 1000 instead of 1001 — which is what every desktop session, flatpak permission and `$XDG_RUNTIME_DIR` path assumes. That is the same reasoning the Ubuntu block already gives for removing its account.

STE **1637 → 1640**, budget untouched at 1650.

---

hanthor:
#2556 merged first, so this is the "merges second" case I flagged above. Resolved in `ff69a20`.

The conflict was the textual one predicted — both PRs append a section to the end of `docs/ci-troubleshooting.md`. Both sections are kept in full, in order, and **no renumbering was needed**: §24 (from #2556) and §25 (this PR) were already numbered correctly relative to each other.

Re-verified on the merged tree rather than assumed:

- both code fixes present — `mkdir -p /var/home …` in `99-cleanup.sh` and the `alarm` guard in `01-workarounds.sh`
- both bats suites pass, 9 cases
- `shellcheck --severity=error` clean on both scripts
- zero conflict markers left
- STE **1641** against the budget of 1650

A merge commit, not a rebase, so anyone with this branch checked out keeps a valid checkout.

---

PR #2566, opened 2026-09-17
fix(bootc): rescue the kernel before /boot is wiped — marlin's arm64 ISO

hanthor:
## Summary

marlin's arm64 gnome ISO failed a **third** time on the same message after #2563 indexed the module tree:

```
no kernel found under /usr/lib/modules (looked for modules.dep):
7.2.6-1-aarch64-ARCH
```

The obvious read is that the depmod fix never reached the build. It did — the build log shows it at step 25/37, and the index is in the published image. tacklebox prints **one message for a loop with two requirements**:

```sh
for d in /usr/lib/modules/*/; do
  [ -f "$d/modules.dep" ] || continue
  if [ -f "$d/vmlinuz" ] && [ -f "$d/initramfs.img" ]; then
```

Past the first line, an absent `vmlinuz` ends at the same `exit 1` with the same text about `modules.dep`. The kernel binary was the missing file.

**Root cause.** Arch Linux ARM's `linux-aarch64` ships the kernel as `/boot/Image` alone, where Arch's x86_64 `linux` owns `/usr/lib/modules/<kver>/vmlinuz`. `build_scripts/bootc/ostree-layout.sh` runs `rm -rf /boot` before the dracut step, so on aarch64 it deleted the only copy. **The arm64 images have been shipping with no kernel at all.**

`Containerfile.debian` and `Containerfile.gentoo` (twice) already copy the kernel across by hand, one step before they call that script, and Gentoo's comment names the hazard outright: *"Both must happen BEFORE the ostree layout below, which deletes /boot"*. Arch was a fourth copy nobody wrote — so the rescue goes in the shared script, for the same reason the alias list lives there. It is a no-op wherever the kernel is already in place, which is every other variant.

`Containerfile.arch` now also refuses to build an initramfs for a kernel that is not there, so this class of fault fails at the build instead of shipping an image that cannot boot.

**Why it stayed hidden.** No Gate runs on arm64, so nothing booted the image and nothing reported that it could not. The ISO job was the first reader of that path, an hour downstream and in another repository.

## Testing

Local:

- **Full pytest: 1597 passed, 16 skipped, 70 subtests, 0 failed** — the same totals as the previous two PRs on this branch.
- **20/20** in `tests/bats/test_bootc_ostree_layout.bats`, including its existing shellcheck case. 6 new cases: rescue from `/boot/Image`; rescue from `/boot/vmlinuz-<ver>`; `Image` preferred over the non-EFI-bootable `Image.gz`; an existing `vmlinuz` left alone (with a `/boot` decoy present); no kernel at all still lays out; two kernel trees refused rather than guessed.
- **Mutation test** — dropping the call turns 4 of the 6 red; the other 2 correctly assert no-ops.
- **No regressions in bats.** Two comparisons against a clean `origin/main` checkout: on the four files that hold the local failures, an identical **18 on each side**; on the full suite, **every failure on this branch also fails on `main`** — the branch set is a strict subset, with **nothing branch-only**.
- **shellcheck** with the repo's `just lint` excludes: no new findings. The three that remain on this file (SC2115 ×2, SC2015) are byte-identical on `main` and sit on lines this PR does not touch.
- **STE:** 1641 findings, unchanged against the 1650 budget; `docs/ci-troubleshooting.md` back to its baseline of 8, and the `MIGRATION.md` section adds none.

On the bats failures, stated plainly: this sandbox has no container runtime, and its `yq` is the python jq-wrapper rather than the mikefarah build CI installs, so a set of tests is red here regardless of branch. The totals also drift run to run under parallel load, so those counts are not stable numbers — the load-bearing claim is the *branch-only* one, which is zero. CI is the arbiter.

Evidence for the diagnosis, since no container runtime is available here — the published image was read straight from the registry (manifest via a ghcr pull token, then each layer through `zstd -dc | tar -t`):

- `modules.dep` is present, in layer 55 — #2563's fix did land
- no `vmlinuz` or `Image` in **any** of the 65 layers

The control that settled it is `iso:gnome (linux-amd64)` in the *same run* (35236471968): built, booted, signed and published, from the same Containerfile and the same chunkah rechunk. The recipe is written up in `docs/ci-troubleshooting.md` §30, which also corrects §28 — that entry claimed the image had `vmlinuz` and booted, and it had neither.

Not verified locally: an actual arm64 image build, which needs the matrix.

## Related issues

Follows #2563 and #2558 on the same cell — each fix exposed the next fault behind it.


hanthor:
Acting on the `pr-nudges` reminder above — it is correct here, so `897362a8` adds a `MIGRATION.md` section rather than leaving it in the commit message.

Every marlin `linux-arm64` image published before this change holds no kernel, so a `bootc upgrade` onto one stages a deployment that cannot boot. That is a real behaviour change for an existing install, in both directions: the note covers which systems are affected (arm64 marlin only — x86_64 marlin and every other variant always carried the kernel), how to recover one that already stopped booting (previous boot entry, then `bootc rollback`), and how to check an image for `/usr/lib/modules/*/vmlinuz` before deploying it.

STE is 1641, unchanged, and the new section adds none.

---

PR #2573, opened 2026-09-18
ci: only publish ISOs to R2 on scheduled runs

hanthor:
## Summary

Every dispatch of a variant build published each gated ISO cell to R2. `upload-r2` defaults to `true` in `reusable-build-artifacts.yml`, and none of `build-variant.yml`'s three ISO call sites overrode it — so a test build, and every re-run chasing a red cell, paid to store artifacts nobody had asked for.

The cost is bigger than one copy per cell. A published cell writes the ISO **twice**: a dated object via `rclone copy`, then the `-latest` pointer via `rclone copyto`. At roughly 6 GB an ISO that is ~12 GB of writes per cell, across ~27 ISO cells, per dispatch. `prune-r2.yml` reclaims the dated copies a day later, but the writes are already paid for.

**None of that was ever needed to test an ISO.** The boot gate runs *before* the upload — this workflow's own header says a failed check publishes nothing — so a dispatch can still build the ISO, boot it in QEMU, and keep the workflow artifact and e2e evidence, while publishing nothing.

What changes:

- `build-variant.yml` gains a `publish-isos` input (boolean, default `false`) and passes `upload-r2: ${{ github.event_name == 'schedule' || inputs.publish-isos }}` at all three ISO call sites.
- Scheduled runs still publish, so the download URLs in `docs/TESTING.md` stay current and the NVIDIA download audit keeps reporting against real objects.
- Deliberate manual publishing already has its own workflow (`publish-isos.yml`, whose header says it is kept for exactly this), and `publish-isos` forces it here for a one-off.

No `build-<variant>.yml` needs editing: they all route through `build-variant.yml`, and the new input defaults to off.

## Testing

- **6 new cases** in `tests/bats/test_iso_upload_is_gated.bats`, covering: every call site decides `upload-r2` rather than inheriting; none publishes unconditionally; scheduled runs still publish; `publish-isos` is opt-in and off by default; the boot gate still precedes the upload (the invariant the whole change rests on); and the upload step is still guarded by the input.
- **2 mutations, 2 caught** — restore the unconditional `true`, and drop the override so it inherits the default again.
- **102 pytest cases** across the nine test files that read these two workflows (`test_no_workflow_hoards_multi_gigabyte_artifacts.py`, `test_ci_contract.py`, `test_generate_workflows.py`, `test_nightly_schedule_registry.py`, and others), all passing, plus the four existing bats files that read them.
- `yamllint -d relaxed` clean; the only warnings on this file are pre-existing line-length ones.
- STE 1641, unchanged against the 1650 budget.

One test in that set, `the repo actually has reusable-workflow call sites to check`, fails here — it fails identically on a clean `origin/main` checkout in this sandbox (one of the `yq`-related environment failures), so it is not a regression from this change.

## Related issues

Follow-up to the ISO work in #2566 / #2563 / #2558, which is what surfaced the cost: the four dispatches used to chase marlin's arm64 cell each published a full set of ISOs.

Worth a separate decision, not done here: each published cell still writes the ISO twice. Dropping the dated copy, or keeping only the `-latest` pointer plus a checksum, would roughly halve what the nightlies write — but that changes what is published, so it is the repo owner's call rather than a CI cost fix.

---


hanthor:
On the `pr-nudges` reminder above, checked each part:

- **`gates` block in `.github/green-criteria.yml`** — no change needed. No green gate moved: this PR changes only whether the *publish* step runs, and `green-criteria.yml` has no gate tied to the R2 upload. `tests/test_ci_contract.py` passes against the change (it is one of the nine workflow-reading test files run here, 102 cases, all green).
- **`docs/PIPELINE.md` / `docs/build-pipeline.md`** — neither mentions R2 or ISO publishing, so neither goes stale.
- **`docs/R2-COST-VISIBILITY.md`** — this one did need it, and `7d19fff5` updates it. It is the maintainer runbook that tracks who writes to each prefix, and its `live-isos/` row listed "the variant workflows" without qualification. That row now says *on their nightly schedule*, and a new subsection records what a variant build actually writes, including the part that makes the bill larger than it looks: each published cell writes its ISO twice, a dated object and the `-latest` pointer, both Class A operations.

STE is 1641, unchanged, and that file is back to its baseline of 18.

---

PR #2592, opened 2026-09-19
docs: record the Rekor outage decision

### LINKED ISSUE #1750: [decision] What to do when Rekor goes down again — signing-backend options, trigger thresholds, and the one that is not what it looks like
## Why this exists

Rekor 502s cost us two consecutive nightlies. On 08-14 it ran 45+ minutes and failed 25 `Sign + attest` jobs across 7 variants (#1560). On 08-15 ([run 31858324517](https://github.com/tuna-os/tunaOS/actions/runs/31858324517)) every albacore flavor built, pushed, produced an SBOM and passed its gates, and **all twelve Promote jobs were skipped** because `cosign attest` could not reach the log.

#1748 has since cut the exposure a lot. This issue is the thing to open **the next time it happens**: how to tell it is Rekor, what to do that day, and the threshold at which the answer stops being "retry harder" and becomes "change the trust model".

Written down now, while the evidence is fresh and nobody is under pressure.

---

## 1. Recognising it

The signature, from `cosign-retry.sh` when it gives up:

```
::error::SIGSTORE_OUTAGE: cosign still failing after 40m of Sigstore unavailability; giving up.
```

and underneath it:

```
Post "https://rekor.sigstore.dev/api/v1/log/entries": status 502: 502 Bad Gateway
```

Two checks before assuming:

```bash
curl -sS -o /dev/null -w '%{http_code}\n' https://rekor.sigstore.dev/api/v1/log
```

and https://status.sigstore.dev. A 5xx from `ghcr.io` is a different problem with a different fix — `cosign-retry.sh` deliberately fails fast on those rather than burning the budget, so if you see the budget being burned, it really is Sigstore.

## 2. What already absorbs it

Nothing needs doing for a short outage. As of #1748:

- **SBOM attestation is off the critical path.** `attest_sbom` is `continue-on-error` and absent from Promote's `needs`. A log outage delays attestations; it no longer blocks promotion.
- **Critical-path round trips per flavor dropped from `2 + 4N` to `N + 2`** (N = platforms). For albacore's three platforms: 14 → 5.
- **Two retry windows.** The sign job retries against a 40-minute wall-clock deadline; `rerun-infra-failures.yml` then re-runs failed jobs once, ~45 minutes after the run ends, giving a second 40-minute window.

Rough coverage: an outage up to ~40 minutes is absorbed on the first attempt, and one lasting up to ~1.5–2 hours is likely absorbed by the re-run. **The 08-15 outage (16m19s) would not have blocked anything. The 08-14 one (45+ min) is marginal** — first attempt lost, re-run probably catches it.

So: **first response is to do nothing and let it recover.** Check the next morning whether the re-run promoted.

## 3. The part the retries do not fix

`cosign verify` hits Rekor too. With keyless, verification checks log inclusion — so an outage does not only block our promotions, **it blocks users verifying images they already have**, using the exact commands `docs/VERIFY-ARTIFACTS.md` tells them to run.

No amount of CI-side retrying touches that. It is the strongest argument for changing the model, and it is worth weighing separately from build reliability.

## 4. Options, with the facts checked

Verified against the pinned binary (`cosign v3.1.3`), not from memory.

### ⚠️ First, a correction: the key-based approach does not do what it appears to

`RazorfinOS-org/Razorfin` signs with a password-less key pair, and this was suggested as a way off Sigstore:

```bash
cosign sign -y --key env://COSIGN_PRIVATE_KEY "${IMAGE}@${DIGEST}"
```

**There is no tlog flag anywhere in their workflows, and `--tlog-upload` defaults to true.** Razorfin uploads to Rekor on every signature exactly as we do, and would have failed identically in the same outage. Using a key instead of Fulcio does not move the log dependency — you have to disable it explicitly. Recording this because it is a natural and reasonable thing to assume.

### Why this is a trust-model change, not a config change

Keyless *structurally requires* the log. The Fulcio certificate is valid for ~10 minutes; the Rekor entry is the proof that the signature was made while it was valid. Remove the log and a keyless signature stops being verifiable shortly after it is created. There is no "same guarantees, different provider" option.

### A. Key pair + `--tlog-upload=false`

Removes sigstore.dev entirely. Verified supported: `cosign sign --key cosign.key --tlog-upload=false <IMAGE DIGEST>` is a documented example in v3.1.3.

Costs, and they are not small:

- A long-lived private key in repository secrets, with rotation and revocation now our problem.
- cosign's own help text on the verify side: *"ignore transparency log verification... **Artifacts cannot be publicly verified when not included in a log**"*. Verifiers need `cosign.pub` **and** `--insecure-ignore-tlog` — a flag whose name will, correctly, alarm people.
- A leaked key can sign anything retroactively with no log to detect it. Detecting that is what the log is for.
- Directly contradicts `SECURITY.md` ("Published with signed SPDX SBOM attestations", keyless framing) and #1187.

We do already carry a `cosign.pub` at the repo root, inherited from the bluefin port and currently unused by CI — so the plumbing is half-present, which makes this look cheaper than it is.

### B. Keyless + RFC3161 timestamps ✅ *recommended if it recurs*

Keeps ephemeral certificates, replaces the log's timestamping role with a timestamp authority. Verified present in v3.1.3: `--timestamp-server-url` on sign, `--use-signed-timestamps` on verify. `https://timestamp.sigstore.dev` exists and responds.

Why this and not A:

- **Fulcio was never what failed.** Every failure line names `rekor.sigstore.dev/api/v1/log/entries`. This removes precisely the component that broke.
- No long-lived key, so the current key-management story (there isn't one, deliberately) stays intact.
- Fixes the user-side verification problem in §3, which A also fixes but at the cost of §4A.

Still loses public log inclusion, so `SECURITY.md` and `docs/VERIFY-ARTIFACTS.md` need honest rewriting either way.

### C. Keep Rekor, keep shrinking the blast radius

Where we are. Cheapest remaining increment: move the image signature off the critical path the same way attestation went, with promotion gated on "signed **or** recorded pending re-sign" plus a backfill job. This preserves availability but weakens the invariant that nothing unsigned reaches a bare tag — the one thing #1748 deliberately kept.

---

## 5. The decision rule

**Do not re-architect signing on two bad nights.** Rekor was returning `200` when this was written; the observed outages are already inside what #1748 absorbs, and one of them would not have caused any damage at all.

Adopt **option B** when either of these is true:

- A Sigstore outage blocks a promotion **after** #1748 is in effect — i.e. a `SIGSTORE_OUTAGE` marker where the automatic re-run *also* failed. That is the case the current design does not cover.
- Two or more such events in a rolling 30 days, even if each is individually absorbed. Repeated near-misses are a signal, not luck.

Until then, the work is **measurement, not migration**: count how often `SIGSTORE_OUTAGE` appears and whether the re-run cleared it. That turns this from a judgement call into a threshold.

## 6. Open questions for whoever picks this up

- Which TSA? `timestamp.sigstore.dev` keeps us on the same operator whose service just failed — different service, same organisation. A commercial or self-hosted TSA is more independent but more work.
- Do any downstream consumers pin our keyless identity (`--certificate-identity .../reusable-build-image.yml@refs/heads/main`)? Changing the model breaks them and they need warning. `bootc-installer-asahi` and the ISO verification path are the ones to check.
- `publish-iso-groups.yml` signs ISO checksums with `cosign sign-blob` and has the same dependency. Any change has to cover both, or the story becomes inconsistent across artifact types.

Refs #1560, #1748, #1187. Depends on nothing; blocks nothing — this is a decision record, not a task.

hanthor:
## Summary

- record the decision to keep Rekor until a bounded outage threshold is crossed
- define outage-event counting and the immediate response in an operator runbook
- document why a key pair alone does not remove Rekor and why keyless RFC 3161 timestamps are the preferred fallback to evaluate
- link the policy from artifact-verification and maintainer documentation

## Testing

- `git diff --check`
- local Markdown-link validation
- Simplified Technical English lint (3,037 findings; budget ratcheted to 3,037)
- `python3 -m pytest -q <signing contract tests> tests/injection/test_sigstore_rekor_outage.py` (20 passed)

Closes #1750


hanthor:
Closing: the decision in #1750 is to keep a Sigstore outage from colouring the nightly, and that behaviour lands on main with the separate `attest-sbom.yml` run in #2670. This v4 doc conflicts with main's much lower STE budget; the decision itself stays recorded in #1750.

---

Issue #1888, opened 2026-08-18
[ci] Provide reproducible GPU/DRM capacity for desktop boot gates


hanthor:
## Problem

GNOME/NVIDIA gate failures cannot be cleanly distinguished between image defects and a headless QEMU graphics environment with no render node/virgl. XFCE passing on the same class of runner is useful evidence, but it does not establish that GNOME and NVIDIA have equivalent graphics capacity.

## Sequencing

1. Preserve the existing contract diagnostics in #1751 and #1801; do not turn gates advisory.
2. Define one reproducible CI runner capability: KVM, a render node or virgl-compatible QEMU stack, and a recorded QEMU/mesa/DRM fingerprint.
3. Add a capability preflight that reports **infrastructure unavailable** distinctly from a desktop-contract failure.
4. Re-run Guppy GNOME and Albacore GNOME-NVIDIA-HWE on that capacity.
5. Only then make image-side changes for failures that reproduce with the capability preflight passing.

## Acceptance

- The gate summary records GPU/DRM capability and the QEMU display mode.
- A known-good XFCE control and the affected GNOME/NVIDIA jobs run on the same provisioned capability.
- Remaining failures have a desktop-specific log signature, not an ambiguous blank screen.

Depends on #1751 and #1801.

hanthor:
Concrete evidence for step 2, from a Gate that failed today — and it narrows the problem usefully: **there is no GPU capacity to characterise, because the account's quota for it is zero.**

## Measurement

[Run 33041330231](https://github.com/tuna-os/tunaOS/actions/runs/33041330231), `wahoo / cosmic / Gate`, 2026-08-27. The job did not fail a contract or blank-screen — the runner never came up:

```
Failed to launch runner: failed to create fleet: [{
  "ErrorCode":"VcpuLimitExceeded",
  "ErrorMessage":"You have requested more vCPU capacity than your current vCPU
   limit of 0 allows for the instance bucket that the specified instance type
   belongs to. Please visit http://aws.amazon.com/contact-us/ec2-request to
   request an adjustment to this limit.",
  ...
  "InstanceRequirements":{"AllowedInstanceTypes":["g4dn*"],
                          "VCpuCount":{"Max":4,"Min":4}},
  "Lifecycle":"on-demand"
}]
```

Both subnets returned the same code. The launch template asks for a 4-vCPU `g4dn*` on-demand instance; the account's on-demand limit for that bucket is **0**.

## Why this matters to this issue specifically

[...]

PR #2597, opened 2026-09-19
ci: distinguish GPU/DRM infrastructure from boot failures

hanthor:
## Summary

- run a GPU/DRM capability preflight before the candidate image is built or booted
- record KVM, render-node/driver, QEMU, Mesa, GL-device, and QEMU display-mode fingerprints in the Gate summary and evidence artifact
- require virgl for GPU-routed cells so a mis-provisioned runner exits 78 as **infrastructure unavailable** instead of silently falling back and looking like a desktop failure
- pin the subsequent QEMU boot to the mode the preflight proved

## Evidence

- `just fix && just check`
- `297 passed` across the regression convention, boot ownership/CI contract, workflow syntax, and untrusted-context suites
- `shellcheck scripts/gpu-drm-preflight.sh`
- full `just test`: 1479 passed, 18 skipped; its two failures were the new regression docstring convention (fixed and rerun green) and the existing LUKS guest-layout test requiring `sudo`, which is absent in this harness environment

## Infrastructure boundary

Run 33041330231 still establishes that the RunsOn account has a zero-vCPU `g4dn` quota. That failure occurs in **Set up runner**, before checkout, so no repository preflight can execute until AWS capacity is granted. This change makes attached-but-incapable runners unambiguous and produces the requested reproducible fingerprint as soon as capacity attaches; it does not claim to change the external quota.

Closes tuna-os/tunaos#1888


hanthor:
Ported to main in #2670 (main is the canonical branch). Closing this v4 copy.

---

PR #2598, opened 2026-09-19
fix(iso): extend the customize commit deadline

### LINKED ISSUE #1893: ISO build stalls in tacklebox's post-customize `podman commit` when the build runs rootless

hanthor:
Closes #1893.

## Summary

- bump the pinned Tacklebox revision to `ae93e9b`, which adds `TBOX_CUSTOMIZE_COMMIT_TIMEOUT`
- give TunaOS ISO builds a 30-minute commit deadline instead of Tacklebox's 600-second default, while retaining the adapter's 4800-second outer deadline
- pass the setting through both the host-binary and container execution paths and reject invalid values
- document the timeout contract and add an issue-specific regression test

This removes the fixed 600-second failure point measured for `skipjack:xfce` and `marlin:{gnome,kde}`. Callers can still override the inner deadline, including with `0` to rely only on the bounded outer build.

## Verification

- `just fix` — passed
- `just check` — passed
- `bats tests/bats/test_build_iso_tacklebox.bats` — 40 passed
- `python3 -m pytest -q tests/regressions/test_issue_1893_customize_commit_deadline.py tests/test_regression_convention.py` — 27 passed
- `gh api repos/tuna-os/tacklebox/commits/ae93e9b5bacef49746cb1b08dc805c3149d7c9c8` resolves to `fix(customize): make commit timeout configurable (#300)`

The full `just test` run reached 82% before the local 20-minute command budget expired. Its one observed failure is environment-only: this harness has no `sudo`, so the unrelated LUKS guest-layout fixture reports `sudo: command not found`.


hanthor:
Closing this v4 copy: main is the canonical branch, and this change sits on top of v4's TBOX_* environment forwarding in `scripts/lib/tacklebox.sh`, which main doesn't have. It didn't cherry-pick into #2670. #1893 stays open for a main port that brings the forwarding and the 1800s commit deadline together.

---

PR #2608, opened 2026-09-19
fix(el10): hold rolling bases on libselinux 3.10

### LINKED ISSUE #2485: dbus-broker cannot read dbus_contexts on the rolling EL10 bases — 8 gnome cells red, and kde promotes without a system bus

hanthor:
## Summary

- declare libselinux 3.10-2.el10 as a temporary ceiling for the rolling EL10 bases
- downgrade and versionlock the libselinux runtime, utilities, and Python bindings on yellowfin and skipjack only
- fail the build if the requested EVR does not land, and add an incident regression test plus troubleshooting record

## Evidence

The controlled skipjack experiment in run 35191766126 passed the Gate, desktop contract, promotion, and ISO after moving libselinux from 3.11 to 3.10. The otherwise equivalent 3.11 boot in run 35185567440 lost its system bus. 3.10-2 differs from the measured 3.10-1 build only by the restorecon ENOENT fix for RHEL-110181.

I also verified that all three pinned RPMs (`libselinux`, `libselinux-utils`, and `python3-libselinux`) at 3.10-2.el10 remain available in both AlmaLinux Kitten 10 and CentOS Stream 10 BaseOS for x86_64 and aarch64.

## Checks

- `python3 -m pytest -q tests/regressions/test_issue_2485_rolling_el10_libselinux_ceiling.py tests/test_regression_convention.py tests/test_shell_formatting_contract.py` (35 passed)
- `bash -n build_scripts/10-base-packages.sh`
- `shfmt v3.14.1 -d build_scripts/10-base-packages.sh`
- `git diff --check`

Fixes #2485


hanthor:
Ported to main in #2670 (main is the canonical branch). This is the fix #2485 needs; the relabel approach in #2658 is closed because main's probe shows the labels are already correct. Closing this v4 copy.

---

